PR #1, opened 2015-10-18
feat: Closure compile, golden file base tests.

mprobst:
Golden files are easier to author, and easier to
check for correctness with Closure compiler.

This also formulates tests in terms of the final
Closure style output, not the intermediate
annotated TypeScript program.


gulpfile.js
```diff
@@ -73,7 +73,7 @@ gulp.task('test.unit', ['test.compile'], function(done) {
     done();
     return;
   }
-  return gulp.src('build/test/**/*.js').pipe(mocha({timeout: 500}));
+  return gulp.src('build/test/**/*.js').pipe(mocha({timeout: 5000}));
```
mprobst: Closure :-(

mprobst:
Friendly Ping @rkirov :-)

gulpfile.js
```diff
@@ -73,15 +73,24 @@ gulp.task('test.unit', ['test.compile'], function(done) {
     done();
     return;
   }
-  return gulp.src('build/test/**/*.js').pipe(mocha({timeout: 500}));
+  return gulp.src(['build/test/**/*.js', '!build/test/**/e2e*.js']).pipe(mocha({timeout: 1000}));
 });
 
-gulp.task('test', ['test.unit', 'test.check-format']);
+gulp.task('test.e2e', ['test.compile'], function(done) {
+  if (hasError) {
```
rkirov: interesting pattern, there has to be a better way to compose errors in tasks :) (no action required)

package.json
```diff
@@ -14,6 +14,7 @@
   "devDependencies": {
     "chai": "^2.1.1",
     "clang-format": "1.0.30",
+    "closure-compiler": "mprobst/node-closure#void-args",
```
rkirov: looks like this can use 0.12 https://github.com/tim-smart/node-closure/commit/9be478c63e4923010d4c12ca4b9e1975b9c2b639

src/sickle.ts
```diff
@@ -14,10 +14,12 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
-export type AnnotatedProgram = {
+export type StringMap = {
```
rkirov: maybe no need to export this

test/e2e_test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  var CLOSURE_COMPILER_OPTS: {[k: string]: string | string[]} = {
+    'checks-only': null,
+    'jscomp_error': 'checkTypes',
+    'js': jsFiles,
+    'language_in': 'ECMASCRIPT6'
+  };
+
+  compile(null, CLOSURE_COMPILER_OPTS, (err, stdout, stderr) => {
+    // console.log('Closure compilation:', total, 'done after', Date.now() - startTime, 'ms');
```
rkirov: why not keep it

test/test_support.ts
```diff
@@ -1,42 +1,92 @@
...
+
 export function expectSource(src: string) {
-  return expect(transformSource(src));
+  var annotated = annotateSource(src);
+  // console.log('Annotated', annotated);
+  var transformed = transformSource(annotated);
+  return expect(transformed);
+}
+
+export interface GoldenFileTest { name: string, tsPath: string, jsPath: string, }
```
rkirov: my eyes got confused here, lets stick to the line-per-field syntax.

test/test_support.ts
```diff
@@ -1,42 +1,92 @@
...
+
+export function goldenTests(): GoldenFileTest[] {
+  var tsExtRe = /\.ts$/;
+  var testFolder = path.join(__dirname, '..', '..', 'test_files');
+  var files = fs.readdirSync(testFolder).filter((fn) => !!fn.match(tsExtRe));
+  return files.map((fn) => {
```
rkirov: No need for {} block if you only have a return statement, you can use arg => expression syntax as `fn => {name: fn, ...}`

test/test_support.ts
```diff
@@ -1,42 +1,92 @@
...
+    throw new Error(formatDiagnostics(emitRes.diagnostics));
+  }
+  expect(Object.keys(transformed)).to.deep.equal(['main.js']);
+  return transformed['main.js'];
+}
+
 export function expectSource(src: string) {
-  return expect(transformSource(src));
+  var annotated = annotateSource(src);
+  // console.log('Annotated', annotated);
```
rkirov: remove

test/e2e_test.ts
```diff
@@ -0,0 +1,33 @@
...
+    done(err);
+  });
+}
+
+describe('golden file tests', () => {
+  it('generates correct Closure code', (done: (err: Error) => void) => {
+    var jsMasters = goldenTests().map((t) => t.jsPath);
```
rkirov: masters -> goldens ?

rkirov:
LGTM.

This strategy of goldens works pretty well for clutz, so it is nice to have it here too.

gulpfile.js
mprobst: Yeah, gulp has no systematic way to deal with errors, and surprisingly it's still better than all the other tools in that regard...

package.json
mprobst: Done (`0.2.10`).

src/sickle.ts
mprobst: `error TS4081: Exported type alias 'AnnotatedProgram' has or is using private name 'StringMap'.`

Which makes sense, I guess?

test/e2e_test.ts
mprobst: Done.

test/e2e_test.ts
mprobst: Done.

test/test_support.ts
mprobst: Actually I'd like to keep. It's very spammy, but super useful to understand what's going on in one test. Maybe medium term we should have an `expect with message` for this.

test/test_support.ts
mprobst: Whoops, done.

test/test_support.ts
mprobst: You'd think - but not if what you're returning is an object literal:

This code gets parsed as a block with a goto label `foo:` and a sole property access `bar`:

```js
var fn = (x) => { foo: bar };
```

PR #2, opened 2015-11-02
Support Fields

mprobst:


mprobst:
Note that this is based on #1.

rkirov:
no diffbase :( I will take a look again tomorrow after you submit #1

mprobst:
Should be more reviewable now.

src/sickle.ts
```diff
@@ -58,10 +62,49 @@ class Annotator {
...
+      case ts.SyntaxKind.ClassDeclaration: {
+        let classNode = <ts.ClassDeclaration>node;
+        let hasCtor = classNode.members.some((e) => e.kind === ts.SyntaxKind.Constructor);
+        if (hasCtor) {
+          this.writeNode(classNode);
+          break;
+        }
+        // Emit a synthetic ctor.
+        // TODO(martinprobst): Handle inherited parent ctors.
+        this.writeTextBetween(classNode, classNode.getFullStart(), classNode.getLastToken());
```
rkirov: every single call in this cl passes `writeTextBetween(node, node.getFullStart(), child)` maybe create a helped `writeTextUntilChild(node, child)`

src/sickle.ts
```diff
@@ -58,10 +62,49 @@ class Annotator {
...
+      case ts.SyntaxKind.Constructor: {
+        let ctor = <ts.ConstructorDeclaration>node;
+        this.writeTextBetween(ctor, ctor.getFullStart(), ctor.body);
+        if (ctor.body.statements.length) {
+          this.writeTextBetween(ctor.body, ctor.body.getFullStart(), ctor.body.statements[0]);
+        } else {
+          this.emit(' {');
+        }
+
+        let paramProps = ctor.parameters.filter((p) => !!(p.flags & VISIBILITY_FLAGS));
```
rkirov: how about transfering the visibility info too, not just using it for filtering? Are there any complications?

src/sickle.ts
```diff
@@ -58,10 +62,49 @@ class Annotator {
...
+        this.writeNode(classNode.getLastToken());
+        break;
+      }
+      case ts.SyntaxKind.Constructor: {
+        let ctor = <ts.ConstructorDeclaration>node;
+        this.writeTextBetween(ctor, ctor.getFullStart(), ctor.body);
+        if (ctor.body.statements.length) {
+          this.writeTextBetween(ctor.body, ctor.body.getFullStart(), ctor.body.statements[0]);
+        } else {
+          this.emit(' {');
```
rkirov: is that even possible, I am reading this as a constructor with not body, i.e. ```class A{ constructor(); }```. Typescript playground says thats invalid. Otherwise, we need a test of such a beast.

test_files/fields.ts
```diff
@@ -0,0 +1,14 @@
+class Klass {
+  field1: string;
+  field2: number;
+  field3: number;
```
rkirov: need some param props to test too.

src/sickle.ts
```diff
@@ -114,17 +178,20 @@ class Annotator {
       this.visit(child);
       lastEnd = child.getEnd();
     }
+    // Write any trailing text.
```
rkirov: lol, Daria had exactly the same bug, in the basically the same code. I should have caught it in the first PR pass.

rkirov:
I think you have a few features that have no tests, otherwise LGTM.

src/sickle.ts
mprobst: Yeah, I'm not happy with these functions. There are users that pass a write offset; I split it into two functions.

src/sickle.ts
mprobst: Let's do that in a latter CL.

src/sickle.ts
mprobst: Sorry, the code was unclear. The problem is that I didn't know how to find a good insertion place if the `ctor.body` contains no `statements` at all. Fixed using `getEnd()`.

test_files/fields.ts
mprobst: Done.

PR #8, opened 2016-01-07
improve logging of visited nodes

martine:
When logging visted nodes for debugging purposes, indent them
based on the current recursion level of visit().  This helps
keep clear which piece of code is responsible for which outputs.

Output (with the logging uncommented) looks like:

```
node: SourceFile
| node: SyntaxList
| | node: VariableStatement
| | | node: VariableDeclarationList
| | | | node: VarKeyword
| | | | node: SyntaxList
| | | | | node: VariableDeclaration
| | | | | | node: StringKeyword
| | | | | | node: Identifier
| | | | | | node: ColonToken
| | | | | | node: StringKeyword
| | | node: SemicolonToken
| | node: VariableStatement
[...]
```

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -28,8 +28,9 @@ const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.Node
  */
 class Annotator {
   private output: string[];
+  private indent: number;
 
-  constructor() {}
+  constructor() { this.indent = 0; }
 
   annotate(args: string[]): AnnotatedProgram {
     let tsArgs = ts.parseCommandLine(args);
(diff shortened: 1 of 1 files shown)


martine:
Not certain you want to merge this, but I thought I'd try out the code review process anyway to make sure I got it all right.  Feel free to reject if you don't want it. :)

src/sickle.ts
```diff
@@ -55,8 +56,17 @@ class Annotator {
...
+    var prefix = '';
+    for (let i = 1; i < this.indent; i++) {
+      prefix += '| ';
+    }
+    console.log(prefix + message);
+  }
+
   private visit(node: ts.Node) {
-    // console.log('node:', (<any>ts).SyntaxKind[node.kind]);
+    this.indent++;
```
mprobst: nit/optional: maybe add a sanity check `indent == 0` when done visiting nodes? That way we can be certain we don't mess this up.

src/sickle.ts
```diff
@@ -55,8 +56,17 @@ class Annotator {
 
   private emit(str: string) { this.output.push(str); }
 
+  private logWithIndent(message: string) {
+    var prefix = '';
```
mprobst: YMMV, but `new Array(this.indent + 1).join('| ')` is fewer lines of code ;-)

mprobst:
LGTM.

As to the merging: please do that yourself. DO NOT USE THE MERGE BUTTON, but rather rebase onto master, merge into master with `git merge --ff-only mybranch`, and then push to the upstream remote. We prefer to have a linear history, if possible.

PR #10, opened 2016-01-12
translate "type" statements as Closure @typedefs

martine:


mprobst:
LGTM.

src/sickle.ts
```diff
@@ -129,6 +129,8 @@ class Annotator {
         this.writeTextFromOffset(writeOffset, fnDecl.body);
         this.visit(fnDecl.body);
         break;
+      case ts.SyntaxKind.TypeAliasDeclaration:
+        this.visitTypeAlias(<ts.TypeAliasDeclaration>node);
```
mprobst: break?

martine: Nice catch!  Wow, that's what I get for writing a lot of Go code...

(The reason this even worked is that "writeNode" previously did nothing for "type" statements so the fallthrough didn't break anything.  I fixed it though.)

mprobst:
FYI @martine is a Googler, this is just a random hiccup of the automation around googlebot.

willnorris:
Not a hiccup.  Evan should be using his @google.com email address (not @danga.com) in the git commit if you want this to be recognized as a Googler contribution.

mprobst:
@willnorris I'm using a non-@google.com address and it works fine. Maybe @martine just has to sign the CLA with his private mail account?

willnorris:
you shouldn't be.  All Googlers should be using their @google.com email for any work related projects.

martine:
I'm already registered with this address I think. At least this github
account has the Google org on it. I'll tr y switching the email address.

martine:
OK, I'll ensure my subsequent changes are from @google.com addresses.

If it really matters I can also rewrite all my commits to @google.com addresses.  It's not so horrible because they're at the top of the commit log but it potentially breaks anyone's branches.  How much do we care?

Issue #13, opened 2016-01-15
Interface should accept a ts.SourceFile, not a ts.Program


martine:
We don't want to worry about the compile flags etc. that a ts.Program needs for construction.

PR #14, opened 2016-01-19
Change Sickle main entry point to ts.SourceFile

martine:
ts.Program (the previous parameter) is a fully compiled program of
potentially multiple files, while ts.SourceFile is a single parsed file.
It appears from how we're using sickle that we want the latter, not the
former.

(It may be the case that we need a fully compiled program to get all the
references right, but for now this appears to work.)

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -14,12 +14,6 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
-export type StringMap = {
-  [fileName: string]: string
-};
-
-export type AnnotatedProgram = StringMap;
-
 const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.NodeFlags.Public;
 
 /**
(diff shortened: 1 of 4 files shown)


src/sickle.ts
```diff
@@ -235,13 +213,16 @@ function last<T>(elems: T[]): T {
...
-export function annotateProgram(program: ts.Program): AnnotatedProgram {
-  return new Annotator().annotateProgram(program);
+export function annotate(file: ts.SourceFile): string {
+  return new Annotator().annotate(file);
 }
 
 // CLI entry point
 if (require.main === module) {
-  let res = new Annotator().annotate(process.argv);
   // TODO(martinprobst): Do something useful here...
```
rkirov: looks like you handled the TODO (it was quite vague anyways), remove it.

src/sickle.ts
```diff
@@ -235,13 +213,16 @@ function last<T>(elems: T[]): T {
...
-  console.log(JSON.stringify(res));
+  for (let path of process.argv.splice(2)) {
+    let sourceText = ts.sys.readFile(path, 'utf-8');
+    let sf = ts.createSourceFile(path, sourceText, ts.ScriptTarget.ES6, true);
+    console.log(annotate(sf));
```
rkirov: output the filename too.

rkirov:
If you add `closes #13` to the commit message, github will autoclose that issue.

PR #16, opened 2016-01-22
keep existing closure annotations

### LINKED ISSUE #12: Merge existing closure annotations with output ones
For example for angular1 templates, we can have a field with export annotation:
```typescript
class Ctrl {
  /** @export */
  foo: string;
}
```
sickle needs to output
```javascript
/** @export @type {string} */
this.foo;
```

rkirov:
Certain closure annotations have not equivalent in TS, especially the
ones concerned with minification, not type-checking (like @export). The
authors of the code will have to write them in TS and sickle should perserve
them.

Closes #12

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,9 @@ typings
 
 # Dependency directory
 node_modules
+
+# WebStorm/IntelliJ directory
+.idea
+
+# VSCode directory
+.vscode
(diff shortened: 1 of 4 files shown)


rkirov:
Added a test for a non jsdoc comment.

rkirov:
merged

martine:
Note: this doesn't quite work, see issue #22.

PR #17, opened 2016-01-23
use full file text for grabbing substrings from offsets

martine:
When a node has a file-relative offset index, you must use the file's
full text (this.file.text / this.file.getFullText()) and not the file's
text with initial comments removed (this.file.getText()).

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -82,13 +82,9 @@ class Annotator {
 
         if (ctor.body.statements.length) {
           let firstStmt = ctor.body.statements[0];
-          this.emit(
-              ctor.body.getSourceFile().getText().substring(
-                  firstStmt.getFullStart(), ctor.body.getEnd()));
+          this.writeRange(firstStmt.getFullStart(), ctor.body.getEnd());
         } else {
-          let remaining = ctor.getSourceFile().getText().substring(
-              ctor.body.getLastToken().getFullStart(), ctor.body.getEnd());
-          this.emit(remaining);
+          this.writeRange(ctor.body.getLastToken().getFullStart(), ctor.body.getEnd());
         }
 
         break;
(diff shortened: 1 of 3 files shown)


test_files/file_comment.js
```diff
@@ -0,0 +1,5 @@
+/** @return { string} */ // This test verifies that initial comments don't confuse offsets.
```
rkirov: Filed #18 to fix this.

src/sickle.ts
```diff
@@ -82,12 +82,10 @@ class Annotator {
 
         if (ctor.body.statements.length) {
           let firstStmt = ctor.body.statements[0];
-          this.emit(
-              ctor.body.getSourceFile().getText().substring(
-                  firstStmt.getFullStart(), ctor.body.getEnd()));
+          this.emit(this.file.text.substring(firstStmt.getFullStart(), ctor.body.getEnd()));
```
mprobst: maybe extract `this.file.text` into a method, and add a comment to it explaining this? It's pretty subtle...

src/sickle.ts
```diff
@@ -205,24 +203,24 @@ class Annotator {
       lastEnd = child.getEnd();
     }
     // Write any trailing text.
-    let text = node.getSourceFile().getText().slice(lastEnd, node.getEnd());
+    let text = this.file.text.slice(lastEnd, node.getEnd());
```
mprobst: nit: maybe use either slice or substring, consistently?

src/sickle.ts
```diff
@@ -200,29 +196,31 @@ class Annotator {
...
   }
 
-  private writeTextFromOffset(from: number, to: ts.Node) {
-    let text = to.getSourceFile().getText().slice(from, to.getFullStart());
-    if (text) this.emit(text);
+  private writeTextBetween(node: ts.Node, to: ts.Node) {
+    this.writeRange(node.getFullStart(), to.getFullStart());
   }
 
-  private writeSourceBefore(offset: number, node: ts.Node) {
```
martine: This function was actually the same as the function above it, so I merged them!

martine:
PTAL

src/sickle.ts
```diff
@@ -200,29 +196,31 @@ class Annotator {
...
+    this.writeRange(lastEnd, node.getEnd());
   }
 
-  private writeTextBetween(node: ts.Node, to: ts.Node) {
-    let text = node.getSourceFile().getText().slice(node.getFullStart(), to.getFullStart());
+  // Write a span of the input file as expressed by absolute offsets.
+  // These offsets are found in attributes like node.getFullStart() and
+  // node.getEnd().
+  private writeRange(from: number, to: number) {
+    let text = this.file.text.slice(from, to);
```
mprobst: IMHO still needs a comment explaining why `getSourceFile().getText()` is wrong here.

mprobst:
One nit, LGTMs otherwise.

Issue #18, opened 2016-01-23
sickle should strip all comments


rkirov:
in file_comments.ts we emit jsdoc before existing comment:

```
/** @return { string} */ // This test verifies that initial comments don't confuse offsets.
 function foo() {
     return 'foo';
 }
```

we should strip all comments, since we already don't emit comments in properties (see comments.ts golden).

mprobst:
I think this is obsolete. It turns out emitting some comments is fine, and useful for downstream code.

PR #19, opened 2016-01-26
add an option to disable all type translation

martine:
This causes types to be translated as {?}, causing the Closure
compiler to not check them.  This is not as awful as it might
seem at first because we believe the TypeScript compiler already
has checked the types.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -1,5 +1,11 @@
 import * as ts from 'typescript';
 
+export interface SickleOptions {
+  // If true, convert every type to the Closure {?} type, which means
+  // "don't check types".
+  untyped?: boolean;
+}
+
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
   return diags.map((d) => {
                 let res = ts.DiagnosticCategory[d.category];
@@ -25,7 +31,7 @@ class Annotator {
   private indent: number;
   private file: ts.SourceFile;
 
-  constructor() { this.indent = 0; }
+  constructor(private options: SickleOptions) { this.indent = 0; }
 
   annotate(file: ts.SourceFile): string {
     this.output = [];
@@ -164,14 +170,18 @@ class Annotator {
   }
 
   private maybeVisitType(type: ts.TypeNode, jsDocTag?: string) {
-    if (!type) return;
+    if (!type && !this.options.untyped) return;
     this.emit(' /**');
     if (jsDocTag) {
       this.emit(' ');
       this.emit(jsDocTag);
       this.emit(' {');
     }
-    this.visit(type);
+    if (this.options.untyped) {
+      this.emit(' ?');
+    } else {
+      this.visit(type);
+    }
     if (jsDocTag) {
       this.emit('}');
     }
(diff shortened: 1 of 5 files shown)


martine:
Note: I am not sure how to test this.

src/sickle.ts
```diff
@@ -238,8 +252,8 @@ function last<T>(elems: T[]): T {
   return elems.length ? elems[elems.length - 1] : null;
 }
 
-export function annotate(file: ts.SourceFile): string {
-  return new Annotator().annotate(file);
+export function annotate(file: ts.SourceFile, options?: SickleOptions): string {
+  return new Annotator(options || defaultOptions).annotate(file);
```
mprobst: Create a new object that clones defaultOptions, then assign the properties from options?

mprobst:
Re testing, what about writing a manual test that calls sickle with the right args and checks for the `{?}`?

I don't think we can reasonably test that Closure understands this.

src/sickle.ts
martine: I'm not sure I followed you here -- what problem are you worried about, and what is your suggested fix?

martine:
Added a test (special-cased the filename, it's gross but not so bad).

src/sickle.ts
mprobst: So if you keep adding defaultOptions and values, this currently either takes the defaults, or the given options, but doesn't individually override, which is the behaviour I'd expect. Does that make sense?

mprobst:
LGTMs with the test.

src/sickle.ts
martine: Ah right.  I should make all the fields on options optional.  I will ping you pack when this is done.

martine:
OK, now with the options set up.

test_files/basic.untyped.js
```diff
@@ -0,0 +1,17 @@
+/** @return { ?} */ // This test is just a random collection of typed code, to
+// ensure the output is all with {?} annotations.
```
rkirov: argh issue #18 should fix this terrible output

test_files/basic.untyped.js
```diff
@@ -0,0 +1,17 @@
...
+function func(/** ? */ arg1) {
```
rkirov: this one is missing @type

rkirov:
(not blocking this PR) We need to follow up with js compiler folks about how /** @type {?} */ differs from having no type annotation at all. Because the less we have to muck with the code the better. I rather have untyped mode not output any type annotations except move /** @export */ to the right place and some @supress to not have warnings.

test_files/basic.untyped.js
martine: Would you believe we don't get this right even in the typed case?  I just checked!

rkirov: lol, file an issue.

mprobst: This is actually valid syntax for inline declarations. You could write `/** @param {string} foo */` before, or `function(/** string */ foo) { ... }*/ inline. Also works for variables, `var /** string */ foo;`. It's not much nicer than regular JSDoc, but it's easier to emit because it's at the same syntactical level as the symbol itself.

martine: This had failed to compile before, but rechecking now I think it's due to the comment in the middle, which is issue #18.

PR #21, opened 2016-01-26
don't copy jsdoc annotations into constructors

### LINKED ISSUE #20: Annotations on parameter properties show up in wrong place
See "parameter properties" in this:
http://www.typescriptlang.org/Handbook#classes-privatepublic-modifiers 

Code like

```js
class Foo {
  constructor(/** @export */ public bar: string) {}
}
```

needs to have us move the @export declaration onto the stub decl and remove it from the parameter list.

martine:
Input code like
```js
  constructor(/** @export */ public foo: string) {}
```
cannot be translated to
```js
  constructor(/** @export */ foo) ...
```
as that is illegal to Closure.

Instead we to generate something more or less like:
```js
  constructor(foo) {
    this.foo = foo;
    /** @export */
    this.foo;
  }
```
which is legal.

Fixes #20.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -72,9 +72,32 @@ class Annotator {
         this.writeNode(classNode.getLastToken());
         break;
       }
+      case ts.SyntaxKind.PublicKeyword:
+      case ts.SyntaxKind.PrivateKeyword:
+        // The "public"/"private" keywords are encountered in two places:
+        // 1) In class fields (which don't appear in the transformed output).
+        // 2) In "parameter properties", e.g.
+        //      constructor(/** @export */ public foo: string).
+        // In case 2 it's important to not emit that JSDoc in the generated
+        // constructor, as this is illegal for Closure.  It's safe to just
+        // always skip comments preceding the 'public' keyword.
+        // See test_files/parameter_properties.ts.
+        this.writeNode(node, /* skipComments */ true);
+        break;
       case ts.SyntaxKind.Constructor: {
         let ctor = <ts.ConstructorDeclaration>node;
-        this.writeTextBetween(ctor, ctor.body);
+        // Write the "constructor(...) {" bit, but iterate through any
+        // parameters if given so that we can examine them more closely.
+        let offset = ctor.getFullStart();
+        if (ctor.parameters.length) {
+          for (let param of ctor.parameters) {
+            this.writeTextFromOffset(offset, param);
+            this.visit(param);
+            offset = param.getEnd();
+          }
+        }
+        this.writeTextFromOffset(offset, ctor.body);
+
         if (ctor.body.statements.length) {
           // Insert before the first code in the ctor.
           this.writeTextBetween(ctor.body, ctor.body.statements[0]);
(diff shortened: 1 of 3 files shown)


src/sickle.ts
```diff
@@ -146,12 +168,14 @@ class Annotator {
    * Returns empty string if there is no existing annotation.
    */
   private existingClosureAnnotation(p: ts.PropertyDeclaration | ts.ParameterDeclaration) {
-    let comments = ts.getLeadingCommentRanges(this.file.text, p.pos);
+    let text = p.getFullText();
+    let comments = ts.getLeadingCommentRanges(text, 0);
```
martine: For some reason the old code was unable to get comments unless I did this first.
I think this is harmless though so I don't mind the change.

src/sickle.ts
```diff
@@ -66,9 +66,31 @@ class Annotator {
         this.writeNode(classNode.getLastToken());
         break;
       }
+      case ts.SyntaxKind.PublicKeyword:
```
rkirov: we need to handle Private too

src/sickle.ts
rkirov: odd, but indeed it is harmless. That whole getLeadingCommentRanges api is very odd. I don't know why they dont' have p.getLeadingCommentRanges without option for passing text (and thus getting it wrong by not using fulltext).

src/sickle.ts
```diff
@@ -187,12 +211,15 @@ class Annotator {
     this.emit(';\n');
   }
 
-  private writeNode(node: ts.Node) {
+  private writeNode(node: ts.Node, skipComments?: boolean) {
```
rkirov: how about `skipComments: boolean = false` instead, so that we keep `undefined` out of the runtime.

src/sickle.ts
```diff
@@ -187,12 +211,15 @@ class Annotator {
...
     if (node.getChildCount() == 0) {
       // Directly write complete tokens.
-      this.emit(node.getFullText());
+      this.emit(skipComments ? node.getText() : node.getFullText());
       return;
     }
+    if (skipComments) {
+      this.fail('skipComments unimplemented for complex Nodes');
```
rkirov: can be done in a follow up, but fail should emit the node it failed at.

rkirov:
minor nitpick, looks good after adding private too.

src/sickle.ts
martine: I considered this, but my best guess is that we don't, because there's no reason to @export a private variable.  What do you think?

src/sickle.ts
martine: Good idea, done.

src/sickle.ts
martine: Agreed.

src/sickle.ts
rkirov: agree, we can make sickle throw in that case. But if we do nothing, someone will write it sooner or later and sickle will produce invalid closure code.

martine: OK, I made it match "public".  I think that is correct/harmless anyway.

src/sickle.ts
martine: I noticed there are functions with names like getJSDoc in the TypeScript source but for some reason they're not in the public API.  Might be a version thing.

PR #24, opened 2016-01-28
don't emit typedefs in untyped mode

martine:
In untyped mode we don't use types anymore, so we don't need to
emit typedefs either.


mprobst:
LGTM

PR #28, opened 2016-01-29
preserve type decls when emitting

martine:
Sickle is supposed to emit the same source with extra Closure bits.
This code previously stripped all type declarations.

Because the tests test the behavior including ES5 generation, it's
hard to make this show up in a test (because types are stripped in
ES5 output).  I verified manually (by logging the intermediate state)
that this is doing the right thing.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -138,6 +138,7 @@ class Annotator {
         break;
       case ts.SyntaxKind.TypeAliasDeclaration:
         this.visitTypeAlias(<ts.TypeAliasDeclaration>node);
+        this.writeNode(node);
         break;
       default:
         this.writeNode(node);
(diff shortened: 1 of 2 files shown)


src/sickle.ts
```diff
@@ -144,6 +144,7 @@ class Annotator {
         break;
       case ts.SyntaxKind.TypeAliasDeclaration:
         this.visitTypeAlias(<ts.TypeAliasDeclaration>node);
+        this.writeNode(node);
```
rkirov: add a test?

martine: Fixed!

test/test_support.ts
```diff
@@ -56,9 +57,10 @@ function transformSource(src: string): string {
   };
 
   var program = ts.createProgram(['main.ts'], OPTIONS, host);
-  if (program.getSyntacticDiagnostics().length) {
```
martine: This was only checking "SyntacticDiagnostics", while the latter code checked "PreEmitDiagnostics".  The latter is a superset of the former.

src/sickle.ts
```diff
@@ -212,12 +213,13 @@ class Annotator {
...
   private visitTypeAlias(node: ts.TypeAliasDeclaration) {
     if (this.options.untyped) return;
+    // Write a Closure typedef, which involves an unused "var" declaration.
     this.emit('/** @typedef {');
     this.visit(node.type);
     this.emit('} */\n');
     this.emit('var ');
     this.emit(node.name.getText());
-    this.emit(';\n');
+    this.emit(': any;\n');
```
rkirov: Lets emit `void` type for this variable, so that users can't shoot themselves in the foot by using this generated value.

PR #30, opened 2016-01-29
have range-writing functions return end offset

martine:
We have a repeated pattern where we emit a span of text,
do some logic, then emit more text picking up from where the
previous span left off.  Refactor the functions such that
keeping track of these offsets is simpler.

No functionality change, just simpler code.


mprobst:
LGTM

Issue #31, opened 2016-01-29
add goldens for different steps of the pipeline


rkirov:
We have the following pipeline:

.ts -> (sickle) -> .ts with annotations -> (ts compiler) -> .js with annotations -> (closure compiler) -> minified .js

We currently have goldens only for ".js with annotations". It would be helpful to catch bugs earlier with goldens for ".ts with annotations" and maybe also goldens for "minified .js".

PR #32, opened 2016-01-29
save intermediate steps as goldens

martine:
Our build pipeline is:
1) input test.ts
2) sickle (add annotations, rearrange code)
3) typescript (convert TypeScript to ES6)
4) closure (convert ES6 to ES5)

Our current goldens are only the output of step 3.
This change makes us also save the output of step 2.

I'm leaving off saving the Closure output because minification
doesn't make it too useful.  Maybe we will revisit this.

Fixes part of #31.


martine:
Don't forget me please!

test_files/sickle/static.ts
```diff
@@ -0,0 +1,11 @@
+class Static {
+  // This should not become a stub declaration.
+  static bar = 3;constructor() {
```
rkirov: ew, glad you check in those files, so we can see the ugly output of sickle :) I wonder if it is worth producing better looking code at some point. (no action required).

rkirov:
LGTM

PR #33, opened 2016-02-01
include line:column in error messages

martine:
Track the current ts.Node while traversing, and include the
offset of that node when failing an assertion.  This isn't
great at finding the correct point but it is at least close.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -30,6 +30,9 @@ class Annotator {
   private output: string[];
   private indent: number;
   private file: ts.SourceFile;
+  // The node currently being visited by visit().
+  // This is only used in error messages.
+  private currentNode: ts.Node;
 
   constructor(private options: SickleOptions) { this.indent = 0; }
 
@@ -37,7 +40,7 @@ class Annotator {
     this.output = [];
     this.file = file;
     this.visit(file);
-    assert(this.indent == 0, 'visit() failed to track nesting');
+    this.assert(this.indent == 0, 'visit() failed to track nesting');
     return this.output.join('');
   }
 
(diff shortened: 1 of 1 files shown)


src/sickle.ts
```diff
@@ -259,17 +266,20 @@ class Annotator {
...
-    return this.writeRange(from, to);
+    return this.writeRange(from, node.getFullStart());
   }
 
-  private fail(msg: string) { throw new Error(msg); }
-}
-
-function assert(condition: boolean, msg: string) {
-  if (!condition) throw new Error(msg);
+  private fail(msg: string) {
```
mprobst: Random drive by review comment: maybe compare with this code here, where we solved the problem previously?

https://github.com/angular/ts2dart/blob/master/lib/main.ts#L241

martine: That is way better.  I will send an update when this is ready for re-review.

martine:
PTAL

src/sickle.ts
```diff
@@ -261,15 +265,19 @@ class Annotator {
...
+    this.assert(to > from, `Offset must not be smaller; ${to} vs ${from}`);
     return this.writeRange(from, to);
   }
 
-  private fail(msg: string) { throw new Error(msg); }
-}
+  private fail(msg: string) {
+    let offset = this.currentNode.getFullStart();
+    let {line, character} = this.file.getLineAndCharacterOfPosition(offset);
+    throw new Error(`near node starting at ${line+1}:${character+1}: ${msg}`);
```
rkirov: I was thinking whether we should add a test for this, but I don't see yet a failure / assertion that would be useful to the external user of the tool. All the errors we have are inconsistencies in our internal state, as which point you can argue it is not worth the overhead of testing that. What do you think?

rkirov:
LGTM

mprobst:
LGTM.

PR #36, opened 2016-02-02
insert stub decls at bottom of ctors

### LINKED ISSUE #27: Stub decls must appear after "super" call
```javascript
class Foo {
  constructor(public x: string) {
    super(3);
  }
}
```

becomes

```javascript
class Foo {
  constructor(public x: string) {

// Sickle: begin stub declarations.

 /** @type { string} */
this.x;
// Sickle: end stub declarations.

    super(3);
  }
}
```

It's illegal in TS to have code before the super call apparently.

martine:
When we insert stub declarations in a constructor,
we must insert them after any existing super() call,
but only in the special case(!) where there are initialized
properties or if there are parameter properties.

Rather than implement careful logic for this, instead just
always put the stub declaration at the end of the constructor.
I verified that you can put type declarations at the bottom
and they still constrain initializers that occur above them.

Fixes #27.

### DIFF
--- a/test_files/super.ts
+++ b/test_files/super.ts
@@ -1,0 +1,25 @@
+class SuperTestBase {
+  constructor(public x: number) {}
+}
+
+// A ctor with a parameter property.
+class SuperTestDerivedParamProps extends SuperTestBase {
+  constructor(public y: string) {
+    super(3);
+  }
+}
+
+// A ctor with an initialized property.
+class SuperTestDerivedInitializedProps extends SuperTestBase {
+  y: string = 'foo';
+  constructor() {
+    super(3);
+  }
+}
+
+// A ctor with a super() but none of the above two details.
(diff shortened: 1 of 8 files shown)


test_files/super.ts
```diff
@@ -0,0 +1,25 @@
...
+// A ctor with an initialized property.
+class SuperTestDerivedInitializedProps extends SuperTestBase {
+  y: string = 'foo';
+  constructor() {
+    super(3);
+  }
+}
+
+// A ctor with a super() but none of the above two details.
+class SuperTestDerivedOrdinary extends SuperTestBase {
```
rkirov: This case doesn't seem too useful. I think I can argue that if the other two pass this one has to pass.

martine: Yes, it was mostly useful for comparing the output of all three before I simplified the logic to not special case the other two.  I could remove it I guess?  But it is one of the only instances of `super` in this code...

PR #37, opened 2016-02-02
don't mangle decorators on processing

### LINKED ISSUE #35: Sickle chokes on decorators (likely due to a bug in TS < 1.8)
Input:
```javascript
class DecoratorTest {
  @foo
  private x: number;

  constructor() {}
}
```
causes a crash due to a failed assertion about offsets.

With some debugging, I found that the resulting syntax tree when parsing this input is bogus.  Note how the AtToken shows up twice.

```
| | | | node: PropertyDeclaration
| | | | | node: AtToken
| | | | | node: SyntaxList
| | | | | | node: Decorator
| | | | | | | node: AtToken
| | | | | | | node: Identifier
| | | | | node: SyntaxList
| | | | | | node: PrivateKeyword
| | | | | node: Identifier
```

Upgrading our TypeScript dep to typescript@beta (1.8) produces

```
| | | | node: PropertyDeclaration
| | | | | node: SyntaxList
| | | | | | node: Decorator
| | | | | | | node: AtToken
| | | | | | | node: Identifier
| | | | | node: SyntaxList
| | | | | | node: PrivateKeyword
| | | | | node: Identifier
```

which is a layout I expect.

martine:
There were two issues:
1) TypeScript <1.8 has a busted parse of decorators (see bug for
   an example), so update to TS 1.8;
2) We stripped comments when emitting "private" to avoid a Closure
   issue but doing so also stripped preceding whitespace, which
   meant that in a line like "@foo private bar" there would be no
   space between "@foo" and "private" in the output.  Fix this by
   being more careful about emitting whitespace.

Fixes #35.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -222,7 +222,19 @@ class Annotator {
   private writeNode(node: ts.Node, skipComments: boolean = false) {
     if (node.getChildCount() == 0) {
       // Directly write complete tokens.
-      this.emit(skipComments ? node.getText() : node.getFullText());
+      if (skipComments) {
+        // To skip comments, we skip all whitespace/comments preceding
+        // the node.  But if there was anything skipped we should emit
+        // a newline in its place so that the node remains separated
+        // from the previous node.  TODO: don't skip anything here if
+        // there wasn't any comment.
+        if (node.getFullStart() < node.getStart()) {
+          this.emit('\n');
+        }
+        this.emit(node.getText());
+      } else {
+        this.emit(node.getFullText());
+      }
       return;
     }
     if (skipComments) {
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -12,6 +12,8 @@ const OPTIONS: ts.CompilerOptions = {
   noResolve: true,
   skipDefaultLibCheck: true,
   noEmitOnError: true,
+  experimentalDecorators: true,
+  emitDecoratorMetadata: true,
 };
 
 const {cachedLibPath, cachedLib} = (function() {
--- a/test_files/sickle/parameter_properties.ts
+++ b/test_files/sickle/parameter_properties.ts
@@ -1,6 +1,8 @@
 class ParamProps {
   // The @export below should not show up in the output ctor.
-  constructor(public bar: string,public bar2: string) {
+  constructor(
+public bar: string,
+public bar2: string) {
 
 // Sickle: begin stub declarations.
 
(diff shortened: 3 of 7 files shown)


test_files/sickle/parameter_properties.ts
```diff
@@ -1,6 +1,8 @@
 class ParamProps {
   // The @export below should not show up in the output ctor.
-  constructor(public bar: string,public bar2: string) {
+  constructor(
+public bar: string,
+public bar2: string) {
```
martine: I'm not super happy that sickle munges this but I don't care *too* much.

src/sickle.ts
```diff
@@ -222,7 +222,16 @@ class Annotator {
   private writeNode(node: ts.Node, skipComments: boolean = false) {
     if (node.getChildCount() == 0) {
       // Directly write complete tokens.
-      this.emit(skipComments ? node.getText() : node.getFullText());
+      if (skipComments) {
+        if (node.getFullStart() < node.getStart()) {
+          // Emit a space to replace the comment, as the comment served
```
rkirov: emit a newline

src/sickle.ts
```diff
@@ -222,7 +222,16 @@ class Annotator {
   private writeNode(node: ts.Node, skipComments: boolean = false) {
     if (node.getChildCount() == 0) {
       // Directly write complete tokens.
-      this.emit(skipComments ? node.getText() : node.getFullText());
+      if (skipComments) {
+        if (node.getFullStart() < node.getStart()) {
```
rkirov: Maybe extract into a helper function - `nodeHasAComment`, seems useful and not totally obvious from the node api

test/test_support.ts
```diff
@@ -12,6 +12,7 @@ const OPTIONS: ts.CompilerOptions = {
   noResolve: true,
   skipDefaultLibCheck: true,
   noEmitOnError: true,
+  experimentalDecorators: true,
```
rkirov: another option to add here is `emitDecoratorMetadata`, to match our internal TS flags.

test_files/sickle/parameter_properties.ts
rkirov: agree, we have bigger fish to fry :)

rkirov:
LGTM

src/sickle.ts
martine: Unfortunately this doesn't actually check for a comment, so it feels wrong to pull it into a function.  I will add a comment and a TODO.

test/test_support.ts
martine: It looks like that depends on this flag being also set.  I added it.

Issue #38, opened 2016-02-02
Synthetic ctor needs to call base class ctor with proper arguments


martine:
```javascript
class SuperTestBase {
  constructor(public x: number) {}
}
class SuperTestDerivedNoCTor extends SuperTestBase {
  foobar: number = 3;
}
```

Sickle generates a constructor for the derived class here, to have a place to declare the type of `foobar`.

But the existence of a constructor means that we need to call the constructor of the base class, which means we need to be able to reconstruct the constructor parameters as well as the parameters to super.

I think this means we might need to store a map of classes to the expected parameters of the ctor, which might even mean needing to do cross-file analysis... :~(

I tried a hack like this but it didn't go.  I might be able to turn off some type checking and force it though.
```javascript
constructor(...args: any[]) {
  super(...args);
```

mprobst:
`super.call(this, arguments);` ?

Evan Martin <notifications@github.com> schrieb am Di., 2. Feb. 2016 um
15:21 Uhr:

martine:
A good thought, but no:

http://www.typescriptlang.org/Playground#src=class%20SuperTestBase%20%7B%0A%20%20constructor(public%20x%3A%20number)%20%7B%7D%0A%7D%0Aclass%20SuperTestDerivedNoCTor%20extends%20SuperTestBase%20%7B%0A%20%20foobar%3A%20number%20%3D%203%3B%0A%20%20constructor()%20%7B%0A%09%20super(this%2C%20arguments)%3B%20%0A%20%20%7D%0A%7D

martine:
The Closure compiler might accept this:
```javascript
class Foo {
  /** @type {number} */
  y;
}
```

At least, it recognizes it as ES6 typed mode and when it encounters it (if you set the typed flag) it hits this code:
  private void recordTypeSyntax(SourceRange location) {
    if (hasJsDocTypeAnnotations) {
      errorReporter.error("Can only have JSDoc or inline type annotations, not both",
          sourceName, lineno(location.start), charno(location.start));
    }
  }

mprobst:
`ES6_TYPED` is the mode we hacked into JSCompiler when we thought we could just teach it TS (or something close enough). I don't think it's supported or particularly well working, so I'd recommend not relying on it.

martine:
Apparently this works!?

```js
class Foo {
  bar() {
    /** @type {number} */
    this.x = 3;
  }
}

let x = new Foo();
console.log(x.x);
```

[...]

PR #39, opened 2016-02-03
handle some more cases of subclassing needing super() calls

martine:
When we generate our own constructor() for a class (which we do
when we need to declare any properties on it), we must also put
in a call to super().  But this is easy to get wrong so do our
best to avoid calling super() when possible.

Helps with issue #38 but doesn't close it.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -20,6 +20,22 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
+/**
+ * Returns true if a class declaration has a superclass (as declared
+ * with the "extends" keyword), which indicates it must call super()
+ * in its constructor.
+ */
+function classHasSuperClass(classNode: ts.ClassDeclaration): boolean {
+  if (classNode.heritageClauses) {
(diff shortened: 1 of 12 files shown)


martine:
I'm not really happy with how subtle this is (and especially the test case is pretty crazy feeling) but I think it is sufficient to get missive working through sickle so I am sort of ok with it.

src/sickle.ts
```diff
@@ -20,6 +20,20 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
+// Return true if a class declaration has a superclass (as declared with
```
mprobst: Nit: use `/** jsdoc */`, prefer 3rd person voice for method comments (`Returns ...`)

mprobst:
LGTM. I think it's ok.

The problem is that we'll need type checking to really solve the `super` call part, which complicates things quite a bit. It'd be great if there was some Closure syntax that'd allow us to just invoke the super ctor with the correct args, but I don't think that exists.

src/sickle.ts

PR #41, opened 2016-02-03
annotate default arguments as optional

### LINKED ISSUE #40: Default param arguments need to be annotated as optional
```javascript
function foo(x: number, y: string = 'hi')
```

should produce something like

```javascript
function foo(/** number */ x, /** string= */ y = 'hi')
```

Note the equals sign on the 'y' arg.  This is because it's optional for the caller to provide it.

martine:
When a function accepts a default argument, in Closure-land we must
also annotate the type as optional (with the trailing '=' character).

Fixes #40.


rkirov:
LGTM

PR #44, opened 2016-02-03
map TS optional arguments to closure optional arguments

### LINKED ISSUE #43: Optional arguments need to propagate into Closure type
```javascript
function OptionalArgument(x: number, y?: string) {
```

currently does

```javascript
function OptionalArgument(/** number */ x, /** string */ y) {
```

but we need `y` to be `string=`.

martine:
E.g. (x?: number) becomes /** number= */ x.

Fixes #43.

### DIFF
--- a/test_files/es6/optional.js
+++ b/test_files/es6/optional.js
@@ -1,0 +1,2 @@
+function optionalArgument(/** number */ x, /** string= */ y) {
+}
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -131,7 +131,7 @@ class Annotator {
           for (let param of fnDecl.parameters) {
             this.writeTextFromOffset(writeOffset, param);
             writeOffset = param.getEnd();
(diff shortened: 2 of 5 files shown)


test_files/es6/optional.js
```diff
@@ -0,0 +1,2 @@
+function OptionalArgument(/** number */ x, /** string= */ y) {
```
mprobst: not sure if we care, but these should be `lowerCamelCase` in Closure style.

mprobst:
LGTM

test_files/es6/optional.js
martine: I guess it's up to the original author to do this though, right, so it's not sickle's problem?  Or do you just mean our test cases ought to match Closure style?

mprobst: I just meant the test case, this is a 100% pure small batch hand made nitpick :-)

PR #45, opened 2016-02-03
insert newlines into jsdoc comments when adding type annotations

### LINKED ISSUE #22: JSDoc comments need an extra newline
```js
class Foo {
  /** js-doc comment with text in it */
  frotz: number;
}
```

becomes

```js
class Foo {
  constructor() {
        /**  js-doc comment with text in it @type { number} */
        this.frotz;
    }
}
```

which is invalid Closure syntax -- we need a newline before that `@type` bit.

martine:
Also be more careful about spotting comments that are jsdoc comments.
Extend the comments test to include more variations of the forms of
jsdoc comments.

Fixes #22.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -196,7 +196,11 @@ class Annotator {
   }
 
   private visitProperty(p: ts.PropertyDeclaration | ts.ParameterDeclaration) {
-    this.maybeVisitType(p.type, this.existingClosureAnnotation(p) + '@type');
+    let existingAnnotation = this.existingClosureAnnotation(p).trim();
+    if (existingAnnotation) {
+      existingAnnotation += '\n';
+    }
+    this.maybeVisitType(p.type, existingAnnotation + '@type');
     this.emit('\nthis.');
     this.emit(p.name.getText());
     this.emit(';');
--- a/test_files/comments.ts
+++ b/test_files/comments.ts
@@ -1,8 +1,21 @@
-class A {
+class Comments {
   /** @export */
-  foo: string;
+  export1: string;
+  // Note: the below @export doesn't make it into the output because it
+  // it isn't in a JSDoc comment.
+  /// @export
+  export2: string;
   /* non js-doc comment */
-  bar: number;
+  nodoc1: number;
   // non js-doc comment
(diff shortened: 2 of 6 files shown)


src/sickle.ts
```diff
@@ -36,6 +36,11 @@ function classHasSuperClass(classNode: ts.ClassDeclaration): boolean {
   return false;
 }
 
+// trim removes leading and trailing whitespace from a string.
+function trim(str: string): string {
```
rkirov: Looks like ES5 has trim already, and I just checked that node v0.12.0 has it, so might as well use that.

src/sickle.ts
```diff
@@ -223,6 +232,9 @@ class Annotator {
     if (/^\/\*\*/.test(comment) && /\*\/$/.test(comment)) {
       return comment.slice(3, comment.length - 2);
     }
+    if (/^\/\/\//.test(comment)) {
```
rkirov: I am being super pedantic but according to http://usejsdoc.org/about-getting-started.html jsdoc must start with `/**`. Maybe rename method to `getComment`

test_files/comments.ts
```diff
@@ -1,8 +1,19 @@
...
+  export1: string;
+  /// @export
+  export2: string;
   /* non js-doc comment */
-  bar: number;
+  nodoc1: number;
   // non js-doc comment
-  buz: number;
+  nodoc2: number;
+  /// inline jsdoc comment without type annotation
```
rkirov: Aren't you worried that this would change the meaning of (arguably invalid) // @export (something that closure would ignore AFAIK) into valid /** @export @type */ . Feels like we would promote wrong style.

src/sickle.ts
martine: Ahhh I was being too clever here -- I don't think closure obeys them either.  I reduced the test and delete this.

src/sickle.ts

test_files/comments.ts
martine: I think I fixed this, let me know if I got it right.

rkirov:
LGTM

PR #48, opened 2016-02-04
Add @export for typescript enums

calebegg:


mprobst:
Not sure I get the context - why would we want to do that?

calebegg:
Discussed offline. Basically, this is the path-of-least-resistance to be able to use Foo.XYZ in our TS code instead of Foo['XYZ'].

src/sickle.ts
```diff
@@ -263,6 +267,17 @@ class Annotator {
     this.emit(': void;\n');
   }
 
+  private visitEnum(node: ts.EnumDeclaration) {
+    this.emit('\n');
+    for (let member of node.members) {
+      this.emit('/** @export */ ');
```
martine: Would it make sense to only @export these if the enum itself is @exported or something?  I'm imagining a hypothetical size-sensitive codebase where we haven't provided any knob to opt out of this.

calebegg: This shouldn't impact the size of the code at all. All this does is allow you to do `Foo.BAR` instead of having to do `Foo['BAR']`. The typescript enum won't get minified either way, because it uses `Foo['BAR']`.

test_files/sickle/enum.ts
```diff
@@ -0,0 +1,3 @@
+enum FooEnum {XYZ, PI = 3.14159}
+/** @export */ FooEnum.XYZ;
```
mprobst: Could we generate this? I think that should (a) tell Closure what's going on and (b) make sure the value is available after minification and (c) allow the constant (`XYZ`) to be renamed.

```
(<any>FooEnum).XYZ = 1;
(<any>FooEnum).PI = 3.14159;
/** @type {FooEnum} */
FooEnum.X;
/** @type {FooEnum} */
FooEnum.PI;
```

mprobst:
LGTM, thanks for the contribution!

Could you rebase and squash on master?

src/sickle.ts
```diff
@@ -263,6 +266,29 @@ class Annotator {
     this.emit(': void;\n');
   }
 
+  private visitEnum(node: ts.EnumDeclaration) {
+    this.emit('/** @typedef {number} */\n');
+    this.writeNode(node);
+    this.emit('\n');
+    let i = 0;
+    for (let member of node.members) {
+      this.emit('/** @type {');
```
mprobst: nit: consider using a ``template ${string}`` for some of this?

src/sickle.ts
```diff
@@ -263,6 +266,29 @@ class Annotator {
     this.emit(': void;\n');
   }
 
+  private visitEnum(node: ts.EnumDeclaration) {
+    this.emit('/** @typedef {number} */\n');
```
mprobst: One more thing: should this be `@typedef` or `@enum`? I'd prefer `@enum` if Closure swallows a non-initialized enum.

src/sickle.ts
calebegg: Done.

src/sickle.ts
mprobst: Mh. Seems not worth the trouble, Closure doesn't like assignment into the enum later on.

PR #49, opened 2016-02-08
don't emit for ambient module declarations

martine:
Skip over the contents of a "declare module ...", as it's only
making declaraitons for TypeScript's (not Closure's) benefit.

### DIFF
--- a/test_files/sickle/declare.ts
+++ b/test_files/sickle/declare.ts
@@ -1,0 +1,18 @@
+declare module DeclareTest {
+  // This enum should not cause any extra sickle statements to be
+  // emitted, because it is contained within a "declare" block.
+  enum FileType {
+    VALID = 1,
+  }
+}
+
+module DeclareTestIncluded {/** @typedef {number} */
+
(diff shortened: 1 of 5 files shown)


test_files/sickle/declare.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+module DeclareTestIncluded {/** @typedef {number} */
+
+  // This enum should produce statements, as it's not a "declare".
+  enum FileType {
+    VALID = 1,
+  }
+/** @type {FileType} */
+(<any>FileType).VALID =  1;
```
martine: What this change is addressing is that if we emit this bit of the processed enum in the above module (with the "declare"), it is a TypeScript syntax error because you cannot include statements within a "declare" block.

I'm not sure this solution is great.  :\

martine:
This doesn't work due to a different bug in enums.  I will ping on it once I expect it to work.

rkirov:
Ack. Let me know.

Issue #50, opened 2016-02-08
Export of enum typedef doesn't pass e2e test


martine:
If we generate code like the following, it fails to compile in Closure:

```javascript
/** @typedef {number} */
export var EnumTest2;
(function (EnumTest2) {
    EnumTest2[EnumTest2["XYZ"] = 0] = "XYZ";
    EnumTest2[EnumTest2["PI"] = 3.14159] = "PI";
})(EnumTest2 || (EnumTest2 = {}));
/** @type {EnumTest2} */
EnumTest2.XYZ = 0;  // <- ***fails here***
/** @type {EnumTest2} */
EnumTest2.PI = 3.14159;
```

The error message:

```
[...] ERROR - Bad type annotation. Unknown type EnumTest2
/** @type {EnumTest2} */
           ^
```

Something about export and typedef interacting poorly maybe?

The other weird thing is that I can't reproduce this with the Google-internal JS compiler, which makes me wonder if it's a bug.

martine:
(This is different than our existing "enum" test in that it uses `export enum ...`, which puts an extra `export` in the above output.)

PR #51, opened 2016-02-08
fix enum export by upgrading closure compiler

martine:
If you wrote "export enum ...", the generated ES6 JavaScript would
fail to compile in the Closure compiler.  I tracked this down to
a bug that was fixed in the upstream Closure compiler, so to fix
this I had to make us use a newer version of the upstream compiler.

To use a newer version of the upstream compiler, I removed one of
our indirection layers: we previously used the 'closure-compiler'
npm library which pulled in 'google-closure-compiler', but now the
latter module exposes a simple API so it's easy enough to just use
it directly.

Fixes issue #50.

### DIFF
--- a/test_files/enum.ts
+++ b/test_files/enum.ts
@@ -1,1 +1,4 @@
-enum FooEnum {XYZ, PI = 3.14159}
+enum EnumTest1 {XYZ, PI = 3.14159}
+
+// This additional exported enum is here to exercise the fix for issue #51.
+export enum EnumTest2 {XYZ, PI = 3.14159}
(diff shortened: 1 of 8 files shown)


test_files/enum.ts
```diff
@@ -1 +1,3 @@
-enum FooEnum {XYZ, PI = 3.14159}
\ No newline at end of file
+enum EnumTest1 {XYZ, PI = 3.14159}
+
+export enum EnumTest2 {XYZ, PI = 3.14159}
```
rkirov: OOC what difference does export make?

rkirov: Actually, maybe add a comment to explain why an exported enum needs separate testing.

Issue #52, opened 2016-02-09
"export enum" with comment is pretty hard to annotate


martine:
We try to emit a @typedef for enums when we hit them.
Below is the tree of the various nodes and where the emits happen when parsing something like

```
// comment here
export enum ...
```

Note that we emit the @typedef before we hit the ExportKeyword, which means it'll be hard to make the @typedef appear directly adjacent to the enum in the output.  I'm beginning to think stripping comments by default is a better strategy...

```
| | node: EnumDeclaration
| | | emit "/** @typedef {number} */\n"
| | | node: SyntaxList
| | | | node: ExportKeyword
| | | | | emit "\n\n// This additional exported enum is here to exercise the fix for issue #51.\nexport"
| | | node: EnumKeyword
| | | | emit " enum"
```

martine:
It turns out that the type with the comment in the middle works, and I was confused by two *other* bugs!  Closing.

PR #53, opened 2016-02-09
skip comments when exporting enums

martine:
When we export an enum we try to prepend an @typedef on it.
If the enum already had a comment on it, the comment interferes
with the type annotation.  Fix this by stripping comments when
emitting enums.  See the diff of es6/enum.js to see where this
is necessary.

This comment stripping is pretty hacky and will likely fail in
complex scenarios.  Right now it's limited to enums so maybe it
will work out?

See issue #52 for an example of why we can't do something obvious
to skip the comment -- the parse tree is pretty complex.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -67,7 +67,7 @@ class Annotator {
     console.log(prefix + message);
   }
 
-  private visit(node: ts.Node) {
+  private visit(node: ts.Node, skipComments: boolean = false) {
     this.currentNode = node;
     // this.logWithIndent('node: ' + (<any>ts).SyntaxKind[node.kind]);
     this.indent++;
@@ -154,7 +154,7 @@ class Annotator {
         this.visitEnum(<ts.EnumDeclaration>node);
         break;
       default:
-        this.writeNode(node);
+        this.writeNode(node, skipComments);
         break;
     }
     this.indent--;
@@ -268,7 +268,7 @@ class Annotator {
 
   private visitEnum(node: ts.EnumDeclaration) {
     this.emit('/** @typedef {number} */\n');
-    this.writeNode(node);
+    this.writeNode(node, /* skipComments */ true);
     this.emit('\n');
     let i = 0;
     for (let member of node.members) {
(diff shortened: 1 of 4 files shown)


src/sickle.ts
```diff
@@ -268,7 +268,7 @@ class Annotator {
 
   private visitEnum(node: ts.EnumDeclaration) {
```
rkirov: odd to have `visit` and `writeNode` with a skipComments flag, and not visitEnum. Can the flag be added to visitEnum and at callsite in visit - pass true.

src/sickle.ts
```diff
@@ -302,13 +307,13 @@ class Annotator {
       }
       return;
     }
-    if (skipComments) {
-      this.fail('skipComments unimplemented for complex Nodes');
-    }
     let lastEnd = node.getFullStart();
+    if (skipComments) lastEnd = node.getStart();
```
rkirov: I have a preference for single assignments (you can use const then too):
lastEnd = skipComments ? node.getStart() : node.getFullStart()

but really up to you.

test_files/es6/enum.js
```diff
@@ -9,7 +9,6 @@ EnumTest1.XYZ = 0;
 /** @type {EnumTest1} */
 EnumTest1.PI = 3.14159;
 /** @typedef {number} */
-// This additional exported enum is here to exercise the fix for issue #51.
```
rkirov: OOC, what did this do before your fix? Closure didn't take the type? This might be good example of why custom syntax for type declarations is preferable over comments.

rkirov:
LGTM, a few nitpicks in preparation for a whole stripComments pass, which would make our closure output easier.

src/sickle.ts
martine: visitEnum *must* skipComments to work.  I would prefer to only have the flag in the case where the caller might care to skip comments or not.

src/sickle.ts
martine: Done.  (But I can't use const here because lastEnd is updated within the loop.)

test_files/es6/enum.js
martine: I thought the type was ignored, but now that I'm making a smaller test case I see that I might be wrong!  I will revisit this once I confirm it.

PR #54, opened 2016-02-09
emit full jsdoc for function types

### LINKED ISSUE #43: Optional arguments need to propagate into Closure type

martine:
Rather than emitting a type per parameter, emit a jsdoc
before the function declaration.  This allows us to mark optional
arguments properly with the = suffix.  Note the changed "optional.ts"
test; it now verifies that Closure accepts a missing argument when
that argument was missing before.

Fixes #43.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -73,7 +73,7 @@ class Annotator {
     this.indent++;
     switch (node.kind) {
       case ts.SyntaxKind.VariableDeclaration:
-        this.maybeVisitType((<ts.VariableDeclaration>node).type);
+        this.maybeEmitJSDocType((<ts.VariableDeclaration>node).type);
         this.writeNode(node);
         break;
       case ts.SyntaxKind.ClassDeclaration: {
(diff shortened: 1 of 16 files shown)


test_files/es6/arrow_fn.js
```diff
@@ -1 +1 @@
-var fn3 = (/** number */ a) => 12;
+var fn3 = (a) => 12;
```
rkirov: looks like we lost the comment? But it appears in sickle/arrow_fn.ts, so I am very confused.

test_files/es6/decorator.js
```diff
@@ -7,7 +7,11 @@ var __decorate = (this && this.__decorate) || function (decorators, target, key,
 var __metadata = (this && this.__metadata) || function (k, v) {
     if (typeof Reflect === "object" && typeof Reflect.metadata === "function") return Reflect.metadata(k, v);
 };
-function decorator(/** Object */ a, /** string */ b) { }
+/**
+@param { Object} a
```
rkirov: usually these guys look like 

/** 
 * @param ...
 * @return 
 */
Notice the extra *. My guess is that closure doesn't care, but we should try to add them if it is not too hard.

src/sickle.ts
```diff
@@ -241,6 +246,14 @@ class Annotator {
...
       this.emit(' {');
     }
+    this.visitType(type);
+    if (jsDocTag) {
+      this.emit('}');
+    }
+    this.emit(' */');
+  }
+
+  private visitType(type: ts.TypeNode, optional?: boolean) {
```
rkirov: The difference between `maybeVisitType` and `visitType` is not very clear from the function name or implementation. One emits jsDoc annotations and the other doesn't. The mega-loop 'visit' calls into both, so we can't argue that one is an inner helper.

Maybe some renaming would help?

rkirov:
Should we file a bug at https://github.com/google/closure-compiler/issues for `function f(/** number= */)` not working? Sounds like a bug to me.

martine:
`function f(/** number= */)` is expected to not work.  Something about how `=` is not part of the type but actually part of the @param declaration.  E.g. I think you can't declare a variable with a type that includes `=` in it.

src/sickle.ts
martine: I renamed the functions to be `maybeEmitJSDocType` and `emitType`.  I prefer "emit" to "visit" because it's clearer that we're going to traverse the node for the purposes of writing something.  I agree the names are pretty confusing.

martine:
Note that the weird output with the duplicated comment is due to
https://github.com/Microsoft/TypeScript/issues/6982

martine:
Reassigning to Martin because Rado is out today.

martine:
Travis is failing due to the compile timing out.  :~(  I guess I gotta refactor things to run the compilations separately.  The compile succeeds locally though it takes a while.

martine:
With the timeout "fix" in place, this is now ready for review.

test_files/es6/arrow_fn.js
```diff
@@ -1 +1,10 @@
-var fn3 = (/** number */ a) => 12;
+var fn3 = 
+/**
```
rkirov: double emit?

rkirov:
I don't see how you fixed the timeout issue. But LGTM, since it looks fixed.

rkirov:
I thought `T=` is sugar for `T | undefined`, but really it is marker for the function type. The more your know ...

PR #56, opened 2016-02-09
update jscompiler test timeout to 20 seconds

martine:
We're now hitting the 10 second timeout on Travis.
Ugly workaround: wait up to 20 seconds.


mprobst:
Hurray! LGTM!

PR #58, opened 2016-02-10
preserve closure annotations when emitting functions

martine:
The change in 16a0f692312006e68a515d000389e80a08f4a7e1 caused
us to emit a JSDoc comment before functions, which means we
would drop any already existing JSDoc comment.

Fix this by merging existing JSDocs with our new comment.
Add a test in functions.ts to cover this.  (The other changed
tests are just due to changing comment emits.)


rkirov:
LGTM

PR #62, opened 2016-02-12
in untyped mode, don't emit type info on enums

### LINKED ISSUE #61: Support untyped mode in enums
When emitting an enum in untyped mode, we still want to emit some of the glue code we currently emit, but we shouldn't emit any of the types.

See this test output:
https://github.com/angular/sickle/blob/b4c9b8e32c2922626c83a69dbfbef5718272aeeb/test_files/sickle/enum.ts

We should preserve the (<any>EnumTest1).XYZ = 0; bit, but drop all the closure annotations.

martine:
Fixes #61.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -281,12 +281,12 @@ class Annotator {
   }
 
   private visitEnum(node: ts.EnumDeclaration) {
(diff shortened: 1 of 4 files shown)


rkirov:
LGTM

Issue #57, opened 2016-02-10
returning an arrow function triggers ASI


evmar:
``` javascript
function foo() {
    return (x: string) => 3;
}
```

produces

``` javascript
/**
 */
function foo() {
    return 
/**
 * @param { string} x
 */
(x: string) => 3;
}
```

Because we inserted the newline after the return, it triggers ASI and you get

```
error TS7027: Unreachable code detected.
```

martine:
For some reason this doesn't trigger when you return a function literal (non-arrow function).  I wonder if we fail to annotate types there intentionally?

martine:
I tried removing the whitespace around the comment, e.g.

```javascript
return /**
...
*/ (x: string) => 3;
```

but that still fails.

martine:
Idea 1: don't annotate arrow functions.  I don't think normal Closure users would annotate them anyway.

Hypothetical problem with that idea.  If you had code like the following:
```javascript
addCallback((obj: ComplexObject) => obj.foo());
```
and that was the only user of obj.foo, if we dropped the type of obj it may be the case that Closure will fail to mangle the obj.foo properly.

rkirov:
+1 for turning off jsdoc in this case, given that at this stage we are targeting outputting untyped.

Also note that closure can do type-inference at if addCallback is declared of type `@param {function(ComplexObject): void}`.

martine:
Rado suggested in the PR:

For typed mode I think we have to fix this proper, how about adding extra parens?

```javascript
function foo() {
    return (
/**
 * @param { string} x
 */
(x: string) => 3
);}
```

It does trigger the double emit bug, but at least it preserves the meaning of the original code.

evmar:
We now add parens.

PR #63, opened 2016-02-13
in untyped mode, emit no types near arrow functions

martine:
Emitting a type for arrow functions triggers ASI due to
the newlines in the JSDoc type annotation.  Fix this by just
skipping type annotations.

Works around issue #57.


martine:
Maybe I should just always skip arrow functions even in typed mode, 'cause we can't get the annotations right... what do you think?

rkirov:
LGTM, since we agreed to work towards turning off annotations for untyped, https://github.com/angular/sickle/issues/59 and this is going in that direction.

For typed mode I think we have to fix this proper, how about adding extra parens?

```javascript
function foo() {
    return (
/**
 * @param { string} x
 */
(x: string) => 3
);}
```

It does trigger the double emit bug, but at least it preserves the meaning of the original code.

PR #64, opened 2016-02-15
Refactor tests a bit

mprobst:
Fixes an oversight that disabled testing (uh oh), and moves test files a bit closer to each other for easier development.


test/sickle_test.ts
```diff
@@ -6,22 +6,22 @@ import {expect} from 'chai';
...
 describe('golden tests', () => {
-
   goldenTests().forEach((test) => {
     if (RUN_TESTS_MATCHING && !RUN_TESTS_MATCHING.exec(test.name)) {
       it.skip(test.name);
       return;
     }
     let options: SickleOptions = {};
-    if (/\.untyped\./.test(test.name)) {
+    if (/\.untyped\b/.test(test.name)) {
```
martine: Why not leave the `.untyped.` portion in the file name?

martine:
Sorry for the slow review.

Also for the bikeshed: I think it'd be better to use dots instead of underscores in the file name for the different modes.  In your change you have files named e.g. `fields_no_ctor_in.ts` where it's hard to spot which parts are the test name (fields_no_ctor) and which are metadata.  This is especially bad when there's an untyped in there as well.

mprobst:
Good point, I like your bike shed better. Done.

test/sickle_test.ts
martine: Undo this.

test/test_support.ts
```diff
@@ -85,15 +86,15 @@ export interface GoldenFileTest {
...
+    let testName = testPath.match(/\/test_files\/(.*)\.in\.ts$/)[1];
     return {
-      name: fn,
-      tsPath: path.join(testFolder, fn),
-      sicklePath: path.join(testFolder, 'sickle', fn),
-      es6Path: path.join(testFolder, 'es6', fn.replace(tsExtRe, '.js')),
+      name: testName,
+      tsPath: testPath,
+      sicklePath: testPath.replace(/\.in\.ts$/, '.sickle.ts'),
+      es6Path: testPath.replace(/\.in\.ts$/, '.tr.js'),
```
martine: What does "tr" mean.  Test result?  Just curious.

test/sickle_test.ts
mprobst: It's needed, the new regexp strips the `.in.ts` at the end of the test name. Which is admittedly mostly a side effect, but I find the test names a bit cleaner like that.

test/test_support.ts
mprobst: translated, and I thought the entire word is a bit too long. `out` sorts before `sickle`. I could rename `sickle` to `middle` or `intermediate` or so. I really don't think it matters too much as long as we have the files close to each other.

mprobst:
Merged in 51390f43467f65ec05372d6b8a3bd1ae8f1d2e1d.

PR #66, opened 2016-02-17
convert TypeScript type coercions to Closure ones

### LINKED ISSUE #65: Preserve TypeScript type coercions in Closure code
Input:
```javascript
function foo(str: string) {}
foo(<string>JSON.parse('"foo"'));
```

Outputs:
```javascript
// @param {string} str
function foo(str) { }
foo(JSON.parse('"foo"'));
```

We drop the `<string>` so Closure doesn't know about the type cast, producing a warning.

martine:
Fixes #65.

But this doesn't work!  See the comment in that bug about the TypeScript compiler bug...

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -170,6 +170,13 @@ class Annotator {
       case ts.SyntaxKind.EnumDeclaration:
         this.visitEnum(<ts.EnumDeclaration>node);
         break;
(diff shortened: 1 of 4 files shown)


mprobst:
LGTMs for the sickle change.

Maybe we should just fix the issue on the TypeScript side?

PR #70, opened 2016-02-24
don't crash on function overloads

martine:
Function overloads are written as function declarations with
empty bodies, so check that the body is present before emitting.


rkirov:
LGTM

PR #71, opened 2016-02-24
don't emit ctors on classes with only static members

martine:
Avoid emitting stub declarations on classes that only have static
members, as that helps us avoid getting calls to super() wrong.

More avoiding of issue #38.

### DIFF
--- a/test_files/super.tr.js
+++ b/test_files/super.tr.js
@@ -59,3 +59,6 @@ class SuperTestDerivedInterface {
         // Sickle: end stub declarations.
     }
 }
+class SuperTestStaticProp extends SuperTestBaseOneArg {
+}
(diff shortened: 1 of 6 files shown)


test_files/super.tr.js
```diff
@@ -59,3 +59,6 @@ class SuperTestDerivedInterface {
         // Sickle: end stub declarations.
     }
 }
+class SuperTestStaticProp extends SuperTestBaseOneArg {
+}
```
martine: This bit shows why this change is necessary.  It's important here that we don't create a synthetic ctor, because we don't know the argument to pass to the base class.

rkirov:
The less synthetic constructors the better. LGTM.

PR #73, opened 2016-02-25
don't drop typescript type assertion

martine:
When translating, e.g.
  <string>foo
we need to emit both the Closure *and* the TS type, e.g.
  /** string */(<string>foo)

My previous change was only emitting the Closure one.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -197,7 +197,7 @@ class Annotator {
         let typeAssertion = <ts.TypeAssertion>node;
         this.maybeEmitJSDocType(typeAssertion.type);
         this.emit('(');
-        this.visit(typeAssertion.expression);
+        this.writeNode(node);
         this.emit(')');
         break;
       default:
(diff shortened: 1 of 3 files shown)


test_files/coerce.tr.js
```diff
@@ -1 +1 @@
-let x = 'hello, ' + (JSON.parse('"world"'));
+let x = 'hello, ' + JSON.parse('"world"');
```
martine: Note that this output is still wrong (we lost the Closure type), but that is due to the TS bug where comments are dropped.

src/sickle.ts
```diff
@@ -197,7 +197,7 @@ class Annotator {
         let typeAssertion = <ts.TypeAssertion>node;
         this.maybeEmitJSDocType(typeAssertion.type);
         this.emit('(');
-        this.visit(typeAssertion.expression);
+        this.writeNode(node);
```
rkirov: A random check for similar patterns in other branches shows that we have - 'this.visit(fnDecl.body);' should that be also  this.writeNode(fnDecl.body);?

test_files/coerce.tr.js
rkirov: ack.

rkirov:
LGTM

Issue #74, opened 2016-02-25
"declare module" crashes


martine:
```javascript
declare module Foo { enum Bar { ...
```
after sickle we produce the TypeScript code

```javascript
declare module Foo {
  [...]
/** @type {Bar} */
(<any>Bar).BAZ =  1;
}
```

and you can't have statements in a `declare`d module.  I think the workaround is to skip them in this case but I don't really know what the original code is trying to do...

martine:
Worked around this.

PR #75, opened 2016-02-25
skip processing of ambient modules

martine:
Ambient modules contain declarations of external values,
and aren't of interest to sickle.  Skipping them means we
don't need to worry about e.g. a declared constructor that
has no body.

Works around issue #74.


martine:
(The test case passes in that the input compiles without crashing.)

rkirov:
There is an argument to be made for turning those into externs, but closure has no way of expressing inlined externs, so this is good for now.

PR #77, opened 2016-02-25
use a helper instead of a synthetic constructor

martine:
We need Closure to believe that all our member variables exist.
To do so, we need to tag them with an /** @type */ bit.  But it
turns out you don't need to do this in the constructor, it can
be done anywhere!

Doing this annotation in a unused extra method means that we
don't need to create constructors, which means we don't need
to worry about calling super() correctly, completely avoiding
issue #38.

Because the extra method is unused, Closure ought to eliminate
it when you compile, too, so it costs nothing.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -20,35 +20,6 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
-/**
- * Returns true if a class declaration has a superclass (as declared
- * with the "extends" keyword), which indicates it must call super()
- * in its constructor.
- */
-function classHasSuperClass(classNode: ts.ClassDeclaration): boolean {
-  if (classNode.heritageClauses) {
-    for (let heritage of classNode.heritageClauses) {
-      if (heritage.token == ts.SyntaxKind.ExtendsKeyword) {
-        return true;
-      }
-    }
-  }
-  return false;
-}
-
-/**
- * Gathers all non-static properties on a class declaration.  These are
- * the properties that should be mentioned in the synthetic constructor
- * for Closure's benefit.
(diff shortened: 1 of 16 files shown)


test_files/fields.in.ts
```diff
@@ -7,7 +7,11 @@ class Klass {
...
   getF1() {
-    // This access print a warning without a generated field stub declaration.
+    // This access prints a warning without a generated field stub declaration.
     return this.field1;
   }
 }
+
+let fieldsTest = new FieldsTest(3);
+// Ensure the type is understood by Closure.
+fieldsTest.field1 = 'hi';
```
martine: This right here fails to compile if Closure doesn't believe that "field1" is a string, so by including it in the test here we get coverage of the Closure check that the field exists.

src/sickle.ts
```diff
@@ -217,39 +183,37 @@ class Annotator {
...
-    this.emitStubDeclarations(classNode, nonStaticProps, []);
-    this.emit('}\n');
-  }
 
-  private emitStubDeclarations(
-      classDecl: ts.ClassLikeDeclaration, nonStaticProps: ts.PropertyDeclaration[],
-      paramProps: ts.ParameterDeclaration[]) {
-    this.emit('\n\n// Sickle: begin stub declarations.\n');
-    this.emit('\n');
+    this.emit('\n_typeAnnotationsHelper() {\n');
```
mprobst: nit: call it `_sickle_typeAnnotationsHelper`?

mprobst:
LGTM.

Kind of scary that this works, but I think we're flying well within what's allowed by Closure.

src/sickle.ts

PR #78, opened 2016-02-26
make the type annotations helper static

martine:
For fields on a class, we emit a helper function that mentions
all the fields for Closure's sake.  But the helper function cannot
be a method, as that affects the type of the generated class (it
has the additional method).

Fix this by making the helper static.
See the new "structural.untyped" test for a verification of this fix.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -211,22 +211,19 @@ class Annotator {
       return;
     }
 
-    this.emit('\n_sickle_typeAnnotationsHelper() {\n');
-    nonStaticProps.forEach((p) => this.visitProperty(p));
-    paramProps.forEach((p) => this.visitProperty(p));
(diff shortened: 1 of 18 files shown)


src/sickle.ts
```diff
@@ -211,22 +211,19 @@ class Annotator {
       return;
     }
 
-    this.emit('\n_sickle_typeAnnotationsHelper() {\n');
-    nonStaticProps.forEach((p) => this.visitProperty(p));
-    paramProps.forEach((p) => this.visitProperty(p));
+    this.emit('\nstatic _sickle_typeAnnotationsHelper() {\n');
```
rkirov: add two spaces after \n so my eyes don't bleed looking at the tests :)

rkirov:
LGTM

PR #80, opened 2016-03-02
annotate class methods and constructors

### LINKED ISSUE #68: Annotate class methods
```javascript
class Foo {
  bar(x: string, y?: string) {}
}
```
generates
```javascript
class Foo {
  bar(x: string, y?: string) {}
// Sickle: begin synthetic ctor.
constructor() {
[... more elided...]
```

We don't annotate the function for some reason?  We need to do so for the optional argument.

martine:
Fixes #68.

### DIFF
--- a/test_files/optional.in.ts
+++ b/test_files/optional.in.ts
@@ -1,3 +1,11 @@
 function optionalArgument(x: number, y?: string) {
 }
 optionalArgument(1);
+
+class OptionalTest {
+  constructor(a: string, b?: string) {}
+  method(c: string = 'hi') {}
+}
+
+let optionalTest = new OptionalTest('a');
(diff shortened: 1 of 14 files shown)


test_files/optional.in.ts
```diff
@@ -1,3 +1,11 @@
...
 }
 optionalArgument(1);
+
+class OptionalTest {
+  constructor(a: string, b?: string) {}
+  method(c: string = 'hi') {}
+}
+
+let optionalTest = new OptionalTest('a');
+optionalTest.method();
```
martine: In particular, this new bit of test code passes with the change in place.

mprobst:
LGTM.

At some point we should consider running clang-format on our output.

PR #81, opened 2016-03-03
translate "declare interface" as "@export all fields"

martine:
To do so, refactor the annotations helper logic so that
it can be used from multiple places where the fields have
lots of different incompatible types.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -22,6 +22,17 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
 
 const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.NodeFlags.Public;
 
+// TypedProperty represents a property on a class we'd like Closure to
+// know about.  We gather these up from multiple sources (field
+// declarations, constructor parameters) and emit them all with type
+// annotations in a function on the side (see emitPropertyDeclarations).
+interface TypedProperty {
+  name: string;
+  type: ts.TypeNode;
+  // extraTags holds extra Closure tags, such as @export.
+  extraTags: string;
+}
+
 /**
  * A source processor that takes TypeScript code and annotates the output with Closure-style JSDoc
  * comments.
@@ -64,6 +75,40 @@ class Annotator {
           break;
         } else {
           this.writeNode(node);
+        }
+        break;
+      case ts.SyntaxKind.InterfaceDeclaration:
+        this.writeRange(node.getFullStart(), node.getEnd());
+        if (node.flags & ts.NodeFlags.Ambient) {
+          // It's a "declare interface ..."; write a Closure interface
+          // declaration so we can @export all properties.
+          let decl = <ts.InterfaceDeclaration>node;
+          let className = decl.name.getText();
+          let props: TypedProperty[] = [];
+          for (let member of decl.members) {
+            // Members can include things like index signatures, for e.g.
+            //   interface Foo { [key: string]: number; }
+            // For now, just die unless all the members are regular old
+            // properties.
+            if (member.kind != ts.SyntaxKind.PropertySignature) {
+              this.fail(`unhandled member of kind ${ts.SyntaxKind[member.kind]}`);
+            }
+            let prop = <ts.PropertySignature>member;
+            props.push({
+              name: prop.name.getText(),
+              type: prop.type,
+              extraTags: '@export',
+            });
+          }
+          // Note: be careful with this code!  It's easy to make something
+          // that Closure will accept but then also have it silently rename
+          // fields you intended to @export.
+          // @record tells Closure that this type is structural (not sure
+          // how well that works), while @struct makes it complain if you
+          // accidentally add a field that isn't declared.
+          this.emit('\n/** @record @struct */\n');
+          this.emit(`function ${className}() {}\n`);
+          this.emitPropertyDeclarations(className, props);
         }
         break;
       case ts.SyntaxKind.VariableDeclaration:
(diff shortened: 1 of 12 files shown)


test_files/declare.sickle.ts
```diff
@@ -1,5 +1,17 @@
-declare module DeclareTest {
+declare module DeclareTestModule {
   export class Foo {
     constructor();
   }
 }
+
+declare interface DeclareTestInterface {
```
martine: This is the test for the new "declare" code.  The rest of the changes are whitespace changes due to the refactor.

src/sickle.ts
```diff
@@ -66,6 +77,34 @@ class Annotator {
...
+          // declaration so we can @export all properties.
+          let decl = <ts.InterfaceDeclaration>node;
+          let className = decl.name.getText();
+          let props: TypedProperty[] = [];
+          for (let member of decl.members) {
+            // Members can include things like index signatures, for e.g.
+            //   interface Foo { [key: string]: number; }
+            // For now, just die unless all the members are regular old
+            // properties.
+            if (member.kind != ts.SyntaxKind.PropertySignature) {
```
mprobst: Would this be `@dict` at some point?

src/sickle.ts
```diff
@@ -66,6 +77,34 @@ class Annotator {
...
+              this.fail(`unhandled member of kind ${ts.SyntaxKind[member.kind]}`);
+            }
+            let prop = <ts.PropertySignature>member;
+            props.push({
+              name: prop.name.getText(),
```
mprobst: You're always pushing the same signature for a `prop`, only varying the `extraTags`. Maybe use a ctor and a class for this? `props.push(new TypedProperty(prop, '@export'));`

src/sickle.ts
```diff
@@ -197,39 +236,56 @@ class Annotator {
...
+          let prop = <ts.PropertyDeclaration>member;
+          if (!(member.flags & ts.NodeFlags.Static)) {
+            props.push({
+              name: prop.name.getText(),
+              type: prop.type,
+              extraTags: this.existingClosureAnnotation(prop),
+            });
+          }
+          break;
+      }
```
mprobst: nit: Google style is to always have a default case, even if it's just `break;`

src/sickle.ts
```diff
@@ -66,6 +77,34 @@ class Annotator {
...
+              this.fail(`unhandled member of kind ${ts.SyntaxKind[member.kind]}`);
+            }
+            let prop = <ts.PropertySignature>member;
+            props.push({
+              name: prop.name.getText(),
+              type: prop.type,
+              extraTags: '@export',
+            });
+          }
+          this.emit(`\nclass ${className} {\n`);
```
mprobst: Generating a class here is quite surprising - this needs a comment, IMHO.

test_files/declare.sickle.ts
mprobst: This is pretty far away from what a Closure user would write. I think it'd be more idiomatic to at least use an @interface with @record here, but I guess we cannot do that?

I'm fine with this, but I guess at some point we need infrastructure that let's us generate truly arbitrary JS code. If only in an "overflow buffer" that we later append to the final .js file before writing it.

mprobst:
Some nits, but LGTM overall.

test_files/declare.sickle.ts
martine: It also works to use an @interface here, but I was afraid of that because I don't know the full consequences of all these crazy Closure attributes.  (Like I think you meant @struct, not @record ?)

src/sickle.ts
martine: Possibly?  I think especially in the case where you're mixing index signatures and properties the mapping to Closure gets funky.

src/sickle.ts
martine: The prop is actually different in each: one is a PropertySignature, one is a ParameterDeclaration, and one is a PropertyDeclaration.  They don't have a shared supertype and they don't have a shared type for their "name" field (e.g. you can have class properties with literal expressions for their names, whereas constructors can have a BindingPattern for their name field).

Re a class in general, I prefer the record because it gives named arguments (three arguments is dangerously close to being hard to track the proper arg order).

src/sickle.ts
martine: Oh weird, I'll add it.  Do you know why?

src/sickle.ts
martine: A bit more on the class thing: when I use a class, Closure figures out all the mapping of "class Foo" to its @interface/@struct/etc stuff.  And I can just emit the ES6-style annotations helper as a static method.  If I don't emit a class, I have to create an interface type and I have to figure out where to stuff the declarations of the fields.  (I guess I put them in the ctor?  I will try it.)

martine:
I pushed a new version that uses @record @struct on a function with properties, rather than an ES6 class.

src/sickle.ts
mprobst: We'll have to think about that at some point, but it's probably worth tackling that together with other type system matching issues, such as structural vs nominal typing.

test_files/declare.sickle.ts
mprobst: `@record` on an `@interface` should make it structurally typed, `@struct` is a good idea, too.

If this works, it's fine. I remember some discussion about how Closure only does structural optimization if it finds code that flows into an `@record` annotated type. We currently don't do anything like that (no type annotations in the output code, just `?`), so let's not go there.

src/sickle.ts
```diff
@@ -66,6 +77,40 @@ class Annotator {
...
+              extraTags: '@export',
+            });
+          }
+          // Note: be careful with this code!  It's easy to make something
+          // that Closure will accept but then also have it silently rename
+          // fields you intended to @export.
+          // @record tells Closure that this type is structural (not sure
+          // how well that works), while @struct makes it complain if you
+          // accidentally add a field that isn't declared.
+          this.emit('\n/** @record @struct */\n');
```
mprobst: Would it make sense to insert a comment into the output, explaining to the baffled debugger what this is doing here?

martine:
After talking this through with Martin, I'm going to abandon this particular approach (@export).  I might use code like this for supporting plain interfaces though.

Issue #67, opened 2016-02-18
Don't create redundant param declarations when original source already has jsdoc


martine:
Input:
```javascript
/**
 * @param x
 */
function foo(x: string) {}
```
currently generates
```javascript
/**
 * * @param x
 * @param { string} x
 */
function foo(x: string) {}
```

which fails with an error like "Bad type annotation. duplicate variable name".

PR #84, opened 2016-03-04
parse jsdoc to handle @param/@return properly

martine:
When emitting a function, we want to preserve any JSDoc describing
the parameters but supply our own values for the types.

TypeScript has a JSDoc parser, but it's not exposed in a place
where we can use it.  So for now do a hacky regexp-based parse
of JSDoc.  The interfaces used here are pretty similar to the
TypeScript ones so if the API does expose them, it shouldn't be
too hard to adjust.

Fixes issue #67.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -20,6 +20,75 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
+/**
+ * TypeScript has an API for JSDoc already, but it's not exposed.
+ * https://github.com/Microsoft/TypeScript/issues/7393
+ * For now we create types that are similar to theirs so that migrating
+ * to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
+ */
+export interface JSDocTag {
+  // tagName is e.g. "param" in an @param declaration.  It's absent
+  // for the plain text documentation that occurs before any @foo lines.
+  tagName?: string;
+  // parameterName is the the name of the function parameter, e.g. "foo"
+  // in
+  //   @param foo The foo param.
+  parameterName?: string;
+  type?: ts.TypeNode;
+  // optional is true for optional function parameters.
+  optional?: boolean;
+  text?: string;
+}
+
+export interface JSDocComment { tags: JSDocTag[]; }
+
+/**
+ * getJSDocAnnotation parses JSDoc out of a comment string.
+ * Returns null if comment is not JSDoc.
+ */
+export function getJSDocAnnotation(comment: string): JSDocComment {
+  // TODO(evanm): this is a pile of hacky regexes for now, because we
+  // would rather use the better TypeScript implementation of JSDoc
+  // parsing.  https://github.com/Microsoft/TypeScript/issues/7393
+  let match = comment.match(/^\/\*\*([\s\S]*?)\*\/$/);
+  if (!match) return null;
+  comment = match[1].trim();
+  // Strip all the " * " bits from the front of each line.
+  comment = comment.replace(/^\s*\* /gm, '');
+  let lines = comment.split('\n');
+  let tags: JSDocTag[] = [];
+  for (let line of lines) {
+    let match = line.match(/^@(\S+) *(.*)/);
+    if (match) {
+      let [_, tagName, text] = match;
+      if (text[0] == '{') {
+        throw new Error('type annotations (using {...}) are not allowed');
+      }
+
+      // Grab the parameter name from @param tags.
+      let parameterName: string;
+      if (tagName === 'param') {
+        let match = text.match(/^(\S+) ?(.*)/);
+        if (match) [_, parameterName, text] = match;
+      }
+
+      let tag: JSDocTag = {tagName};
+      if (parameterName) tag.parameterName = parameterName;
+      if (text) tag.text = text;
+      tags.push(tag);
+    } else {
+      // Text without a preceding @tag on it is either the plain text
+      // documentation or a continuation of a previous tag.
+      if (tags.length === 0) {
+        tags.push({text: line.trim()});
+      } else {
+        tags[tags.length - 1].text += ' ' + line.trim();
+      }
+    }
+  }
+  return {tags};
+}
+
 const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.NodeFlags.Public;
 
 /**
@@ -53,7 +122,7 @@ class Annotator {
 
   private visit(node: ts.Node) {
     this.currentNode = node;
-    // this.logWithIndent('node: ' + (<any>ts).SyntaxKind[node.kind]);
+    // this.logWithIndent('node: ' + ts.SyntaxKind[node.kind]);
     this.indent++;
     switch (node.kind) {
       case ts.SyntaxKind.ModuleDeclaration:
@@ -159,32 +228,78 @@ class Annotator {
   }
 
   private emitFunctionType(fnDecl: ts.FunctionLikeDeclaration) {
+    // Construct the JSDoc comment by reading the existing JSDoc, if
+    // any, and merging it with the known types of the function
+    // parameters and return type.
+    let jsDoc = this.getJSDoc(fnDecl) || {tags: []};
+    let newDoc: JSDocComment = {tags: []};
+
+    // Copy all the tags other than @param/@return into the new
+    // comment without any change; @param/@return are handled later.
+    for (let tag of jsDoc.tags) {
+      if (tag.tagName === 'param' || tag.tagName === 'return') continue;
+      newDoc.tags.push(tag);
+    }
+
+    // Parameters.
+    if (fnDecl.parameters.length) {
+      for (let param of fnDecl.parameters) {
+        let name = param.name.getText();
+        let paramDoc: string;
+        // Search for this parameter in the JSDoc @params.
+        for (let { tagName, parameterName, text } of jsDoc.tags) {
+          if (tagName === 'param' && parameterName === name) {
+            paramDoc = text;
+            break;
+          }
+        }
+        newDoc.tags.push({
+          tagName: 'param',
+          parameterName: name,
+          type: param.type,
+          optional: param.initializer != null || param.questionToken != null,
+          text: paramDoc,
+        });
+      }
+    }
+
+    // Return type.
+    if (fnDecl.type) {
+      let returnDoc: string;
+      for (let { tagName, text } of jsDoc.tags) {
+        if (tagName === 'return') {
+          returnDoc = text;
+          break;
+        }
+      }
+      newDoc.tags.push({
+        tagName: 'return',
+        type: fnDecl.type,
+        text: returnDoc,
+      });
+    }
+
     // The first \n makes the output sometimes uglier than necessary,
     // but it's needed to work around
     // https://github.com/Microsoft/TypeScript/issues/6982
     this.emit('\n/**\n');
-    let existingAnnotation = this.existingClosureAnnotation(fnDecl);
-    if (existingAnnotation) {
-      this.emit(' * ' + existingAnnotation + '\n');
-    }
-    // Parameters.
-    if (fnDecl.parameters.length) {
-      for (let param of fnDecl.parameters) {
-        if (param.type) {
-          let optional = param.initializer != null || param.questionToken != null;
-          this.emit(' * @param {');
-          this.emitType(param.type, optional);
-          this.emit('} ');
-          this.writeNode(param.name);
-          this.emit('\n');
-        }
+    for (let tag of newDoc.tags) {
+      this.emit(' * ');
+      if (tag.tagName) {
+        this.emit(`@${tag.tagName}`);
       }
-    }
-    // Return type.
-    if (fnDecl.type) {
-      this.emit(' * @return {');
-      this.emitType(fnDecl.type);
-      this.emit('}\n');
+      if (tag.type) {
+        this.emit(' {');
+        this.emitType(tag.type, tag.optional);
+        this.emit('}');
+      }
+      if (tag.parameterName) {
+        this.emit(' ' + tag.parameterName);
+      }
+      if (tag.text) {
+        this.emit(' ' + tag.text);
+      }
+      this.emit('\n');
     }
     this.emit(' */\n');
   }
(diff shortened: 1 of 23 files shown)


src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+// TypeScript has an API for JSDoc already, but it's not exposed.
+// https://github.com/Microsoft/TypeScript/issues/7393
+// For now we create types that are similar to theirs so that migrating
+// to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
+export interface JSDocTag {
+  tagName?: string;
+  parameterName?: string;
+  text: string;
+}
+export interface JSDocComment { tags: JSDocTag[]; }
```
mprobst: nit: empty line above?

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
       .join('\n');
 }
 
+// TypeScript has an API for JSDoc already, but it's not exposed.
+// https://github.com/Microsoft/TypeScript/issues/7393
+// For now we create types that are similar to theirs so that migrating
+// to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
+export interface JSDocTag {
```
mprobst: Please use JSDoc?

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+}
+export interface JSDocComment { tags: JSDocTag[]; }
+
+// getJSDocAnnotation parses JSDoc out of a comment string.
+// Returns null if comment is not JSDoc.
+export function getJSDocAnnotation(comment: string): JSDocComment {
+  // TODO(evanm): this is a pile of hacky regexes for now, because we
+  // would rather use the better TypeScript implementation of JSDoc
+  // parsing.  https://github.com/Microsoft/TypeScript/issues/7393
+  let match = comment.match(/^\/\*\*([\s\S]*)\*\/$/);
```
mprobst: why not a dotall regexp?

mprobst: this is definitely just a single comment, no need to parse e.g. non-greedy?

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+  if (!match) return null;
+  comment = match[1].trim();
+  // Strip all the " * " bits from the front of each line.
+  comment = comment.replace(/^\s*\* /gm, '');
+  let lines = comment.split('\n');
+  let tags: JSDocTag[] = [];
+  for (let line of lines) {
+    let match = line.match(/^@(\S+) *(.*)/);
```
mprobst: you have two variables called `match`, matching different things. I don't feel strongly about it, but maybe `lineMatch`?

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+    if (match) {
+      let [_, tagName, text] = match;
+      // Attempt to discard any type annotation from the text.
+      // This regex is not correct but users shouldn't be providing
+      // types anyway.  TODO(evanm): consider making this an error.
```
mprobst: I'd support making it an error if we're sure enough this is precise. Which I think it would be, AFAICT.

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+      }
+
+      let tag: JSDocTag = {tagName, text};
+      if (parameterName) tag.parameterName = parameterName;
+      tags.push(tag);
+    } else {
+      // Text without a preceding @tag on it is either the plain text
+      // documentation or a continuation of a previous tag.
+      if (tags.length == 0) {
+        tags.push({text: line.trim()});
```
mprobst: I think it'd be a lot cleaner if you just had a "text" field in `JSDocComment`

src/sickle.ts
```diff
@@ -20,6 +20,62 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+        if (match) [_, parameterName, text] = match;
+      }
+
+      let tag: JSDocTag = {tagName, text};
+      if (parameterName) tag.parameterName = parameterName;
+      tags.push(tag);
+    } else {
+      // Text without a preceding @tag on it is either the plain text
+      // documentation or a continuation of a previous tag.
+      if (tags.length == 0) {
```
mprobst: use `===`

src/sickle.ts
```diff
@@ -53,7 +109,7 @@ class Annotator {
 
   private visit(node: ts.Node) {
     this.currentNode = node;
-    // this.logWithIndent('node: ' + (<any>ts).SyntaxKind[node.kind]);
+    // this.logWithIndent('node: ' + ts.SyntaxKind[node.kind]);
```
mprobst: intentional change?

src/sickle.ts
```diff
@@ -163,28 +219,66 @@ class Annotator {
...
     }
     // Parameters.
     if (fnDecl.parameters.length) {
       for (let param of fnDecl.parameters) {
+        let paramDoc: string;
+        if (jsDoc) {
+          // Search for this parameter in the JSDoc @params.
+          let name = param.name.getText();
+          for (let { tagName, parameterName, text } of jsDoc.tags) {
+            if (tagName == 'param' && parameterName == name) {
```
mprobst: `===`

src/sickle.ts
```diff
@@ -163,28 +219,66 @@ class Annotator {
...
+              paramDoc = text;
+              break;
+            }
+          }
+        }
         if (param.type) {
```
mprobst: Now you're dropping the docs if there's no type on the param...

What about formulating this as building up the JSDoc by taking existing JSDoc, then adding the type to the @param tag as needed (and adding empty docs for non-documented params), and then emitting the entire JSDoc?

src/sickle.ts
```diff
@@ -224,35 +318,38 @@ class Annotator {
...
   private visitProperty(className: string, p: ts.PropertyDeclaration | ts.ParameterDeclaration) {
-    let existingAnnotation = this.existingClosureAnnotation(p);
-    if (existingAnnotation) {
-      existingAnnotation += '\n';
+    let jsDoc = this.getJSDoc(p);
+    let existingAnnotation = '';
+    if (jsDoc) {
+      for (let { tagName, text } of jsDoc.tags) {
+        if (tagName == 'type') {
+          // TODO(evanm): warn here?
```
mprobst: error?

src/sickle.ts
```diff
@@ -224,35 +318,38 @@ class Annotator {
...
+          continue;
+        }
+        if (tagName) {
+          existingAnnotation += `@${tagName}\n`;
```
mprobst: what if this has tag and text?

src/sickle.ts
```diff
@@ -224,35 +318,38 @@ class Annotator {
...
+        }
+      }
     }
     this.maybeEmitJSDocType(p.type, existingAnnotation + '@type');
     this.emit(`\n    ${className}.prototype.${p.name.getText()};\n`);
   }
 
   /**
-   * Returns empty string if there is no existing annotation.
+   * Returns empty null if there is no existing comment.
```
mprobst: empty null?

test/sickle_test.ts
```diff
@@ -50,3 +50,37 @@ describe('golden tests', () => {
...
+describe('getJSDocAnnotation', () => {
+  it('does not get non-jsdoc values', () => {
+    let source = '/* ordinary comment */';
+    expect(getJSDocAnnotation(source)).to.equal(null);
+  });
+  it('grabs plain text from jsdoc', () => {
+    let source = '/** jsdoc comment */';
+    expect(getJSDocAnnotation(source)).to.deep.equal({
+      tags: [
+        {text: 'jsdoc comment'},
```
mprobst: drop the comma, and it'll be on one line

src/sickle.ts
martine: by dotall do you mean /.*/ ?  that doesn't match newlines

this is definitely a single comment, yeah

src/sickle.ts
martine: There's actually 3 match variables.  Each one is replacing the one before it, and at least they're all of the same type?

src/sickle.ts
mprobst: Sorry, I slipped in the language. I meant this here:
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#DOTALL

... which doesn't exist in JS.

src/sickle.ts
mprobst: OK

martine:
PTAL.  I rewrote it to your suggested pattern of "merge existing JSDoc with known types into a data structure, then emit".  I think it is much nicer now.

src/sickle.ts
mprobst: Ping on making type annotations an error?

src/sickle.ts
```diff
@@ -20,6 +20,70 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
...
+ */
+export interface JSDocTag {
+  tagName?: string;
+  parameterName?: string;
+  type?: ts.TypeNode;
+  optional?: boolean;
+  text?: string;
+}
+
+export interface JSDocComment { tags: JSDocTag[]; }
```
mprobst: Document that the main comment is expressed as a tag without a tag name?

src/sickle.ts
```diff
@@ -159,32 +223,83 @@ class Annotator {
...
-        if (param.type) {
-          let optional = param.initializer != null || param.questionToken != null;
-          this.emit(' * @param {');
-          this.emitType(param.type, optional);
-          this.emit('} ');
-          this.writeNode(param.name);
-          this.emit('\n');
+        let name = param.name.getText();
+        let paramDoc: string;
+        if (jsDoc) {
```
mprobst: shouldn't you make this always defined?

src/sickle.ts
```diff
@@ -159,32 +223,83 @@ class Annotator {
...
-    // https://github.com/Microsoft/TypeScript/issues/6982
-    this.emit('\n/**\n');
-    let existingAnnotation = this.existingClosureAnnotation(fnDecl);
-    if (existingAnnotation) {
-      this.emit(' * ' + existingAnnotation + '\n');
+    // Construct the JSDoc comment by reading the existing JSDoc, if
+    // any, and merging it with the known types of the function
+    // parameters and return type.
+    let jsDoc = this.getJSDoc(fnDecl);
+    let newDoc: JSDocComment = {tags: []};
```
mprobst: why the distinction of `newDoc` and `jsDoc`? You could just augment `jsDoc`, right?

src/sickle.ts
```diff
@@ -224,35 +339,37 @@ class Annotator {
...
   private visitProperty(className: string, p: ts.PropertyDeclaration | ts.ParameterDeclaration) {
-    let existingAnnotation = this.existingClosureAnnotation(p);
-    if (existingAnnotation) {
-      existingAnnotation += '\n';
+    let jsDoc = this.getJSDoc(p);
+    let existingAnnotation = '';
+    if (jsDoc) {
+      for (let { tagName, text } of jsDoc.tags) {
+        if (tagName === 'type') {
+          this.fail('@type not allowed in property declarations');
```
mprobst: We should prioritize bringing proper error reporting online (multiple errors, with locations) now that we add features like this.

test/sickle_test.ts
```diff
@@ -50,3 +50,33 @@ describe('golden tests', () => {
     });
   });
 });
+
+describe('getJSDocAnnotation', () => {
+  it('does not get non-jsdoc values', () => {
```
mprobst: Add a test case for the failure for `@type` params?

src/sickle.ts

src/sickle.ts
martine: I need to shuffle together @params from two sources: there are some (but not necessarily all) in the existing jsdoc, and the source of truth comes from TypeScript type info.

If I wanted to only augment the existing jsdoc I'd need to find all the @param tags in it, if any, then find the right places to insert the params known by TypeScript amongst them.  I think it's simpler code to create a new data structure.

src/sickle.ts
martine: Fixed.

test/sickle_test.ts
martine: Done.

src/sickle.ts
mprobst: Sure, your call.

src/sickle.ts
```diff
@@ -159,32 +228,80 @@ class Annotator {
...
       }
     }
+
     // Return type.
     if (fnDecl.type) {
-      this.emit(' * @return {');
-      this.emitType(fnDecl.type);
-      this.emit('}\n');
+      let returnDoc: string;
+      if (jsDoc) {
```
mprobst: more if jsdoc

src/sickle.ts
```diff
@@ -224,35 +341,41 @@ class Annotator {
   }
 
   private visitProperty(className: string, p: ts.PropertyDeclaration | ts.ParameterDeclaration) {
-    let existingAnnotation = this.existingClosureAnnotation(p);
-    if (existingAnnotation) {
-      existingAnnotation += '\n';
+    let jsDoc = this.getJSDoc(p);
+    let existingAnnotation = '';
+    if (jsDoc) {
```
mprobst: more if jsdoc

src/sickle.ts
```diff
@@ -224,35 +341,41 @@ class Annotator {
...
-  }
-
-  // return empty string if comment is not JsDoc.
-  static getJsDocAnnotation(comment: string): string {
-    if (/^\/\*\*/.test(comment) && /\*\/$/.test(comment)) {
-      return comment.slice(3, comment.length - 2);
+    try {
+      return getJSDocAnnotation(comment);
+    } catch (e) {
+      this.fail(e.toString());
```
mprobst: return an empty annotation after this, so that code upstream always gets a non null value? I know this currently terminates execution, but it might not in the future (when we collect errors).

mprobst:
LGTM.

martine:
This failed due to Closure compile timing out after 20 seconds.  It takes 4.2s locally, so I think maybe it just got unlucky with a slow VM?  If will try landing this and if it still times out I will investigate either upping the time out again or figuring out some other approach for how we compile...

PR #85, opened 2016-03-07
change sickle API in preparation for externs

martine:
Change the API to potentially produce an externs file in addition
to the primary output.

Refactor the goldens management in the test to handle this new
output file type.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -6,6 +6,13 @@ export interface SickleOptions {
   untyped?: boolean;
 }
 
+export interface SickleOutput {
+  // The TypeScript source with Closure annotations inserted.
+  output: string;
+  // Generated externs declarations, if any.
+  externs: string;
+}
+
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
   return diags.map((d) => {
                 let res = ts.DiagnosticCategory[d.category];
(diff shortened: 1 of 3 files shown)


mprobst:
Random drive by comment: it's hard to have a rule optionally generate a file in Bazel if you cannot tell from the rule spec itself (but only from file contents). I think we'd have to unconditionally generate the externs, at least an empty file.

src/sickle.ts
```diff
@@ -36,12 +43,15 @@ class Annotator {
...
+  annotate(file: ts.SourceFile): SickleOutput {
     this.output = [];
     this.file = file;
     this.visit(file);
     this.assert(this.indent == 0, 'visit() failed to track nesting');
-    return this.output.join('');
+    return {
+      output: this.output.join(''),
+      externs: null,
+    };
```
rkirov: In general, do you prefer this pattern "interface + object literal", vs "class + new constructor" for simple objects where the differences don't matter? I don't have much of an opinion either way, but we should be consistent in the sickle codebase.

martine: I always prefer the former, but I admit I didn't think about it too hard.  This is mostly consistent with how TypeScript does it as well (e.g. when we're compiling a file via the TS API, it gives us back an object with a few fields like this).

martine:
@mprobst maybe I shouldn't have worried so much about optional files, but I figured it's nice for this change to not include another 20 empty .sickle_externs files since it wasn't too hard to implement.

src/sickle.ts
rkirov: sounds good, just making sure we are on the same page. We can learn more about the downsides of this pattern as the codebase grows.

rkirov:
LGTM

PR #86, opened 2016-03-08
don't reject @suppress {checkTypes} annotations

martine:
Rather than always disallowing {} in JSDoc, disallow it in the
specific cases we know that we already handle.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -68,7 +68,7 @@ export function getJSDocAnnotation(comment: string): JSDocComment {
     let match = line.match(/^@(\S+) *(.*)/);
     if (match) {
       let [_, tagName, text] = match;
-      if (text[0] == '{') {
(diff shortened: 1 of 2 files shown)


src/sickle.ts
```diff
@@ -68,7 +68,7 @@ export function getJSDocAnnotation(comment: string): JSDocComment {
     let match = line.match(/^@(\S+) *(.*)/);
     if (match) {
       let [_, tagName, text] = match;
-      if (text[0] == '{') {
+      if (['param', 'return', 'type'].indexOf(tagName) != -1 && text[0] == '{') {
```
rkirov: not sure if type without {} is any good?

martine: Yeah, I changed the error message in the follow up CL (I handle @type specially first).

martine: This CL is mostly useful for the additional test. :)

rkirov: I see

rkirov:
LGTM

PR #87, opened 2016-03-08
gather diagnostics rather than throwing exceptions

martine:
Produce an array of ts.Diagnostic, just like the compiler,
when encountering code we'd like the user to treat as invalid.
Keep this.fail() around for assertions.

Extend the test suite to dump the error messages in the golden
file so they can be reviewed.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -11,6 +11,8 @@ export interface SickleOutput {
   output: string;
   // Generated externs declarations, if any.
   externs: string;
+  // Error messages, if any.
+  diagnostics: ts.Diagnostic[];
 }
 
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
@@ -19,7 +21,7 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
                 if (d.file) res += ' at ' + d.file.fileName + ':';
                 if (d.start) {
                   let {line, character} = d.file.getLineAndCharacterOfPosition(d.start);
-                  res += line + ':' + character + ':';
+                  res += (line + 1) + ':' + (character + 1) + ':';
                 }
                 res += ' ' + d.messageText;
                 return res;
@@ -68,7 +70,10 @@ export function getJSDocAnnotation(comment: string): JSDocComment {
     let match = line.match(/^@(\S+) *(.*)/);
     if (match) {
       let [_, tagName, text] = match;
-      if (['param', 'return', 'type'].indexOf(tagName) != -1 && text[0] == '{') {
+      if (tagName == 'type') {
+        throw new Error('@type annotations are not allowed');
+      }
+      if ((tagName == 'param' || tagName == 'return') && text[0] == '{') {
         throw new Error('type annotations (using {...}) are not allowed');
       }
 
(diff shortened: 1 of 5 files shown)


martine:
Note: this is based on the other PR, so review that one first.

src/sickle.ts
```diff
@@ -68,7 +70,10 @@ export function getJSDocAnnotation(comment: string): JSDocComment {
     let match = line.match(/^@(\S+) *(.*)/);
     if (match) {
       let [_, tagName, text] = match;
-      if (text[0] == '{') {
+      if (tagName == 'type') {
+        throw new Error('@type annotations are not allowed');
```
rkirov: There are a few shortcuts for type annotations, for example @const {string}. Can handle them in a separate cl.

src/sickle.ts
```diff
@@ -120,6 +126,7 @@ class Annotator {
     return {
       output: this.output.join(''),
       externs: null,
+      diagnostics: this.diagnostics,
```
rkirov: I think you want to reset the diagnostics on each annotate call, similar to output. My mental model is that a single Annotator can be called multiple times.

src/sickle.ts
```diff
@@ -380,7 +384,14 @@ class Annotator {
...
       return getJSDocAnnotation(comment);
     } catch (e) {
-      this.fail(e.toString());
+      this.diagnostics.push({
+        file: this.file,
+        start: node.getFullStart() + pos,
+        length: end - pos,
+        messageText: e.message,
+        category: ts.DiagnosticCategory.Error,
+        code: 0,
```
rkirov: OOC, what is code: 0?

rkirov: How about having Sickle.Diagnostic interface instead of reusing ts.Diagnostic, with somewhat unclear codes?

rkirov:
Does that mean that we would still fail if there are any diagnostic errors, inside the internal tsc wrapper?

rkirov:
LGTM over all, only real concern is not resetting this.diagnostics on each annotate call.

src/sickle.ts
martine: code:0 is required by the type.

The reason to reuse ts.Diagnostic is that it lets the user of sickle merge sickle errors with TS errors and print them all with the same interface.

martine:
> Does that mean that we would still fail if there are any diagnostic errors, inside the internal tsc wrapper?

I didn't understand this comment.

rkirov:
From a user's perspective, if they write @type {string} prior to your change that is a compile time error, after your change it is no longer an error, but it is aggregated into a diagnostics array. So the tool that calls into sickle (tsc_wrapper in our case) can decide whether to error or just warn based on the diagnostics. I wanted to hear whats your plan on that.

martine:
Ah, I see.  My plan is for tsc_wrapped to treat sickle diagnostics just like it would TS diagnostics.  Currently any diagnostics are treated as compilation-failing errors in tsc_wrapped.

I have a matching change pending on the tsc_wrapped side that makes it obey the new diagnostics, which I will land at the same time as merging sickle.

rkirov:
Sounds good, thanks for the clarification.

PR #88, opened 2016-03-08
move property initialization to constructor

martine:
Move all the parameters to the constructor, so it's clear
that an Annotator can only be used once.  This also allows
moving a bunch of initialization-like code to happen at
construction time.

Also added some JSDoc while I was at it.


rkirov:
LGTM

PR #89, opened 2016-03-09
generate externs from 'declare' statements

martine:
When we encounter a node tagged with "ambient", do a separate
pass through the node's children to emit to an externs.js file.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -18,8 +18,8 @@ export interface SickleOutput {
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
   return diags.map((d) => {
                 let res = ts.DiagnosticCategory[d.category];
-                if (d.file) res += ' at ' + d.file.fileName + ':';
-                if (d.start) {
+                if (d.file) {
+                  res += ' at ' + d.file.fileName + ':';
                   let {line, character} = d.file.getLineAndCharacterOfPosition(d.start);
                   res += (line + 1) + ':' + (character + 1) + ':';
                 }
@@ -109,7 +109,11 @@ const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.Node
  */
 class Annotator {
   /** The primary annotated TypeScript output, as an array of strings. */
-  private output: string[] = [];
+  private tsOutput: string[] = [];
+  /** Output for the externs.js file. */
+  private externsOutput: string[] = [];
+  /** The current output, either tsOutput or externsOutput. */
+  private currentOutput: string[] = this.tsOutput;
   /** Warnings/errors found while examining the code. */
   private diagnostics: ts.Diagnostic[] = [];
 
@@ -127,13 +131,13 @@ class Annotator {
     this.visit(this.file);
     this.assert(this.indent == 0, 'visit() failed to track nesting');
     return {
-      output: this.output.join(''),
-      externs: null,
+      output: this.tsOutput.join(''),
+      externs: this.externsOutput.length > 0 ? this.externsOutput.join('') : null,
       diagnostics: this.diagnostics,
     };
   }
 
-  private emit(str: string) { this.output.push(str); }
+  private emit(str: string) { this.currentOutput.push(str); }
 
   private logWithIndent(message: string) {
     let prefix = new Array(this.indent + 1).join('| ');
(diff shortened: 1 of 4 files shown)


src/sickle.ts
```diff
@@ -18,8 +18,8 @@ export interface SickleOutput {
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
   return diags.map((d) => {
                 let res = ts.DiagnosticCategory[d.category];
-                if (d.file) res += ' at ' + d.file.fileName + ':';
-                if (d.start) {
+                if (d.file) {
+                  res += ' at ' + d.file.fileName + ':';
```
martine: This logic now matches the similar code inside of `tsc`.  It's valid to have a start point of 0.

src/sickle.ts
```diff
@@ -144,106 +148,109 @@ class Annotator {
...
-          // and to annotate methods.
-          this.writeTextBetween(classNode, classNode.members[0]);
-          for (let member of classNode.members) {
-            this.visit(member);
+          break;
+        case ts.SyntaxKind.ClassDeclaration:
+          let classNode = <ts.ClassDeclaration>node;
+          if (classNode.members.length > 0) {
+            // We must visit all members individually, to strip out any
+            // /** @export */ annotations that show up in the constructor
```
martine: All of this green code here is just indented one more level.  I considered refactoring to an additional function to prevent this indenting.  The only real change to this function is the above check for NodeFlags.Ambient, which shuttles this whole piece of the AST off to a helper function instead of doing this processing.

src/sickle.ts
```diff
@@ -399,6 +406,75 @@ class Annotator {
     }
   }
 
+  private visitExterns(node: ts.Node, namespace: string[] = []) {
```
martine: ...and right here is where the actual new code in this CL picks up again.

test_files/declare.sickle.ts
```diff
@@ -1,5 +1,21 @@
-declare module DeclareTest {
-  export class Foo {
-    constructor();
+Error at test_files/declare.in.ts:10:28: IndexSignature not implemented in externs for interface
+====
+declare module DeclareTestModule {
+  module inner {
```
rkirov: https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/Namespaces%20and%20Modules.md 

namespace inner is preferred

src/sickle.ts
```diff
@@ -399,6 +406,75 @@ class Annotator {
...
+    this.currentOutput = this.externsOutput;
+    switch (node.kind) {
+      case ts.SyntaxKind.ModuleDeclaration:
+        let decl = <ts.ModuleDeclaration>node;
+        namespace = namespace.concat([decl.name.text]);
```
rkirov: why not just `push(a)` instead of `concat([a])`

src/sickle.ts
```diff
@@ -399,6 +406,75 @@ class Annotator {
...
+        break;
+    }
+    this.currentOutput = this.tsOutput;
+  }
+
+  private writeExternsInterface(decl: ts.InterfaceDeclaration, namespace: string[]) {
+    this.emit('/** @record @struct */\n');
+    let className = namespace.concat([decl.name.getText()]).join('.');
+    if (namespace.length > 0) {
+      this.emit(`${className} = function() {};\n`);
```
rkirov: a helper emitLn(a) => emit(a + '\n') would help with readability.

src/sickle.ts
```diff
@@ -399,6 +406,75 @@ class Annotator {
...
+    for (let member of decl.members) {
+      switch (member.kind) {
+        case ts.SyntaxKind.PropertySignature:
+          let prop = <ts.PropertySignature>member;
+          this.maybeEmitJSDocType(prop.type, '@type');
+          this.emit(`\n${className}.prototype.${prop.name.getText()};\n`);
+          break;
+        default:
+          // Members can include things like index signatures, for e.g.
+          //   interface Foo { [key: string]: number; }
```
rkirov: not related to the code review, but guess is that index signatures can be expressed in closure by implementing the IObject<string, number> interface.

src/sickle.ts
```diff
@@ -399,6 +406,75 @@ class Annotator {
...
+        break;
+      default:
+        this.errorUnimplementedKind(node, 'externs generation');
+        break;
+    }
+    this.currentOutput = this.tsOutput;
+  }
+
+  private writeExternsInterface(decl: ts.InterfaceDeclaration, namespace: string[]) {
+    this.emit('/** @record @struct */\n');
```
rkirov: for my own understanding does @struct add anything on top of @record? I am used to seeing @struct on concrete classes to prevent adding fields post construction.

rkirov:
LGTM, just minor nitpicks.

src/sickle.ts
martine: I don't want to mutate the array, because I need the element to pop back off when I'm done (in the cases where I recurse multiple times).

src/sickle.ts
martine: Will do.

src/sickle.ts
martine: Yes, I figured I'd wait until I encountered one before worrying about it.

test_files/declare.sickle.ts
martine: Done.

src/sickle.ts
martine: I am not sure, to be honest.  Do you think I should just use @record?  I don't really understand what any of these mean...

rkirov: They are too ill-defined to truly "mean" much :) The closest thing I have to a specification are the examples here (which don't specify behavior of @struct on interfaces) :/
https://github.com/google/closure-compiler/wiki/@struct-and-@dict-Annotations
https://github.com/google/closure-compiler/wiki/Structural-Interfaces-in-Closure-Compiler

So the only thing left is poking at the implementation:
https://closure-compiler-debugger.appspot.com/home#input0%3D%252F**%2520%2540record%2520%2540struct%2520*%252F%250Afunction%2520Recursive()%257B%257D%253B%250A%252F**%2520%2540type%2520%257B%253FRecursive%257D%2520*%252F%250ARecursive.prototype.next%253B%250A%250Avar%2520x%2520%253D%2520%257B%257D%250Ax.next%2520%253D%2520x%253B%250A%252F**%2520%2540type%2520%257BRecursive%257D%2520*%252F%250Avar%2520a%2520%253D%2520x%253B%250A%250Aa%255B'foo'%255D%2520%253D%25200%253B%26input1%26conformanceConfig%26externs%26refasterjs-template%26includeDefaultExterns%3D1%26CHECK_SYMBOLS%3D1%26CHECK_TYPES%3D1%26LANG_IN_IS_ES6%3D1%26MISSING_PROPERTIES%3D1%26PRESERVE_TYPE_ANNOTATIONS%3D1%26PRETTY_PRINT%3D1%26TRANSPILE%3D1

Which indeed shows that @struct makes a difference for allowing/disallowing index setters. So good call to keep both.

PR #90, opened 2016-03-10
re-export each name from "export * from ..."

### LINKED ISSUE #47: De-reference all properties in `export * from ...`
When compiling `export * from 'foo';` to CommonJS module format, we end up with code like this:

```js
function __export(m) { for (var k in m) exports[k] = m; }
__export(require('somemodule'));
```

Now surprisingly this works in Closure Compiler when replacing `require` with `goog.require`. It simply patches the `exports` to be the module name (which is a global symbol) and then the right properties end up on the right objects (yay!).

However when importing a property that was re-exported like this, Closure Compiler complains about a missing property. It's possible to shut that off (`--jscomp_off=missingProperties`), but that seems overall dangerous - for example, dead code elimination might kill that property.

If sickle was to "de-reference" and re-export the particular symbols, this would avoid the problem for the time being, and allow us to continue using the hacky CommonJS --> `goog.require` transformation without having to rely on ES6 modules.

An alternative, possibly cleaner, solution would be to convert all module imports to `goog.module` and `goog.require` within sickle (including the `*` dereferencing). This would break the second compilation pass - TypeScript would no longer understand all those imports. However we have already type checked, so at that point we could just ignore all errors and emit.

WDYT? @martine @rkirov

martine:
If we leave "export *" alone, TypeScript emits a for loop over
the exports of the re-exported module, but for Closure purposes
we want to instead emit the more verbose "exports.foo = module.foo"
for each exported symbol.

Fixes #47.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -15,6 +15,16 @@ export interface SickleOutput {
   diagnostics: ts.Diagnostic[];
 }
 
+/** The compiler options used by the test suite and sickle command line. */
+export const compilerOptions: ts.CompilerOptions = {
+  target: ts.ScriptTarget.ES6,
+  noImplicitAny: true,
+  skipDefaultLibCheck: true,
+  noEmitOnError: true,
+  experimentalDecorators: true,
+  emitDecoratorMetadata: true,
+};
+
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {
   return diags.map((d) => {
                 let res = ts.DiagnosticCategory[d.category];
@@ -125,7 +135,8 @@ class Annotator {
   /** The node currently being visited by visit(). This is only used in error messages. */
   private currentNode: ts.Node;
 
-  constructor(private options: SickleOptions, private file: ts.SourceFile) {}
+  constructor(
+      private program: ts.Program, private file: ts.SourceFile, private options: SickleOptions) {}
 
   annotate(): SickleOutput {
     this.visit(this.file);
@@ -156,6 +167,22 @@ class Annotator {
       this.writeRange(node.getFullStart(), node.getEnd());
     } else {
       switch (node.kind) {
+        case ts.SyntaxKind.ExportDeclaration:
+          let exportDecl = <ts.ExportDeclaration>node;
+          if (!exportDecl.exportClause && exportDecl.moduleSpecifier) {
+            // It's an "export * from ..." statement.
+            // Rewrite it to re-export each exported symbol directly.
+            this.emit('export {');
+            let typeChecker = this.program.getTypeChecker();
+            let exports = typeChecker.getExportsOfModule(
+                typeChecker.getSymbolAtLocation(exportDecl.moduleSpecifier));
+            this.emit(exports.map(e => e.name).join(','));
+            this.emit('} from');
+            this.writeRange(exportDecl.moduleSpecifier.getFullStart(), node.getEnd());
+          } else {
+            this.writeNode(node);
+          }
+          break;
         case ts.SyntaxKind.InterfaceDeclaration:
           let decl = <ts.InterfaceDeclaration>node;
           this.writeRange(node.getFullStart(), node.getEnd());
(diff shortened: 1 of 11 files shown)


martine:
The hardest part of this was figuring out the tests.  I need this imported module `extern_helper` to simultaneously:
1. be importable from TypeScript
2. be importable from JavaScript
3. show up in the command line for the final closure test

I had to hack it somewhere, hope the place I hacked it is adequate.

test_files/export.sickle.ts
```diff
@@ -0,0 +1,2 @@
+export {foo,bar} from './export_helper';
```
martine: Here you can see that we found "foo" and "bar" in the helper and rewrote the export accordingly.

src/sickle.ts
```diff
@@ -15,6 +15,16 @@ export interface SickleOutput {
   diagnostics: ts.Diagnostic[];
 }
 
+/** The compiler options used by the test suite and sickle command line. */
+export const compilerOptions: ts.CompilerOptions = {
+  target: ts.ScriptTarget.ES6,
+  noImplicitAny: true,
+  skipDefaultLibCheck: true,
```
rkirov: OOC, what was the issue with noResolve?

test/test_support.ts
```diff
@@ -103,4 +95,17 @@ export function goldenTests(): GoldenFileTest[] {
       es6Path: testPath.replace(/\.in\.ts$/, '.tr.js'),
     };
   });
+
+  // export_helper.ts is special, because it is imported by another test.
+  // It it must not have a ".in" or ".tr" in its filenames.
```
rkirov: .ts

test_files/export.in.ts
```diff
@@ -0,0 +1,2 @@
+export * from './export_helper';
```
rkirov: It might be too much to ask, but from experience I find that another level of reexport might have issues where a single level doesn't. That is adding `export * from './export_helper_2';` in `export_helper`.

src/sickle.ts
```diff
@@ -627,16 +654,21 @@ function last<T>(elems: T[]): T {
...
-    console.log(path + ':');
-    console.log(annotate(sf));
+  let paths = process.argv.splice(2);
+  let program = ts.createProgram(paths, compilerOptions);
+  let diags = ts.getPreEmitDiagnostics(program);
+  if (diags.length > 0) {
+    console.log(formatDiagnostics(diags));
+    process.exit(1);
+  }
+  for (let fileName of paths) {
```
rkirov: let filePath of paths? In my mind fileName is purely a name, no path.

src/sickle.ts
```diff
@@ -156,6 +167,22 @@ class Annotator {
       this.writeRange(node.getFullStart(), node.getEnd());
     } else {
       switch (node.kind) {
+        case ts.SyntaxKind.ExportDeclaration:
+          let exportDecl = <ts.ExportDeclaration>node;
+          if (!exportDecl.exportClause && exportDecl.moduleSpecifier) {
+            // It's an "export * from ..." statement.
```
rkirov: Can't help but wonder what happens when you have circular 'export *',

a.ts
export * from 'b'

b.ts
export * from 'a'
export var b;

No action just wondering if you tested that.

rkirov:
LGTM over all, I would love to see an extra level of nesting if the test infrastructure can take it :)

alexeagle:
found another issue in google3 (maybe):

core.ts:
```
export * from './src/core/metadata';
export * from './src/core/util';
export * from './src/core/prod_mode';
export * from './src/core/di';
export * from './src/facade/facade';
export {enableProdMode} from 'angular2/src/facade/lang';
```

result from blaze
```
third_party/javascript/angular2/github/modules/angular2/core.ts:1:829: Duplicate identifier 'enableProdMode'.
third_party/javascript/angular2/github/modules/angular2/core.ts:2:9: Duplicate identifier 'enableProdMode'.
```

martine:
Dang, that is exactly the thing I was talking about in
https://github.com/angular/sickle/issues/47#issuecomment-194610941
!

martine:
https://github.com/angular/sickle/issues/47#issuecomment-195124529

src/sickle.ts
martine: From the docs: "Do not add triple-slash references or module import targets to the list of compiled files."

This took me like an hour to figure out why I couldn't get the damn test to load the extra module... :~(

test/test_support.ts
martine: It is correct but confusing.  I'll reword it.

test_files/export.in.ts
martine: You're killing me. :)

OK, done.

src/sickle.ts

src/sickle.ts
martine: I did not.  I expect it to break.  :)

martine:
Because this works for the basic case, I'm gonna merge this and think more about Alex's case on the bug.

PR #91, opened 2016-03-11
when converting export *, avoid local symbol names

martine:
We convert "export *" into "export {foo,bar,...}".
But if there's a locally exported symbol of the same name,
we shouldn't include it in the export list.

More work on #47.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -172,10 +172,19 @@ class Annotator {
           if (!exportDecl.exportClause && exportDecl.moduleSpecifier) {
             // It's an "export * from ..." statement.
             // Rewrite it to re-export each exported symbol directly.
-            this.emit('export {');
             let typeChecker = this.program.getTypeChecker();
             let exports = typeChecker.getExportsOfModule(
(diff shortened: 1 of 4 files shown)


rkirov:
This was educational :) LGTM.

src/sickle.ts
```diff
@@ -172,10 +172,19 @@ class Annotator {
...
             let exports = typeChecker.getExportsOfModule(
                 typeChecker.getSymbolAtLocation(exportDecl.moduleSpecifier));
+
+            // Avoid reexporting any names that are already locally exported.
+            let locals = typeChecker.getSymbolsInScope(this.file, ts.SymbolFlags.Export);
+            let localSet: {[name: string]: boolean} = {};
+            for (let local of locals) {
+              localSet[local.name] = true;
+            }
+            exports = exports.filter(e => !localSet[e.name]);
```
mprobst: Should be `hasOwnProperty` here - this way you cannot ever export `toString` or `prototype`, etc.

PR #92, opened 2016-03-11
when re-exporting *, avoid other named exports

martine:
When exporting each var by name from an "export *",
we already avoid locals like "export var foo".
But we were missing re-exports like "export {foo} from ..."

More fixing on #47.


test_files/export_helper_2.js
```diff
@@ -1,3 +1,5 @@
 // This file isn't itself a test case, but it is imported by the
 // export.in.ts test case.
-export let baz = 3;
+export let export2 = 3;
```
rkirov: export2 is already exported by export_helper.ts,

rkirov:
LGTM

src/sickle.ts
```diff
@@ -290,6 +280,40 @@ class Annotator {
...
+    let typeChecker = this.program.getTypeChecker();
+
+    // Gather the names of local exports, to avoid reexporting any
+    // names that are already locally exported.
+    let locals =
+        typeChecker.getSymbolsInScope(this.file, ts.SymbolFlags.Export | ts.SymbolFlags.Alias);
+    let localSet: {[name: string]: boolean} = {};
+    for (let local of locals) {
+      localSet[local.name] = true;
+    }
```
rkirov: This guy doesn't depend on exportDecl, I wonder if there is a better way to put it, instead of rebuilding it per decl.

martine: Hmm, yeah.  But it's only a performance thing, not a correctness thing right?

src/sickle.ts
```diff
@@ -290,6 +280,40 @@ class Annotator {
...
+
+    // Expand the export list, then filter it to the symbols we want
+    // to reexport
+    let exports =
+        typeChecker.getExportsOfModule(typeChecker.getSymbolAtLocation(exportDecl.moduleSpecifier));
+    let reexports: {[name: string]: boolean} = {};
+    for (let sym of exports) {
+      let name = sym.name;
+      if (localSet.hasOwnProperty(name)) {
+        // Already exported via "export var ..." or "export {foo} from ...".
```
rkirov: 'already exported' might be confused that the value is the same too, can we add the word "shadowing" to the comment.

rkirov:
LGTM

src/sickle.ts
martine: Good idea, I rewrote the comment.

src/sickle.ts
rkirov: yes, just performance

PR #94, opened 2016-03-12
when re-exporting *, don't avoid imports

martine:
The TypeScript API that let us retrieve aliased exports also
ended up retrieving aliased imports.  Avoid this by checking
the symbols we got.

More work on #47.


martine:
I won't submit this until it works in google3, but you can take a look now.

rkirov:
LGTM

PR #98, opened 2016-03-14
fix(enums): Increment enum initializer from previous constant

### LINKED ISSUE #97: Enum written with incorrect reverse map
from angular2/github/modules/angular2/src/compiler/url_resolver.ts
```
enum _ComponentIndex {
  Scheme = 1,
  UserInfo,
  Domain,
  Port,
  Path,
  QueryData,
  Fragment
}
```

TS/sickle produces this ES6 code:
```
var _ComponentIndex;
(function (_ComponentIndex) {
    _ComponentIndex[_ComponentIndex["Scheme"] = 1] = "Scheme";
    _ComponentIndex[_ComponentIndex["UserInfo"] = 2] = "UserInfo";
    _ComponentIndex[_ComponentIndex["Domain"] = 3] = "Domain";
    _ComponentIndex[_ComponentIndex["Port"] = 4] = "Port";
    _ComponentIndex[_ComponentIndex["Path"] = 5] = "Path";
    _ComponentIndex[_ComponentIndex["QueryData"] = 6] = "QueryData";
    _ComponentIndex[_ComponentIndex["Fragment"] = 7] = "Fragment";
})(_ComponentIndex || (_ComponentIndex = {}));
_ComponentIndex.Scheme = 1;
_ComponentIndex.UserInfo = 0;
_ComponentIndex.Domain = 1;
_ComponentIndex.Port = 2;
_ComponentIndex.Path = 3;
_ComponentIndex.QueryData = 4;
_ComponentIndex.Fragment = 5;
```
Causes a runtime failure, we have 
`["components/youtube_app.html", undefined, undefined, undefined, undefined, "components/youtube_app.html", undefined, undefined]`
and then `parts[_ComponentIndex.Path][0] == '/'`
fails because `_ComponentIndex.Path` is 3 rather than 5.

alexeagle:
fixes #97

### DIFF
--- a/test_files/enum.sickle.ts
+++ b/test_files/enum.sickle.ts
@@ -13,4 +13,19 @@ export enum EnumTest2 {XYZ, PI = 3.14159}
 (<any>EnumTest2).XYZ = 0;
 /** @type {EnumTest2} */
 (<any>EnumTest2).PI =  3.14159;
+/** @typedef {number} */
+
+
+// Repro for #97
+enum ComponentIndex {
+  Scheme = 1,
+  UserInfo,
+  Domain
+}
+/** @type {ComponentIndex} */
+(<any>ComponentIndex).Scheme =  1;
+/** @type {ComponentIndex} */
+(<any>ComponentIndex).UserInfo = 2;
+/** @type {ComponentIndex} */
+(<any>ComponentIndex).Domain = 3;
 
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -592,6 +592,10 @@ class Annotator {
       this.emit(`(<any>${node.name.getText()}).${member.name.getText()} = `);
       if (member.initializer) {
         this.visit(member.initializer);
+        let enumConstValue: number;
+        if (enumConstValue = this.program.getTypeChecker().getConstantValue(member)) {
+          i = enumConstValue + 1;
+        }
       } else {
         this.emit(String(i));
         i++;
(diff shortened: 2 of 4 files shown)


martine:
What did you think of my suggestion to just drop the definitions of these values?

alexeagle:
I was suspicious. We do need the enums to be backward-mapped, don't we?
I'll try it and see if anythnig breaks.

mprobst:
We need the non-quoted property names to allow for code renaming, i.e. Closure wants to rename `FooEnum.Bar` to `FooEnum.aahsdj`. TypeScript only generates quoted assignments, which do not get renamed, so we have to insert a code snippet with a non-quoted symbol that gets the value assigned.

martine:
Ah, right.  This is correct then.

test_files/enum.sickle.ts
```diff
@@ -13,4 +13,19 @@ export enum EnumTest2 {XYZ, PI = 3.14159}
 (<any>EnumTest2).XYZ = 0;
 /** @type {EnumTest2} */
 (<any>EnumTest2).PI =  3.14159;
+/** @typedef {number} */
```
martine: This comment is a docstring on the below enum.  I bet it's not applying right.

To fix this, you need to emit the enum with no preceding comments.

martine: Change the above call from `this.writeNode(node);` to `this.writeTextBetween(node.getStart(), node.getEnd())` to try working around this.

alexeagle: Note, this issue was already present at line 7. I'll try to fix in both places.

martine: Oh in that case, leave it alone and we can do another CL if it's an issue.

martine:
LGTM++, sorry for all the false directions.

PR #101, opened 2016-03-14
support arrays in type declarations

### LINKED ISSUE #99: Support array destructuring
eg.
```
function f([x,y]: number[]) {}
```
See #96 where I started to do this, but needs more work because `number[]` must be represented as `Array<number>` in the closure typedef.

martine:
Expand a TypeScript array into Array<...>.

Fixes #99.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -271,6 +271,12 @@ class Annotator {
           this.writeNode(node);
           this.emit(')');
           break;
+        case ts.SyntaxKind.ArrayType:
+          let arrayType = <ts.ArrayTypeNode>node;
+          this.emit('Array<');
+          this.visit(arrayType.elementType);
+          this.emit('>');
+          break;
         default:
           this.writeNode(node);
           break;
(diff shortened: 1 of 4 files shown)


martine:
Jeepers!

```
  test_files/functions.tr.js:27: ERROR - actual parameter 1 of $jscomp.makeIterator does not match formal parameter
  found   : (Array<number>|null)
  required: (Array<number>|Iterable<number>|Iterator<number>|string)
  function Destructuring2([a, b], [[c]]) { }
                          ^
  
  test_files/functions.tr.js:27: ERROR - actual parameter 1 of $jscomp.makeIterator does not match formal parameter
  found   : (Array<(Array<string>|null)>|null)
  required: (Array<(Array<string>|null)>|Iterable<(Array<string>|null)>|Iterator<(Array<string>|null)>|string)
  function Destructuring2([a, b], [[c]]) { }
                                  ^
```

alexeagle:
Maybe that is the IArrayLike work that John Lenz has been doing?

alexeagle:
LGTM, barring maybe a different type than "Array" makes the error go away

martine:
To get this right, we have to mark all the arrays as non-nullable.  In the case of the double-nested destructed array, that means we have to recursively mark values as non-nullable!

PR #102, opened 2016-03-15
blacklist some variable names for closure externs

martine:
Don't create externs for variables that exist in Closure externs
already, such as "module" and "global".


rkirov:
reviewed on http://reviews.angular.io/D4

PR #103, opened 2016-03-16
fix(angular2): blacklist two more declare'd vars from angular2

alexeagle:
I know, Phabricator.

I'm burning down a big list of hacks I had to add for angular2 to run, and would rather not take a detour into php-cli right now.

### DIFF
--- a/src/sickle.ts
+++ b/src/sickle.ts
@@ -33,6 +33,8 @@ export let closureExternsBlacklist: string[] = [
   'exports',
   'global',
   'module',
+  'WorkerGlobalScope',
+  'Symbol',
 ];
 
 export function formatDiagnostics(diags: ts.Diagnostic[]): string {


src/sickle.ts
```diff
@@ -33,6 +33,8 @@ export let closureExternsBlacklist: string[] = [
   'exports',
   'global',
   'module',
+  'WorkerGlobalScope',
```
rkirov: This list needs to be more structured (we can keep hacking on it for now). Here is what I did in clutz:

- fix which files are considered "platform apis" on both sides:
  - lib.d.ts for TS (TSsymbols)
  - closure/externs:COMMON (ClosureSymbols)

- modify lib.d.ts to be a strict subset of closure externs
- enumerate symbols in missingPlatformSymbols = ClosureSymbols \ TSSymbols ( \ is set difference)

clutz now emits everything not in closure externs + missingPlatformSymbols.

Beyond "platform externs", I don't think the tool should be doing any white/black listing, as we can't have any visibility into what those symbols are. It won't scale to keep adding them here.

alexeagle: Not clear to me - are you suggesting I do that reorganization in this PR,
or is this a followup issue?

rkirov: follow up, just kicking off the proper design for this.

alexeagle: cool, filed https://github.com/angular/sickle/issues/106 for followup

alexeagle:
merged as 9ec7ac8

PR #122, opened 2016-04-08
Update Closure Compiler to v20160315

Dominator008:


martine:
Thanks for looking!  The test failure is due to needing a `s/checks-only/checks_only/` in the compiler driver, I think, but upon fixing that I see some deeper problem.  Still looking...

martine:
The other error is due to us using a type like
```js
/**
 * @param {!Array<Array<number>}} p
 */
function x([[foo]]) {
}
```

For now I think I'll just drop that code.

martine:
FYI, the diff to drop the bad code: https://reviews.angular.io/D51

martine:
Landed in ad0b43f290f2d36c2a7310e4d11c0cffbd2d45b5

PR #127, opened 2016-04-11
chore: disable the no-string-literal lint check

### LINKED ISSUE #126: tslint no-string-literal check is wrong
```
[16:19:04] [gulp-tslint] error (no-string-literal) main.ts[43, 18]: object access via string literals is disallowed
```

However, the object in question is from minimist and defined as

```js
		export interface ParsedArgs {
			[arg: string]: any;
			_: string[];
		}
```

so I should be able to index into it with a string.

What's the fix -- fix tslint or fix our settings?

mprobst:
It turns out it only syntactically checks object accesses, and is overly aggressive in doing so. Fixes #126.

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -18,7 +18,6 @@
     "no-inferrable-types": true,
     "no-internal-module": true,
     "no-shadowed-variable": true,
-    "no-string-literal": true,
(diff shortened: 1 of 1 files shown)


martine:
LGTM

PR #134, opened 2016-04-28
allow import {convertDecorators} from 'tsickle';

alexeagle:
note, I haven't had time to set up php-cli for Arcanist, and we agreed simple PRs would be okay on github.

I plan to use this because otherwise there is no `import {convertDecorators} from '??'` that works at type-check and runtime. (`build/definitions` vs `build/src/`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,8 +11,10 @@
   "dependencies": {
     "minimist": "^1.2.0",
     "source-map": "^0.4.2",
-    "source-map-support": "^0.3.1",
-    "typescript": "^1.8.0"
+    "source-map-support": "^0.3.1"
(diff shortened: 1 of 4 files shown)


alexeagle:
Fails on travis - circular reference problem? I think we need a bit of reorganizing, probably the Rewriter class moves to its own file?

alexeagle:
note, the peerDependencies change is needed too, otherwise sickle will `require` a 1.8 typescript and return objects which don't typecheck against the app (angular uses 1.9)

martine:
LGTM

alexeagle:
thanks, merged.

PR #135, opened 2016-04-28
Allow abstract class in ctorParameters

alexeagle:


martine:
LGTM

alexeagle:
thanks, merged as 33fc32f

PR #140, opened 2016-05-02
fix(ctorParams): allow undefined ctorParam type

### LINKED ISSUE #139: Emit `ctorParameters` also if there is not type but decorators
E.g.
```
class ViewUtils {
  constructor(@Inject(APP_ID) private _appId: string) {}
}
```

Right now, this emits `ViewUtils.ctorParameters = [null];`, but it should emit
```
    ViewUtils.ctorParameters = [
        { type: undefined, decorators: [{ type: decorators_1.Inject, args: [application_tokens_1.APP_ID,] },] },
    ];
```

alexeagle:
Fixes #139

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -49,6 +49,7 @@ class ClassRewriter extends Rewriter {
       let decorators: ts.Decorator[];
       if (param.decorators) {
         decorators = param.decorators.slice();
+        hasDecoratedParam = true;
       }
       if (param.type) {
         switch (param.type.kind) {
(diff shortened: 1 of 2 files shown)


src/decorator-annotator.ts
```diff
@@ -65,16 +66,17 @@ class ClassRewriter extends Rewriter {
...
         ctorParameters.push([paramCtor, decorators]);
       } else {
         ctorParameters.push(null);
       }
     }
 
+
     // Use the ctor parameter metadata only if the class or the ctor was decorated.
     if (this.decorators || hasDecoratedParam) {
+      console.log(ctorParameters);
```
tbosch: Remove this

src/decorator-annotator.ts
```diff
@@ -143,6 +145,7 @@ class ClassRewriter extends Rewriter {
           `static ctorParameters: {type: Function, decorators?: DecoratorInvocation[]}[] = [\n`);
       for (let param of this.ctorParameters) {
         if (!param) {
+          console.log("found a null elemen t!");
```
tbosch: Remove this

PR #142, opened 2016-05-02
fix(props): emit decorators for props and set/getters

alexeagle:


src/decorator-annotator.ts
```diff
@@ -120,6 +120,11 @@ class ClassRewriter extends Rewriter {
       case ts.SyntaxKind.Decorator:
         // Skip emit of all decorators, as they are specially handled.
         return true;
+      case ts.SyntaxKind.PropertyDeclaration:
```
tbosch: Move up the the `ts.SyntaxKind.MethodDeclaration` so there is only one `case` block for all of them.

src/decorator-annotator.ts
```diff
@@ -82,7 +82,7 @@ class ClassRewriter extends Rewriter {
   /**
    * gatherMethod grabs the decorators off a class method and emits nothing.
    */
-  private gatherMethod(method: ts.MethodDeclaration) {
+  private gatherMethod(method: ts.Declaration) {
```
tbosch: Rename to `gatherMethodOrProperty`

alexeagle:
merged

PR #147, opened 2016-05-31
DO NOT MERGE: Failing test for re-exporting variable with __ prefix

vikerman:


rkirov:
fix at https://reviews.angular.io/D130

PR #152, opened 2016-06-08
Correctly compare paths to files in the current directory.

jjudd:
If the tsickle main function is run against files in the same directory,
then closure type comments are not added to the outputted JavaScript. For
example, if you run tsickle against './hello.ts' then the outputted
hello.js will not have type comments.

This occurs because the sourceReplacingCompilerHost compares fileNames to
determine whether we ran tsickle on the source file. 'hello.ts' is compared
to './hello.ts' and we incorrectly skip the file.

This is fixed by comparing the path to the files using the TypeScript
sys utilities.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -143,11 +143,12 @@ function createSourceReplacingCompilerHost(
       fileName: string, languageVersion: ts.ScriptTarget,
       onError?: (message: string) => void): ts.SourceFile {
     let sourceText: string;
-    if (substituteSource.hasOwnProperty(fileName)) {
-      sourceText = substituteSource[fileName];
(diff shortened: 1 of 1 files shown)


jjudd:
I signed the CLA. Also, I hope this is a decent way to fix this. I downloaded tsickle and was considering using it for a large closure codebase. When I couldn't get it to work on a small example program, I was a bit concerned :)

martine:
Thanks for looking!  We are successfully using tsickle within Google, and this main.ts is an attempt at open-sourcing the general gist of how it works.  (Our internal code is full of Google-specific hacks for our weird source code layout.)  So as you have observed, it doesn't quite work yet, but I am pretty confident that it can be made to work because the bulk of the tsickle library is in successful use.

PR #154, opened 2016-06-09
Replace all slashes with periods in the module name instead of just the first

jjudd:


martine:
Can you ensure a test covers this?  I think perhaps the best option might be to move the `pathToModuleName` found in `tsickle_test.ts` out into a place where `main.ts` could also call it, so that those tests would cover it.

(Another way of saying this is that all the code in `main.ts` is currently not covered by tests, so moving code out from there and into other files will improve test coverage.)

jjudd:
Can do.

jjudd:
I have found a couple different versions of this function. Curious which one would be preferred? Or should they be combined?

```typescript
function pathToModuleName(context: string, fileName: string): string {
  if (fileName[0] === '.') {
  }
  return fileName.replace(/\.js$/, '').replace(/\//g, '.');
}


function pathToModuleName(context: string, fileName: string) {
  if (fileName[0] === '.') {
  }
  return fileName.replace(/\//g, '$').replace(/_/g, '__');
}````

martine:
It doesn't matter too much.  In a pure-TS app, the module names disappear after the closure compile.  It could matter if you import a TS module into a Closure JS module (you need to know what the name is).

The former one seems better to me.

jjudd:
I went with the first function and used it in both the CLI and the tests. I'm not sure what the naming convention is and naming things is just hard in general. If we want to rename things, I'm totally fine with that.

martine:
Thanks for the patch!

Issue #153, opened 2016-06-08
tsickle only has a devDependency on google-closure-compiler, but the binary depends on it


jjudd:
As the title mentions, `google-closure-compiler` is in the `devDependencies` section of `package.json`. However, the binary supplied by `main.ts` requires `google-closure-compiler`. As a result, unless you have `google-closure-compiler` installed, tsickle fails to run. 

For example:
```
Error: Cannot find module 'google-closure-compiler'
```

To resolve this, `google-closure-compiler` should probably move to `peerDependencies` or `dependencies`. Alternatively, the cli could be split into its own module, as suggested in #137.

Thoughts?

martine:
(Are you working on an Angular2 app, by chance?  As mentioned in #137, for Angular2 it's likely they'll have their own tsickle wrapper etc., which is why we haven't put much work into the tsickle command line.)

For now we can fix this by moving the dependency into a non devdependency.  @alexeagle will this break your ngc stuff?

Otherwise, I am kinda new to nodejs packaging stuff -- if you have suggestions on how to best set up a separate CLI package (ideally it'd still be in this repository) I'd welcome it.

alexeagle:
If we make it a `dependency` then all angular 2 users will download closure compiler, which we don't want.
I think you need the `@angular/tsc-wrapped` package to support closure-annotating the output. I did an experiment with it https://github.com/alexeagle/angular/commit/2d1c21f3705078eb3332f73fdd52bd698508b8e7 which I think we need to get into the main repo.

jjudd:
Hi @martine. I'm working on Lucidchart, which has Angular2 parts and plain on JavaScript parts. All of it is run through the closure compiler. I'm interested in getting TypeScript to work with the entire codebase, so I'm trying out tsickle and clutz. 

I was planning on the workflow being .ts -> (tsickled .js) + (regular closure .js) -> closure compiler. Our build system should be able to handle everything once I run it through tsickle. The command line tool seemed like a great way to make that happen. Just run it with `saveTemporaries` and then send those with the rest of the js into the closure compiler.

With what we plan on doing, it is more useful for to have the CLI not do closure compilation. Instead it would just annotate the js and output the files. That would also solve the dependency problem on the closure compiler, but I'm not sure if it is in line with the goal of the CLI tool.

martine:
[...]

PR #155, opened 2016-06-10
Remove closure compilation from the CLI

jjudd:
This removes the need for google-closure-compiler to be a dependency. Now
it can be a dev dependency and the CLI should just work when downloaded
from npm.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,5 @@
 #!/usr/bin/env node
 
-import * as closure from 'google-closure-compiler';
 import * as fs from 'fs';
 import * as minimist from 'minimist';
 import * as path from 'path';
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -76,8 +67,8 @@ function loadSettingsFromArgs(args: string[]): {settings: Settings, tscArgs: str
         process.exit(1);
     }
   }
-  if (!settings.outputPath && !settings.externsPath) {
-    console.error('must specify --output or --externs path');
+  if (!settings.externsPath) {
```
jjudd: Because this is required, it might be worth making this an argument instead of an option. Alternatively, do not make it required.

jjudd:
This would resolve #153, it does not have the rest of the enhancements discussed there, but this gets the  npm package working without needing the google-closure-compiler installed.

src/main.ts
```diff
@@ -266,23 +217,13 @@ function main(args: string[]) {
     process.exit(1);
   }
 
-  if (settings.saveTemporaries) {
-    for (let fileName of Object.keys(jsFiles)) {
-      fs.writeFileSync(fileName, jsFiles[fileName]);
-    }
+  for (let fileName of Object.keys(jsFiles)) {
+    fs.writeFileSync(fileName, jsFiles[fileName]);
```
martine: Does this always write the files next to the .ts files?  Or does it obey outDir?  Just curious.

jjudd: As far as I can tell, it obeys all the output rules of the TypeScript compiler. We're using outDir without any problems.

PR #156, opened 2016-06-13
Add declareLegacyNamespace option for codebases migrating to goog.module

jjudd:
tsickle uses goog.module. Not all closure codebases have fully migrated to
goog.module. In an attempt to support codebases which still use
goog.provide and goog.require, enable tsickle to output
goog.module.declareLegacyNamespace().

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -19,6 +19,9 @@ const internalExternsFileName = 'tsickle_externs.js';
 interface Settings {
   /** If provided, path to save externs to. */
   externsPath?: string;
+
+  /** If provided, declareLegacyNamespace when replacing modules with goog.module */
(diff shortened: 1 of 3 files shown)


mprobst:
Just FYI, I think internally the recommended way to migrate to `goog.module()` is using `goog.require('my.module');` and then `goog.module.get('my.module');` at the place where you use it. Using that has some advantages, e.g. you no longer have code accidentally accessing global properties that should be in modules, and should require explicit dependencies.

jjudd:
Nice, that's great to know. Make sense. We've been using only goog.provide/require up until this point, so I'm not very familiar with modules.

If that's the case, then it is probably best to just close this PR.

Issue #146, opened 2016-05-19
Can't emit interfaces into same namespace as values


martine:
```js
interface Foo { x: number; }
var Foo;
```
is legal TS.  But if you translate the `interface` into Closure directly it goes into the single (value) namespace and the two definitions conflict.

It's easy enough to just emit interfaces under some mangled name in Closure, e.g.

```js
interface Foo { x: number }
/** @record */
function tsickle_Foo() {}
/** @type {number}
tsickle_Foo.prototype.x;
```

but then you need to fix every *reference* to that interface in any Closure type, while at the same time leaving other references (e.g. `Array`) alone.  And you need to figure out how to import it across files (if a file does `import {Foo} from bar`, you'd need to change it to `import {Foo, tsickle_Foo} from bar`).

martine:
Rado suggested we might be able to do some trickery like
```js
interface ts_Foo { ... }
type Foo = ts_Foo;
```
and then when `Foo` is used, note the alias and use that to adjust the emitted type.

But a local experiment found that the place where we emit types gets the type after aliasing has been resolved.

martine:
Idea: *always* emit TS interfaces into a mangled Closure name, like `tsickle_$name`.  Then there's no collisions possible, and it's always easy to guess the mangled name -- whenever you see an interface reference, emit it with the mangled name.

Problem 1.  This breaks for builtin types like "Array" and "Object".  However, emitting those interfaces verbatim as we do right now is saying that they are equivalent between TS and Closure's definitions of those terms.  I think this is fine for Array and Object, but it might not be fine for more complex types.  I think a whitelist (perhaps derived from lib.d.ts) of types that are considered compatible might get us pretty far.

Problem 2.  Mangling names breaks interop with Closure code.  E.g. a function returns a `FooBar`, and we see that and emit it as a `tsickle_FooBar` in our code.

I think collisions are unlikely enough that this approach is not gonna work.  I will leave it here as a rejected proposal.

martine:
Idea 2: locally detect when there is a naming conflict and just dodge it there.

It might be possible to look up, given a symbol, whether that symbol represents both a type and a name in a TS file.  When that scenario occurs, rewrite all the instances of the type to something else when emitting.

The hard part then is making the dependent code recognize it.  Within a compilation unit perhaps we can just construct a map by doing one pass over all the files first before emitting any.  Across compilation units, we can stash the mangled name into a comment and extract it from the docstring in the d.ts file(!).

[...]

Issue #112, opened 2016-03-25
Translate namespaced type references in jsdoc


martine:
Input:
```js
import {SomeClass} from './foo';

function bar(x: SomeClass) {}
bar(new SomeClass());
```

Current sickle output is something like:
```js
var foo_1 = require('./foo');

/** @param {SomeClass} x */
function bar(x) {}
bar(new foo_1.SomeClass());
```

The problem is that `SomeClass` is known as `foo_1.SomeClass` in the output, so the type annotation in the comment is wrong.  We need `foo_1.SomeClass` in the JSDoc.

One idea is to emit something like this in the TS:
```js
SomeClass; // sickle:SomeClass
```
which will become (after TS compilation)
```js
foo_1.SomeClass; // sickle:SomeClass
```
and then we could regex for these lines to build up the mapping between `SomeClass` and `foo_1.SomeClass`.  Pretty crazy hacks though.

Related: #111

martine:
Causes warning:
WARNING - Bad type annotation. Unknown type SomeClass

martine:
We could probably find a way to hack around this with some terrible regexes, but it's make more sense to fix it in the compiler.  So I filed this upstream:
https://github.com/Microsoft/TypeScript/issues/8120

martine:
I think a better fix for this is to undo the namespacing that TypeScript does.

E.g. the input is:
```js
import {foo} from ...;
let x = foo;
```

TypeScript produces:
```js
[module goop];
var x = bar_1.foo;
```

We translate that back to:
```js
[module goop]; var foo = bar_1.foo;
var x = foo;
```

To be honest, I'm not sure why TypeScript does the namespacing thing it does.

martine:
What's tricky about my proposal above is:

Input:
```js
import {foo} from 'foo';
import * as bar from 'bar';

let x = foo;
let y = bar.a;
```

Current output:
```js
var foo_1 = require('foo');
var bar = require('bar');
var x = foo_1.foo;
var y = bar.a;
```

I was suggesting identifying `foo_1.foo` as special and rewriting it, but somehow we'd need to know to leave `bar.a` alone.

mprobst:
Idea: if TSC choses a non-predictable prefix, can we just prime it on a prefix that we can predict?

```ts
import {Symbol} from 'module';
import * as prefix from 'module';

Symbol.access();
prefix.Symbol.access();
```

Gives:

```js
"use strict";
var module_1 = require('module');
var prefix = require('module');
[...]

PR #157, opened 2016-06-13
Configure typed/untyped converstion from the CLI and default to typed

jjudd:
tsickle was defaulting to untyped conversions, i.e., every TypeScript type
was converted to the Closure {?} type. Now it defaults to typed and allows
the user to configure whether types should be converted to {?} or an actual
type.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -12,6 +12,9 @@ import * as tsickle from './tsickle';
 interface Settings {
   /** If provided, path to save externs to. */
   externsPath?: string;
+
+  /** If provided, convert every type to the Closure {?} type */
+  isUntyped: boolean;
 }
 
 function usage() {
@@ -22,6 +25,7 @@ example:
 
 tsickle flags are:
   --externs=PATH     save generated Closure externs.js to PATH
+  --untyped          convert every type in TypeScript to the Closure {?} type
 `);
 }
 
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -19,6 +19,9 @@ const internalExternsFileName = 'tsickle_externs.js';
 interface Settings {
   /** If provided, path to save externs to. */
   externsPath?: string;
+
+  /** If provided, convert every type to the Closure {?} type */
+  isTyped: boolean;
```
martine: The comment seems to say the opposite of what the variable name indicates, maybe just a typo.

I think this would be better expressed as an optional "untyped" field, so that it can default to falsey (either unprovided or false).  At least, that's how it works in the underlying tsickle library.

martine:
I'm still working on typed mode.  It's maybe close to working but I think currently it produces code that doesn't compile. Did it work for you?

See e.g. https://github.com/angular/tsickle/issues/112 and https://github.com/angular/tsickle/issues/146 .

(Once it works, I intend to remove the untyped setting completely.)

src/main.ts
jjudd: Nice catch. That's what I get for copy pasting. I'll change it to untyped to keep it the same as the tsickle library.

jjudd:
Ah. So that's why that was there :) I prefer having a flag, so users can at least try to use typed mode without having to modify the source.

martine:
Makes sense, thanks for the patch!

PR #158, opened 2016-06-13
Stop writing the externs file twice

jjudd:
The externs were being written twice: once to their output location
specified on the CLI and again to './tsickle_externs.js'. This removes
tsickle_externs.js. The file appears to be a remnant of the closure
compilation part of the CLI, which has been removed.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -8,13 +8,6 @@ import * as ts from 'typescript';
 import * as cli_support from './cli_support';
 import * as tsickle from './tsickle';
 
-/**
- * Internal file name for the generated Closure externs.
- * This is never written to disk, but should be unique with respect to
- * the js files that may actually exist.
- */
-const internalExternsFileName = 'tsickle_externs.js';
-
 /** Tsickle settings passed on the command line. */
 interface Settings {
   /** If provided, path to save externs to. */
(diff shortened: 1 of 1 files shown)


mprobst:
@jjudd just FYI, @martine is out today, but will be back tomorrow and get to this :-)

jjudd:
Thanks. No rush. :)

src/main.ts
```diff
@@ -192,9 +185,7 @@ function toClosureJS(options: ts.CompilerOptions, fileNames: string[]):
     jsFiles[fileName] = output;
   }
 
-  jsFiles[internalExternsFileName] = tsickleExterns;
```
martine: Ah yep, this was to get the externs file into the list of input files to the closure command.

src/main.ts
```diff
@@ -222,7 +214,7 @@ function main(args: string[]) {
   }
 
   if (settings.externsPath) {
-    fs.writeFileSync(settings.externsPath, jsFiles[internalExternsFileName]);
+    fs.writeFileSync(settings.externsPath, externs);
```
martine: Since this is now conditional, it's worth highlighting that we may silently miscompile code if you forget to include externs.  I wonder if it's worth requiring the --externsPath flag.

jjudd: It might be worth just making it an argument, that way there is no confusion about it being required.

PR #159, opened 2016-06-13
In the CLI, do not run module converison on sourcemaps.

jjudd:
If you turn on sourcemaps in the TypeScript compiler, the CLI runs module conversion on them. This breaks the sourcemaps :( This change prevents them from being converted.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -187,9 +187,11 @@ function toClosureJS(options: ts.CompilerOptions, fileNames: string[]):
   }
 
   for (let fileName of Object.keys(jsFiles)) {
-    let {output} = tsickle.convertCommonJsToGoogModule(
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -187,9 +187,11 @@ function toClosureJS(options: ts.CompilerOptions, fileNames: string[]):
   }
 
   for (let fileName of Object.keys(jsFiles)) {
-    let {output} = tsickle.convertCommonJsToGoogModule(
-        fileName, jsFiles[fileName], cli_support.pathToModuleName);
-    jsFiles[fileName] = output;
+    if (path.extname(fileName) !== '.map') {
```
martine: This makes me wonder if we also choke on the .d.ts generated by the --declarations flag to tsc.

PR #171, opened 2016-06-28
don't overide typescript source argument, mkdirp before write

davityle:
Currently Tsickle will always transpile the whole program every time it
is run. This pull request adds support for transpiling only the files
supplied to ts via the command line. For example tsickle -- foobar.ts
would now only transpile foobar.ts instead of the whole program.

This also calls mkdirp on the directory path of a file before it is
written. Previously, if an output directory was given to the ts
transpiler via the command line (i.e. tsickle -- --outDir build/)
tsickle would be unsuccessful in writing the file if the entire
directory tree had not been created at the specified output path.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,6 +2,7 @@
 
 import * as fs from 'fs';
 import * as minimist from 'minimist';
+import * as mkdirp from 'mkdirp';
 import * as path from 'path';
 import * as ts from 'typescript';
 
@@ -86,6 +87,9 @@ function loadTscConfig(args: string[]):
     return {errors};
   }
 
+  // Store file arguments
+  let tsFileArguments = fileNames;
+
   // Read further settings from tsconfig.json.
   let projectDir = options.project || '.';
   let configFileName = path.join(projectDir, 'tsconfig.json');
@@ -99,6 +103,9 @@ function loadTscConfig(args: string[]):
   if (errors.length > 0) {
     return {errors};
   }
+
+  // if file arguments were given to the typescript transpiler than transpile only those files
+  fileNames = tsFileArguments.length > 0 ? tsFileArguments : fileNames;
 
   return {options, fileNames};
 }
@@ -222,10 +229,12 @@ function main(args: string[]) {
   }
 
   for (let fileName of Object.keys(jsFiles)) {
+    mkdirp.sync(path.dirname(fileName));
     fs.writeFileSync(fileName, jsFiles[fileName]);
   }
 
   if (settings.externsPath) {
+    mkdirp.sync(path.dirname(settings.externsPath));
     fs.writeFileSync(settings.externsPath, externs);
   }
 }
--- a/package.json
+++ b/package.json
@@ -11,7 +11,8 @@
   "dependencies": {
     "minimist": "^1.2.0",
     "source-map": "^0.4.2",
-    "source-map-support": "^0.3.1"
+    "source-map-support": "^0.3.1",
+    "mkdirp": "^0.5.1"
   },
   "peerDependencies": {
     "typescript": "^1.8.0 || ^1.9.0-dev"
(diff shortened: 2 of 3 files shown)


davityle:
I signed it!

src/main.ts
```diff
@@ -86,6 +87,9 @@ function loadTscConfig(args: string[]):
     return {errors};
   }
 
+  // Store file arguments
```
martine: This comment is redundant with the code that immediately follows it, don't you think?

martine:
Can you tell me why you'd want to only transpile some files out of a larger program?  Perhaps I can help you achieve your end goal better.

davityle:
We are going to have a process that watches for individual file changes and transpiles only the files that have changed. This is important because we plan to migrate 50k+ lines of code in a very short time and we would like to streamline this process as much as possible. We currently plan to use tools external to tsickle so specifying files via cli is crucial.

martine:
Suppose you have a file `a.ts`:
```js
export let foo = 3;
```

and another `b.ts`:
```js
export * from './a';
```

TypeScript implements the "export *" by generating a for loop over the exports of `a`.  This doesn't work in Closure (you can't dynamically create symbols), so at tsickle time we expand the `export *` into a line like:
```js
export {foo} from './a';
```

This means if you change "a" (e.g to export another symbol), you need to regenerate "b" as well to add that export to the export list.  (And if there's a library `c` that exports from `b`, you'll need to rebuild it as well, and so on.)

There are a few other places where we have global effects.  This is why tsickle takes a tsconfig as input -- it has to typecheck your entire project to be able to generate any of it.  I think the actual file emission is probably pretty fast compared to that.  I think the best practical thing is to just regenerate everything on any change; the Closure compilation step afterwards is tens of seconds anyway so it doesn't save you much.

With all of that said, if you still think this change will work for you, I'm not opposed to it.  I just wanted to be sure you understood the potential consequences.

jjudd:
I'm curious if we could accomplish a similar goal by using the tsconfig and somehow supporting the `--watch` flag in tsickle. It seems odd and kind of counter intuitive to say that tsickle is a wrapper around tsc and then have behavior completely different from tsc.

martine:
TypeScript has this principle that the output doesn't depend on types at all (I wrote a blog post about it! http://neugierig.org/software/blog/2016/04/typescript-types.html) which means it's capable of emitting files independently.  To massage TS into Closure form (for stuff like the above), tsickle unfortunately must depend on the types, so it necessarily has different behavior here.

Should I merge this?  Or have I scared you off?  :)

davityle:
Our use case is for development purposes only. We won't be running the code through the closure compiler during this process. Essentially we just need our typescript to be transpiled to js with google closure integration (i.e. goog.require etc), so that we can run it in the browser with our current code base. If the watcher only transpiled changed files than, as you mentioned, it would not work if exports changed and would require the developer to manually run tsickle if they changed some external interface on a file. However, as you said 

>  I think the actual file emission is probably pretty fast 

this could be a premature optimization. What are your thoughts?

martine:
Aaaah developer mode I completely forgot about it!

I actually am not super familiar with how this works (before my time on this project) but I believe all you need to do there is:
1. transpile the files using plain TSC
2. call the one function that does commonjs->goog.require munging, see https://github.com/angular/tsickle/blob/abc8ef129d12490c8c68358446507a5c8e4c2173/src/es5processor.ts#L6 

Which means you still want this code:
https://github.com/angular/tsickle/blob/abc8ef129d12490c8c68358446507a5c8e4c2173/src/main.ts#L198
but you skip lines 157 through 186 above it (which are the parts that rely on all the intermodule dependency junk).  That's how we do it inside the Google build system.

I think you can just add a tsickle flag for it, like "developermode" or something.

@mprobst I think there's also something about module manifests for devmode.  Should this be part of tsickle?  What generates the closure deps.js in devmode?

In any case, all of the above is an optimization that can be done later, and this change is correct in devmode as-is.

martine:
Moved the above discussion of developer mode to issue #172.

PR #182, opened 2016-07-26
workaround closure issue: always produce ctorParameters

alexeagle:


alexeagle:
the closure issue is that
```
class Base {}
class Derived extends Base {}
```
=>
```
/**
 @struct
 @constructor
 @extends {Base}
 */
var Derived = function(var_args) {
  Base.apply(this, arguments);
};
$jscomp.inherits(Derived, Base);
```

Angular 2 looks at the count of params to a constructor to know whether it needs to do dependency injection.
Since Derived now has non-zero count of params, the dependency injector gives an error that it does not have bindings to know how to call the constructor.

evmar:
LGTM.

Due to the @nocollapse, I suspect this will never get dead code eliminated. :~(

alexeagle:
No, but ultimately, Angular 2 apps will have all of the metadata consumed by the compiler, and produced into codegen. We will no longer want any of the metadata repeated in static fields. Then we'll want a 'no emit' option for decorators, and stop using tsickle for this altogether.

PR #185, opened 2016-07-27
ctorParameters can be of any type, not just Function

### LINKED ISSUE #184: decorator-annotator emit doesn't permit injecting all values
Angular allows any type to be provided for a binding, so the signature of an injection site can be eg. `@Inject('port') port:number`

Here's a live example: http://plnkr.co/edit/GNAyj6K6PfYg2NBzgwZ5?p%3Dpreview&p=preview

tsickle currently types these as `Function`

alexeagle:
Fixes #184


evmar:
lgtm

PR #192, opened 2016-08-26
generate module.id in ES6 mode that can be optimized out by closure.

vikerman:
The current module.id generation is careful to preserve the global module if it's already present. However in that form closure cannot optomize it out even if module.id is not referenced in the file - Leading to the final bundle bloat.

The fix is to preserve the existing module only in ES5 mode and just output module.id in ES6 mode(where we don't expect module to already exist).

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -6,7 +6,8 @@ import {toArray} from './util';
 
 /**
  * ES5Processor postprocesses TypeScript compilation output JS, to rewrite commonjs require()s into
- * goog.require().
+ * goog.require(). Contrary to it's name it handles converting the modules in both ES5 and ES6
+ * outputs.
  */
 class ES5Processor extends Rewriter {
   /**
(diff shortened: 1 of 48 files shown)


src/es5processor.ts
```diff
@@ -35,8 +35,8 @@ class ES5Processor extends Rewriter {
   unusedIndex: number = 0;
 
   constructor(
-      file: ts.SourceFile,
-      private pathToModuleName: (context: string, fileName: string) => string) {
+      file: ts.SourceFile, private pathToModuleName: (context: string, fileName: string) => string,
+      private isES5: boolean) {
```
rkirov: isES5 can be confusing given this class is called ES5 processor. Add some documentation to explain that this class sometimes consumes es6 code that still has commonjs module output, as a preprocessor for closure optimization.

rkirov:
LGTM

src/es5processor.ts
vikerman: Done. Changed doc of class to indicate it handles both ES5 and ES6,

src/es5processor.ts
```diff
@@ -6,7 +6,8 @@ import {toArray} from './util';
 
 /**
  * ES5Processor postprocesses TypeScript compilation output JS, to rewrite commonjs require()s into
- * goog.require().
+ * goog.require(). Contrary to it's name it handles converting the modules in both ES5 and ES6
```
evmar: s/it's/its/

vikerman: Fixed

PR #193, opened 2016-08-27
Don't add @nocollapse closure annotation to decorators and propDecora…

vikerman:
…tors but only to ctrParameters which is used for DI.


alexeagle:
LGTM

PR #195, opened 2016-08-29
update to typescript@2.1.0-dev.20160827

IgorMinar:


package.json
```diff
@@ -26,14 +26,14 @@
...
-    "gulp-tslint": "^4.3.4",
+    "gulp-tslint": "^6.1.1",
     "gulp-typescript": "^2.7.6",
     "gulp-util": "^3.0.4",
     "merge2": "^0.3.1",
     "temp": "^0.8.1",
     "tsd": "^0.6.0",
-    "tslint": "^3.7.1",
-    "typescript": "2.0.0"
+    "tslint": "^3.15.0-dev.0",
```
IgorMinar: because of peerDeps I had to bump tslint as well

src/decorator-annotator.ts
```diff
@@ -51,7 +51,7 @@ class ClassRewriter extends Rewriter {
   }
 
   private decoratorsToLower(n: ts.Node): ts.Decorator[] {
-    return (n.decorators || []).filter(d => this.shouldLower(d));
+    return (n.decorators as ts.NodeArray<ts.Decorator>|| []).filter(d => this.shouldLower(d));
```
IgorMinar: I'm not quire sure why this change is required

src/tsickle.ts
```diff
@@ -749,18 +749,7 @@ class ExternsWriter extends ClosureRewriter {
...
-                name = getIdentifierText(decl.name as ts.Identifier);
-                break;
-              case ts.SyntaxKind.StringLiteral:
-                name = (decl.name as ts.StringLiteral).text;
-                break;
-              default:
-                this.errorUnimplementedKind(decl.name, 'namespace name');
-                break;
-            }
+            let name = getIdentifierText(decl.name as ts.Identifier);
```
IgorMinar: decl.name.kind is already of type ts.SyntaxKind.Identifier so this whole switch block is not needed.

src/type-translator.ts
```diff
@@ -27,17 +27,27 @@ function isClosureProvidedType(symbol: ts.Symbol): boolean {
...
-    ts.TypeFlags.Any,           ts.TypeFlags.String,
-    ts.TypeFlags.Number,        ts.TypeFlags.Boolean,
-    ts.TypeFlags.Void,          ts.TypeFlags.Undefined,
-    ts.TypeFlags.Null,          ts.TypeFlags.Enum,
-    ts.TypeFlags.StringLiteral, ts.TypeFlags.TypeParameter,
-    ts.TypeFlags.Class,         ts.TypeFlags.Interface,
-    ts.TypeFlags.Reference,     ts.TypeFlags.Tuple,
-    ts.TypeFlags.Union,         ts.TypeFlags.Intersection,
-    ts.TypeFlags.Anonymous,     ts.TypeFlags.Instantiated,
-    ts.TypeFlags.ObjectLiteral, ts.TypeFlags.ESSymbol,
```
IgorMinar: I just removed `ObjectLiteral` because that's not marked as `@internal` - I don't know what kind of implications this has on the rest of the code - I couldn't find a good replacement for this type in typescript.d.ts. please advice if this is the right change.

tsconfig.json
```diff
@@ -10,7 +10,6 @@
     "strictNullChecks": true
   },
   "files": [
-    "node_modules/typescript/lib/lib.es2015.collection.d.ts",
```
IgorMinar: this is not needed since we use `lib` above

src/decorator-annotator.ts
mprobst: I think `node.decorators` ends up as some odd union type because it's defined multiple times in the dependency hierarchy of `ts.Node`. No idea why it happens in this version though.

src/tsickle.ts
```diff
@@ -749,18 +749,7 @@ class ExternsWriter extends ClosureRewriter {
...
             if (name === undefined) break;
```
mprobst: can name still be `undefined` here, then?

src/type-translator.ts
mprobst: That's _now_ marked as `@internal`? No idea, maybe @chuckjaz knows?

rkirov:
closing in favor of https://github.com/angular/tsickle/pull/196

src/type-translator.ts
evmar: This function is only used for debugging anyway so it's likely fine.

PR #202, opened 2016-09-09
feature(annotation): annotate types of getter and setter methods

alexeagle:


evmar:
LGTM

alexeagle:
merged as 03521fc thanks

PR #206, opened 2016-09-14
drop tsd, use @types

evmar:
Also make gulp build use tsconfig.json, rather than having
its own copy of compiler flags.


tsconfig.json
```diff
@@ -3,17 +3,19 @@
     "experimentalDecorators": true,
     "module": "commonjs",
     "noImplicitAny": true,
-    "noEmit": true,
+    "noEmitOnError": true,
```
alexeagle: did you mean to submit this?

evmar: yes, we want to emit when tsc runs this.

PR #207, opened 2016-09-14
Replace illegal characters when converting to goog.module

evmar:


evmar:
@jjudd I'm merging your tsickle changes.

src/cli_support.ts
```diff
@@ -2,8 +2,16 @@ import * as path from 'path';
...
+  fileName = fileName.replace(/\.js$/, '');
+
   if (fileName[0] === '.') {
+    // './foo' or '../foo'.
+    // Resolve the path against the dirname of the current module.
     fileName = path.join(path.dirname(context), fileName);
   }
-  return fileName.replace(/\.js$/, '').replace(/\//g, '.');
+  // Replace characters not supported by goog.module.
+  let moduleName =
```
alexeagle: we use a different, reversible escaping scheme for goog.module internally, where we URI escape and then replace % with $. Should we use the same?

evmar: I dunno.  :~(  I figure this worked for @jjudd so maybe we should just merge it unless there are problems.

alexeagle: okay, it doesn't need to be identical, just seems like it's yet another moving part somewhere...

jjudd:
@evmar Thanks for doing this! If you have any questions or concerns please let me know. There's a few tsickle changes we have been meaning to commit back upstream, but haven't had the time to do so yet. :)

PR #208, opened 2016-09-16
generate typedefs in externs

evmar:
Given a type alias in TypeScript, generate a Closure typedef.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -828,11 +828,14 @@ class ExternsWriter extends ClosureRewriter {
         break;
       case ts.SyntaxKind.VariableStatement:
         for (let decl of (<ts.VariableStatement>node).declarationList.declarations) {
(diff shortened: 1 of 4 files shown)


src/tsickle.ts
```diff
@@ -926,22 +929,27 @@ class ExternsWriter extends ClosureRewriter {
...
+      this.emit('\n');
+      this.writeExternsVariable(name, namespace);
     } else {
       this.errorUnimplementedKind(decl.name, 'externs for variable');
     }
   }
 
+  private writeExternsVariable(name: string, namespace: string[]) {
+    let qualifiedName = namespace.concat([name]).join('.');
+    if (namespace.length > 0) {
```
mprobst: what about:

```
if (namespace.length === 0) emit('var');
this.emit(`${qualifiedName};\n`);
```

evmar: Good catch!

Issue #180, opened 2016-07-26
Handle function overloads, perhaps by unifying signatures


evmar:
Tsickle currently errors when generating externs from a declaration class with multiple constructors.
However, these exist in d.ts files.  We need to figure out how to support it.

Example:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56024034ccc426af2840e90fe94043c2213dd6b4/google.visualization/google.visualization.d.ts#L238

mprobst:
When declaring the implementation of an overloaded function/method/ctor,
you need to provide an implementation that takes the union of all
parameters, and returns the union of all return types (and then TS trusts
you that you return the right type for the right parameter set).

Sounds like we might have to do this merge ourselves?

evmar:
Some more instances of this problem:
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/56024034ccc426af2840e90fe94043c2213dd6b4/threejs/three.d.ts#L976
- https://github.com/Reactive-Extensions/RxJS/blob/29f85627e4ab71b050480f26dcc5d737f1b840f4/ts/rx.lite.d.ts#L254
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/pixi.js/pixi.js.d.ts#L394

Doing the merge looks like it'd generate some pretty crazy types, especially in that first case.
But I don't see any way around it.

mprobst:
Yes, the types are crazy. But OTOH, that really is the runtime type of the
method without overloads, which is probably what Closure should see, right?

evmar:
Under the rationale of "TypeScript has already type-checked", I might just do something like
```js
/**
 * @constructor
 * @param {...*} args
 */
function Foo(args) {}
```
for now.

evmar:
See for example
  https://github.com/Microsoft/monaco-editor/blob/master/website/playground/monaco.d.ts.txt
which defines more than ten overloads for Promise.then.

hess-g:
From my whiteboard notes on what the generator would need to cover:
- Maximum number of args would be the count of the function(s) with the largest count.
- All args would be "_opt" at the index beyond the minimum count across all functions.
- name "x_or_y" if different names at same index.
- {Tx | Ty} if different types at same index

And a few open questions:
- Some special naming convention for variadic?
- How to handle mix of nullable and not nullable at same ordinal index.

[...]

PR #210, opened 2016-09-18
Add golden test data for overloaded constructors and methods

hess-g:
#Adds golden test code for overloaded constructor and method patterns - #180.
Includes placeholder rendering of first method when more than  
signature is found for a given method name (like existing >1
constructor placeholder).


hess-g:
Going to clean up my local environment and give this another shot.

PR #211, opened 2016-09-19
Add golden test data for overloaded constructors and methods

hess-g:
#180 - Adds golden test code for overloaded constructor and method
patterns.


src/tsickle.ts
```diff
@@ -888,6 +888,36 @@ class ExternsWriter extends ClosureRewriter {
       this.writeExternsFunction(name.getText(), paramNames, namespace);
     }
 
+    // Handle method declarations/signatures first, since we need to deal with overloads.
+    // TODO: Build a collection of the remaining members
+    let methods: {[id: string]: ts.MethodDeclaration[];} = {};
```
evmar: No semi in the {}.

src/tsickle.ts
```diff
@@ -888,6 +888,36 @@ class ExternsWriter extends ClosureRewriter {
...
+    for (let member of decl.members) {
+      if (member.kind === ts.SyntaxKind.MethodSignature ||
+          member.kind === ts.SyntaxKind.MethodDeclaration) {
```
evmar: Is it worth handling the "else" case, if only to report an issue?  Sometimes we use switch statements to make it clear what the default case is.  There's `this.debugWarn()` which is intended for tsickle-internal TODOs (it's never shown to users of tsickle).

src/tsickle.ts
evmar: What is "id" here?  Maybe say "methodName" instead?

src/tsickle.ts
```diff
@@ -888,6 +888,36 @@ class ExternsWriter extends ClosureRewriter {
...
+        let m = <ts.MethodDeclaration>member;
+        let overloaded = methods[m.name.getText()];
```
evmar: getText() tries to get the literal source text of the node, and it turns out to not be usually what you want.  It appears m.name is
```
    export type PropertyName = Identifier | LiteralExpression | ComputedPropertyName;
```
Maybe just handle `m.name.kind === ts.SyntaxKind.Identifier` for now and punt on the others?

src/tsickle.ts
```diff
@@ -888,6 +888,36 @@ class ExternsWriter extends ClosureRewriter {
...
+          methods[m.name.getText()] = [m];
+        }
+      }
+    }
+    for (let methodName in methods) {
+      if (methods.hasOwnProperty(methodName)) {
+        let method: ts.MethodDeclaration[] = methods[methodName];
+        let rootMethod = method[0];
+        if (method.length > 1) {
+          this.debugWarn(method[0], 'overloaded method signatures found');
```
evmar: Another option (your call) is to
```js
this.emit(`/* TODO: method overloaded */\n`);
```
as that shows up in the golden file and is obvious to humans.

src/tsickle.ts
```diff
@@ -903,12 +933,6 @@ class ExternsWriter extends ClosureRewriter {
           break;
         case ts.SyntaxKind.MethodSignature:
         case ts.SyntaxKind.MethodDeclaration:
-          let m = <ts.MethodDeclaration>member;
-          this.emitFunctionType(m);
-          this.emit(
-              `${typeName}.prototype.${m.name.getText()} = ` +
```
evmar: Oh, I see the getText() was my own fault.  Feel free to ignore my other comment. :P

hess-g: Yes. If only the blame layer carried along copypasta issues. :-)

src/tsickle.ts
hess-g: Ignoring this comment per your later one. I do need the method name here since it is the unique key for the map of 1...n methods variants I'm building.

src/tsickle.ts
hess-g: I changed this in the follow up commit. The Signature|Declaration now are added to a separate array when we walk the full list of members. However, I missed removing this recheck. Another commit is coming.

src/tsickle.ts
hess-g: Missed this comment in my first follow up commit; additional one coming.

hess-g: Also, habit. When I've used this ES5 dictionary/map pattern, I've typically left it as id or key, since that is what it is acting as. Changed for clarity here.

src/tsickle.ts
hess-g: Done.

PR #212, opened 2016-09-19
make the "declare" test use a d.ts

evmar:
This is in preparation for the next change, where it turns
out that the file name matters.  The test suite had no d.ts
files before this change so I had to make a minor change to
path handling.


rkirov:
looks like you need to change the closure invocation too (failing on travis)

evmar:
PTAL

rkirov:
LGTM

PR #213, opened 2016-09-19
everything in d.ts should show up in externs

evmar:
Even if there's an "interface Foo" without a "declare" on
it, if it's in a d.ts file, it needs to show up in externs.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -75,6 +75,10 @@ function isValidClosurePropertyName(name: string): boolean {
   return /^[a-zA-Z_][a-zA-Z0-9_]*$/.test(name);
 }
 
+function isDtsFileName(fileName: string): boolean {
+  return /\.d\.ts$/.test(fileName);
+}
+
 const VISIBILITY_FLAGS = ts.NodeFlags.Private | ts.NodeFlags.Protected | ts.NodeFlags.Public;
 
 /**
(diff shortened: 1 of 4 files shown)


src/tsickle.ts
```diff
@@ -300,7 +300,7 @@ class Annotator extends ClosureRewriter {
    *     emit it as is and visit its children.
    */
   maybeProcess(node: ts.Node): boolean {
-    if (node.flags & ts.NodeFlags.Ambient) {
+    if ((node.flags & ts.NodeFlags.Ambient) || /\.d\.ts$/.test(this.file.fileName)) {
```
evmar: I checked and the node.flags on a d.ts file are not Ambient.  :~(

rkirov: extract to a helper isDeclarationFile, to resist writing the regexp on multiple occasions.

src/tsickle.ts
```diff
@@ -779,6 +779,12 @@ class ExternsWriter extends ClosureRewriter {
   /** visit is the main entry point.  It generates externs from a ts.Node. */
   public visit(node: ts.Node, namespace: string[] = []) {
     switch (node.kind) {
+      case ts.SyntaxKind.SourceFile:
```
rkirov: is this intentionally part of this PR? Not a big deal, just making sure you are aware.

evmar: Yes.  We now pass a SourceFile into this visit() function, where before we'd only pass every "declare ..." statement into it.

evmar:
PTAL

rkirov:
Lgtm

Issue #214, opened 2016-09-20
Debugger in vscode fails with "Attribute 'program' does not exist"


hess-g:
Running the debugger in VSCode results in the following error:
"Attribute 'program' does not exist ([workspace_root]/node_modules/gulp-mocha/node_modules/mocha/bin/_mocha).
(root path is correctly expanded)

Change to: 
"program": "${workspaceRoot}/node_modules/mocha/bin/_mocha"
and the debugger runs as expected.

Repro'd in freshly cloned project.

evmar:
Weird, this used to work and "npm ls" shows that mocha is inside gulp-mocha:
```
├─┬ gulp-mocha@2.2.0
│ ├─┬ mocha@2.5.3
```
but it's clearly where you say it is.  I clearly don't understand npm.

hess-g:
And an interesting oddity - You have to run "gulp test" at least once on a newly cloned repo, and then the debugger works with this new path. Note that it never worked with the old path, at least not for me.

hess-g:
The npm ls shows dependencies. npm makes some judgement calls at install time as to whether a particular dependency needs to be physically installed or not. In this repos case, mocha is getting installed at the root packages level, and therefore is already present for gulp-mocha to use. I've tried to find definitive docs on the algorithm, but they didn't come up on a cursory look.

hess-g:
Fixed by #215.

PR #215, opened 2016-09-20
Update launch.json to call mocha directly.

hess-g:
#214
- Change path to call mocha binary directly.

(unrelated)
- Update to new “console” attribute, set to “internalConsole”


hess-g:
Fixes #214.

PR #217, opened 2016-09-20
use an ES6 map to avoid the hasOwnProperty mess of objects

### LINKED ISSUE #216: Method overload check needs to use hasOwnProperty
https://github.com/angular/tsickle/blob/6bdd1eaba35a7aed655f7296aaafaa751391de8c/src/tsickle.ts#L922

This does the wrong thing on methods named `valueOf` .

@hess-g

evmar:
Fixes #216.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -892,7 +892,7 @@ class ExternsWriter extends ClosureRewriter {
     }
 
     // Process everything except (MethodSignature|MethodDeclaration|Constructor)
-    let methods: {[methodName: string]: ts.MethodDeclaration[]} = {};
+    let methods: Map<string, ts.MethodDeclaration[]> = new Map();
     for (let member of decl.members) {
       switch (member.kind) {
         case ts.SyntaxKind.PropertySignature:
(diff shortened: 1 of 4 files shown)


evmar:
@hess-g

evmar:
ping @rkirov, don't forget this one

src/tsickle.ts
```diff
@@ -908,12 +908,12 @@ class ExternsWriter extends ClosureRewriter {
...
         case ts.SyntaxKind.MethodSignature:
         case ts.SyntaxKind.MethodDeclaration:
-          let method = <ts.MethodDeclaration>member;
-          let overloaded = methods[method.name.getText()];
-          if (overloaded) {
-            methods[method.name.getText()].push(method);
+          const method = member as ts.MethodDeclaration;
+          const methodName = method.name.getText();
+          if (methods.has(methodName)) {
+            methods.get(methodName)!.push(method);
```
rkirov: I wish they added map.setdefault method, like in python, then you can do `map.setDefault(key, []).push(...)`. 

no change needed.

src/tsickle.ts
```diff
@@ -933,22 +933,19 @@ class ExternsWriter extends ClosureRewriter {
...
-          // TODO: Pass to overload rendering parameter builder
-          // For now, we just drop all but the first.
-          // See https://github.com/angular/tsickle/issues/180 .
-          this.debugWarn(method[0], 'overloaded method signatures found');
-          this.emit(`/* TODO(tsickle:#180): Method overloaded; only adding first signature. */\n`);
-        }
-        this.emit(
-            `${typeName}.prototype.${rootMethod.name.getText()} = ` +
-            `function(${rootMethod.parameters.map((p) => p.name.getText()).join(', ')}) {};\n`);
+    for (const method of Array.from(methods.values())) {
```
rkirov: thats a nice way to deal with TS flaw with iterables and for-of.

src/tsickle.ts
mprobst: ... and then you get the same list on every access, and get lots of obscure errors where multiple keys point to the same list o_O.

evmar: Nah, setDefault() is poorly named, it means "get and set this as the value if it's not already there and return it".  Separate value for each call.

PR #221, opened 2016-09-20
avoid the parameter name 'arguments'

### LINKED ISSUE #219: Avoid the variable name "arguments"
```js
function foo(arguments: string) {
}
```

is legal TS, but in Closure:

```
input0:1: ERROR - Shadowing "arguments" is not allowed
function foo(arguments) {
             ^
```

evmar:
Refactor method writing to pass the argument list as an array
to a shared function, then filter the argument list in that
function.

Fixes #219.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -75,6 +75,19 @@ function isValidClosurePropertyName(name: string): boolean {
   return /^[a-zA-Z_][a-zA-Z0-9_]*$/.test(name);
 }
 
+/**
+ * TypeScript allows parameters named "arguments", but Closure
+ * disallows this, even in externs.
+ */
+function makeLegalExternsParam(name: string): string {
+  switch (name) {
+    case 'arguments':
+      return 'tsickle_arguments';
+    default:
+      return name;
+  }
+}
+
 function isDtsFileName(fileName: string): boolean {
   return /\.d\.ts$/.test(fileName);
 }
@@ -126,7 +139,7 @@ class ClosureRewriter extends Rewriter {
         let newTag: jsdoc.Tag = {
           tagName: 'param',
           optional: paramNode.initializer !== undefined || paramNode.questionToken !== undefined,
-          parameterName: unescapeName(paramSym.getName()),
+          parameterName: makeLegalExternsParam(unescapeName(paramSym.getName())),
         };
 
         let destructuring =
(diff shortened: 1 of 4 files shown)


evmar:
CC @hess-g

src/tsickle.ts
```diff
@@ -126,7 +139,7 @@ class ClosureRewriter extends Rewriter {
         let newTag: jsdoc.Tag = {
           tagName: 'param',
           optional: paramNode.initializer !== undefined || paramNode.questionToken !== undefined,
-          parameterName: unescapeName(paramSym.getName()),
+          parameterName: makeLegalExternsParam(unescapeName(paramSym.getName())),
         };
```
hess-g: Maybe add:  

```
let originalParameterName = unescapeName(paramSym.getName());
// Use in call to makeLegalExternsParam

if (newTag.parameterName != originalParameterName) {
  newTag.text = 'tsickle: renamed. ' + paramSym.getName() ' is disallowed in closure.'
}
```

src/tsickle.ts
```diff
@@ -75,6 +75,19 @@ function isValidClosurePropertyName(name: string): boolean {
...
 }
 
+/**
+ * TypeScript allows parameters named "arguments", but Closure
+ * disallows this, even in externs.
+ */
+function makeLegalExternsParam(name: string): string {
+  switch (name) {
+    case 'arguments':
+      return 'tsickle_arguments';
```
hess-g: Change to `arguments_tsickle` so that the original intended name is more clearly readable?

src/tsickle.ts
evmar: It's not a bad idea.  So far we haven't been putting in human-readable comments into the JS (except in cases where it helps read the results of the test suite) because we don't intend for humans to use the resulting output.  I will file a bug on this idea.

src/tsickle.ts
evmar: I was attempting to match the other _tsickle-generated variable, but now that I see it, I see I got it wrong.

rkirov:
LGTM

src/tsickle.ts
evmar: https://github.com/angular/tsickle/issues/222

hess-g: Ah, right. My thought would be any real change to the original values warrants at least some comment "tsickle changed x because...closure." Human readable or not...some silly human will try to read it.  Worth an issue discussion to explore, so it's consistent across the output.

src/tsickle.ts
evmar: Renamed `_tsickle_arguments` to match `_tsickle_typeAnnotationsHelper`.

PR #224, opened 2016-09-22
Initial generation of overloads

hess-g:
#180
- Renders existing golden test overloaded ctors and class functions.


src/tsickle.ts
```diff
@@ -957,18 +1040,18 @@ class ExternsWriter extends ClosureRewriter {
...
-        this.debugWarn(method[0], 'overloaded method signatures found');
-        this.emit(`/* TODO(tsickle:#180): Method overloaded; only adding first signature. */\n`);
+    for (const methodVariants of Array.from(methods.values())) {
+      let firstMethodVariant = methodVariants[0];
+      let parameterNames: string[];
+      if (methodVariants.length > 1) {
+        parameterNames = this.emitFunctionType(methodVariants);
+      } else {
+        parameterNames = this.emitFunctionType([firstMethodVariant]);
+        // this.writeExternsFunction(
```
hess-g: TODO: remove comments

hess-g:
I'm unplugging for the night. I'll look to address comments in the AM (I'm EDT). 

Not sure why 5 files that I didn't change and that match master head are failing lint in the travis run; `npm test` passes locally.

src/tsickle.ts
```diff
@@ -105,89 +105,173 @@ class ClosureRewriter extends Rewriter {
...
+            }
+            // push to the newDoc for rendering the single function
+            newDoc.push(newTag);
+          } else {
+            // either add an entry at this index, or update the existing name and/or type as needed
+            // merge parameter names at this index if required
+            if (newTag.parameterName) {
+              if (!paramData.names[i]) {
+                paramData.names[i] = newTag.parameterName;
+              } else if (paramData.names[i].indexOf(newTag.parameterName) === -1) {
```
hess-g: Just build the list here, and join in the last step. This check would skip params if the later one was contained in an earlier: function(aa, a) would end up with only param "aa".

src/tsickle.ts
```diff
@@ -105,89 +105,173 @@ class ClosureRewriter extends Rewriter {
...
+              }
+              if (i >= minCount && paramData.names[i].indexOf('opt_') !== 0) {
+                paramData.names[i] = 'opt_' + paramData.names[i];
+              }
+            }
+            // merget type names at this index if required
+            if (newTag.type) {
+              if (!paramData.types[i]) {
+                paramData.types[i] = newTag.type;
+              } else if (paramData.types[i].indexOf(newTag.type) === -1) {
```
hess-g: Just build the list here, and join in the last step. This check would skip types if the later one was contained in an earlier: 
```
function(a: Array<SomeType>)
function(b: SomeType
```
would end up with only {Array<SomeType>}

src/tsickle.ts
```diff
@@ -105,89 +105,173 @@ class ClosureRewriter extends Rewriter {
...
+            // either add an entry at this index, or update the existing name and/or type as needed
+            // merge parameter names at this index if required
+            if (newTag.parameterName) {
+              if (!paramData.names[i]) {
+                paramData.names[i] = newTag.parameterName;
+              } else if (paramData.names[i].indexOf(newTag.parameterName) === -1) {
+                paramData.names[i] += '_or_' + newTag.parameterName;
+              }
+              if (i >= minCount && paramData.names[i].indexOf('opt_') !== 0) {
+                paramData.names[i] = 'opt_' + paramData.names[i];
```
hess-g: Add prefix in last step. Also, remove an existing instance of "opt_" from any  names, so that we only end up with one.

src/tsickle.ts
```diff
@@ -105,89 +105,173 @@ class ClosureRewriter extends Rewriter {
...
+      // Construct the JSDoc comment by reading the existing JSDoc, if
+      // any, and merging it with the known types of the function
+      // parameters and return type.
+      let jsDoc = this.getJSDoc(fnDecl) || [];
+
+      // Copy all the tags other than @param/@return into the new
+      // comment without any change; @param/@return are handled later.
+      for (let tag of jsDoc) {
+        if (tag.tagName === 'param' || tag.tagName === 'return') continue;
+        newDoc.push(tag);
```
hess-g: Since overloads might have the same tag in their docs, check to see that we haven't already added an existing tag to the new stack.

src/tsickle.ts
hess-g: Done

src/tsickle.ts
hess-g: Done

src/tsickle.ts
hess-g: Done

hess-g:
Travis should clear when #226 is merged.

src/tsickle.ts
hess-g: Done.

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+    const abstract = {tagName: 'abstract'};
+    const isOverloaded = fnDecls.length > 1;
+    const isConstructor = fnDecls[0].kind === ts.SyntaxKind.Constructor;
+
+    for (let fnDecl of fnDecls) {
+      if (fnDecl.parameters.length < minCount) {
+        minCount = fnDecl.parameters.length;
+      }
+      if (fnDecl.parameters.length > maxCount) {
+        maxCount = fnDecl.parameters.length;
```
evmar: FWIW you can use something like Math.max(...arrayOfNumbers), but I am not sure I prefer doing this differently.
```
let lens = fnDecl.paramaters.map(p => p.length);
const minCount = Math.min(...lens);
const maxCount = Math.max(...lens);
```
What do you think?

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
-      newDoc.push({tagName: 'abstract'});
-    }
+    for (let fnDecl of fnDecls) {
+      // Construct the JSDoc comment by reading the existing JSDoc, if
+      // any, and merging it with the known types of the function
+      // parameters and return type.
+      let jsDoc = this.getJSDoc(fnDecl) || [];
+
+      // Copy all the tags other than @param/@return into the new
+      // comment without any change; @param/@return are handled later.
```
evmar: If there are overloads, does this mean we end up with multiple copies of the @foo listed on them?  E.g. if each overload has @see http://foo on them, we'll have multiple?

(I'm not sure there's any @foo we even care about preserving here though!)

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+      if ((fnDecl.flags & ts.NodeFlags.Abstract) && (newDoc.indexOf(abstract) === -1)) {
+        newDoc.push(abstract);
+      }
+      // Merge the parameters into a single list of merged names and list of types
+      let sig = typeChecker.getSignatureFromDeclaration(fnDecl);
+      // Parameters.
+      if (sig.parameters.length) {
+        // Iterate through both the AST parameter list and the type's parameter
+        // list, as some information is only available in the former.
+        for (let i = 0, l = sig.parameters.length; i < l; i++) {
```
evmar: Is the "l = sig.parameters.length" thing for performance?  If so, I'd prefer to leave it to profiling to optimize.

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+            // either add an entry at this index, or update the existing name and/or type as needed
+            // merge parameter names at this index if required
+            if (newTag.parameterName) {
+              if (!paramData.names[i]) {
+                paramData.names[i] = [newTag.parameterName];
+              } else if (paramData.names[i].indexOf(newTag.parameterName) === -1) {
+                paramData.names[i].push(newTag.parameterName);
+              }
+            }
+            // merget type names at this index if required
```
evmar: typo: "merge"

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+            if (newTag.type) {
+              if (!paramData.types[i]) {
+                paramData.types[i] = [newTag.type];
+              } else if (paramData.types[i].indexOf(newTag.type) === -1) {
+                paramData.types[i].push(newTag.type);
```
evmar: If it helps any, you can use ES6's `Set` type.  What you have is fine though.

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+  /**
+   * Handles emittng the jsdoc for methods, including overloads.
+   * If overloaded, merges the signatures in the list of SignatureDeclarations into a single jsdoc.
+   * - Total number of parameters will be the maximum count found across all variants.
+   * - Any parameters beyond the minimum count found across all variants; prefix: "opt_"
+   * - Different names at the same parameter index will be joined with "_or_"
+   * - Variable args (...type[] in TypeScript) will be output as "...type",
+   *    except if found at the same index as another argument.
+   */
+  emitFunctionType(fnDecls: ts.SignatureDeclaration[], extraTags: jsdoc.Tag[] = []): string[] {
```
evmar: Can you mention in the comment about what this returns?  It seems to be the parameter names?

src/tsickle.ts
```diff
@@ -105,89 +105,177 @@ class ClosureRewriter extends Rewriter {
...
+          paramNamesToReturn[i] = 'opt_' + paramNamesToReturn[i];
         }
+        let concatenatedTypes = paramData.types[i].join('|');
+        newDoc.push({
+          tagName: 'param',
+          parameterName: paramNamesToReturn[i],
+          type: concatenatedTypes,
+        });
+      }
+      if (!isConstructor) {
```
evmar: Is it possible to just put the containing block in the earlier "if !isConstructor" check?

src/tsickle.ts
```diff
@@ -896,20 +984,19 @@ class ExternsWriter extends ClosureRewriter {
...
           if (ctors.length > 1) {
-            // TODO: Pass to overload rendering parameter builder
-            // For now, we just drop all but the first.
-            // See https://github.com/angular/tsickle/issues/180 .
-            this.debugWarn(ctors[1], 'multiple constructor signatures in declarations');
+            paramNames = this.emitFunctionType(
+                ctors as ts.ConstructorDeclaration[],
+                [{tagName: 'constructor'}, {tagName: 'struct'}]);
+          } else {
+            paramNames = this.emitFunctionType([firstCtor]);
```
evmar: Does this mean if there is an overloaded constructor, it's not tagged with @constructor?

src/tsickle.ts
hess-g: Hmm. Idiot mode. The new jsDoc instance should be outside the iteration over the variants, to pick up tags that are on one variant but not another. In this case, you'd never get dupes...'cause the jsDoc is torn down before the overload emitting happens later. Will fix.

src/tsickle.ts
hess-g: You wrote it. :-) I was thinking the same thing, but there a bunch of things like this all over tsickle.ts. Like, checking (x) and (x.length) then doing for(...) when x is the result of a filter op on an array - filter always returns a non-null array, and has 0 length. So I was avoiding refactoring existing stuff that didn't match patterns elsewhere...but I will gladly do that here.

src/tsickle.ts
hess-g: Done.

src/tsickle.ts
hess-g: I don't think so. The first is within the for(..) over the different declarations, to build the set of unique list of return types. This is the final "do we emit a tag for @return or not'.

src/tsickle.ts
hess-g: oops. Missed adding the extra tags to the single ctor path. Done.

src/tsickle.ts
hess-g: Done

src/tsickle.ts
hess-g: Corrected your example, but I do like it. Done.
```
let lens = fnDecls.map(fnDecl => fnDecl.parameters.length);
```

src/tsickle.ts
hess-g: See the check below. The existing node is only added to `newDoc` if it's not already in the list. This _may_ lead to some oddities, if an annotation only applied to one overload and not another. However, that will be on the particular API's designer to deal with.

hess-g: In the existing golden tests for functions, for example, there are @ngInject. The bummer is that Array.indexOf doesn't do a good job with equality on the instanced of jsdoc.Tag. For now, I'm going to ignore these for overloads, and will put a TODO in to address that later.

src/tsickle.ts
hess-g: moar ES6, moar better. Change in init above, here, and in the join later (using Array.from).

src/tsickle.ts
```diff
@@ -105,33 +105,61 @@ class ClosureRewriter extends Rewriter {
...
-  emitFunctionType(fnDecl: ts.SignatureDeclaration, extraTags: jsdoc.Tag[] = []) {
+  /**
+   * Handles emittng the jsdoc for methods, including overloads.
+   * If overloaded, merges the signatures in the list of SignatureDeclarations into a single jsdoc.
+   * - Total number of parameters will be the maximum count found across all variants.
+   * - Any parameters beyond the minimum count found across all variants; prefix: "opt_"
+   * - Different names at the same parameter index will be joined with "_or_"
+   * - Variable args (...type[] in TypeScript) will be output as "...type",
+   *    except if found at the same index as another argument.
+   * @param {Array<ts.SignatureDeclaration>} fnDecls Pass > 1 declaration for overloads of same name
```
evmar: Tsickle itself actually rejects input code that specifies a type here.  :)
You can write it without the type, like
@param fnDecls Pass > 1 declaration...
same with @return.

(Also you don't need to document everything, just whatever you think isn't obvious.  So you can drop extraTags unless you want to write something about it.)

src/tsickle.ts
```diff
@@ -105,33 +105,61 @@ class ClosureRewriter extends Rewriter {
...
-    let sig = typeChecker.getSignatureFromDeclaration(fnDecl);
-
-    // Construct the JSDoc comment by reading the existing JSDoc, if
-    // any, and merging it with the known types of the function
-    // parameters and return type.
-    let jsDoc = this.getJSDoc(fnDecl) || [];
     let newDoc = extraTags;
+    let paramNamesToReturn: string[] = [];
+    const abstract = {tagName: 'abstract'};
+    let lens = fnDecls.map(fnDecl => fnDecl.parameters.length);
```
evmar: const, I guess.

src/tsickle.ts
```diff
@@ -105,33 +105,61 @@ class ClosureRewriter extends Rewriter {
...
+      if ((fnDecl.flags & ts.NodeFlags.Abstract) && (newDoc.indexOf(abstract) === -1)) {
+        newDoc.push(abstract);
+      }
+      // Merge the parameters into a single list of merged names and list of types
+      let sig = typeChecker.getSignatureFromDeclaration(fnDecl);
+      // Parameters.
       // Iterate through both the AST parameter list and the type's parameter
       // list, as some information is only available in the former.
-      for (let i = 0; i < sig.parameters.length; i++) {
+      for (let i = 0, l = sig.parameters.length; i < l; i++) {
```
evmar: I'm still confused here -- the diff shows you adding `l`, and the old code doesn't have it? (?)

hess-g: Habit - tracking the length of the iteration in a var on the declaration side so it's not evaluated on every iteration.

src/tsickle.ts
hess-g: Done

src/tsickle.ts
hess-g: Done

src/tsickle.ts
evmar: Oh, then this takes me back to my original comment -- we should use a profiler to guide performance rather than optimizing in anticipation.

hess-g: Changed it back. Again, just JS "best practice" habit. I agree that a compiler should absolutely optimize this scenario.

evmar:
(I'm ready to merge this, but I want to ensure the existing tsickle successfully syncs into g3 before I stack more on top of it.)

Issue #227, opened 2016-09-22
dep version update options (as of 22 Sep 2016)


hess-g:
Current output from [david](https://david-dm.org/angular/tsickle?view=list)

dependencies:
![image](https://cloud.githubusercontent.com/assets/12832244/18754762/ceabb1c2-80b7-11e6-940a-5face8cf1cdd.png)

devDependencies:
![image](https://cloud.githubusercontent.com/assets/12832244/18754767/d2dd633a-80b7-11e6-88a5-884f51dea08b.png)

hess-g:
All of the out of date deps listed above can be updated, with the exception of closure. Attempting to update to the latest (2016911.0.0) yields the following errors on `gulp test`:
```
test_files/jsdoc_types/jsdoc_types.js:29: ERROR - Bad type annotation. Unknown type module$contents$test_files$jsdoc_types$jsdoc_types_RenamedClassTwo
let /** @type {RenamedClassTwo} */ useLocalClassRenamedTwo = new module2_3.ClassTwo();
               ^
test_files/jsdoc_types/jsdoc_types.js:31: ERROR - Bad type annotation. Unknown type module$contents$test_files$jsdoc_types$jsdoc_types_Interface
let /** @type {Interface} */ useLocalInterface = null;
               ^
test_files/jsdoc_types/jsdoc_types.js:32: ERROR - Bad type annotation. Unknown type module$contents$test_files$jsdoc_types$jsdoc_types_ClassWithParams
let /** @type {ClassWithParams<number>} */ useClassWithParams = null;
               ^
3 error(s), 0 warning(s)
```

hess-g:
tslint is currently at a stable 3.15.1. On `npm i`, a warning about unmet peer dependencies is shown, since we are using a dev version of typescript. This is a warning only; tslint latest is installed and works just fine.

evmar:
Looks like everything's updated except for Closure.  I opened a new bug for that.

PR #228, opened 2016-09-23
#227 Update dependencies in package.json

hess-g:
Updates the following dependencies from/to:

dependencies
│ source-map  ^0.4.2 --> 0.5.6
│ source-map-support  ^0.3.1 --> 0.4.2

dev-dependencies
│ merge2 ^0.3.1 --> 1.0.2
│ gulp-typescript  ^2.7.6 --> 3.0.0
│ gulp-mocha  ^2.0.0 --> 3.0.1
│ tslint  3.15.0-dev.0 --> 3.15.1
│ chai  ^2.1.1 --> 3.5.0
│ @types/google-closure-compiler  0.0.17 --> 0.0.18

NOTE: The “UNMET PEER DEPENDENCY” for tslint > typescript is a warning
only; the peer dependency is not picked up since we are installing a
latest-dev version of typescript.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -43,7 +43,7 @@ var onError = function(err) {
 gulp.task('compile', function() {
   hasError = false;
   var tsResult =
-      gulp.src(['src/**/*.ts']).pipe(sourcemaps.init()).pipe(ts(tsProject)).on('error', onError);
+      gulp.src(['src/**/*.ts']).pipe(sourcemaps.init()).pipe(tsProject()).on('error', onError);
   return merge([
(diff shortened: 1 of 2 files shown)


evmar:
Normally I'd worry about using newer versions because it means all of our users must also use newer versions, but it seems this has no consequence on the tsickle code itself, right?

hess-g:
Well, the only non dep change was the change from the deprecated `pipe(ts(tsProject))` to `pipe(tsProject())`. The former is deprecated in the latest gulp-typescript. Your call on approving/merging this or killing this particular update.

evmar:
I manually restarted the travis build and it worked on this pass.

PR #229, opened 2016-09-23
raise the closure compilation timeout to 25 seconds

evmar:
The more tests we write, the slower this gets.


evmar:
@hess-g I think this is the problem that affected your last CL.

rkirov:
Lgtm

PR #232, opened 2016-09-27
avoid an error in ts 2.0

evmar:
This change was necessary for the TS 2.0 migration inside Google.
It's not clear to me why the code works on Travis though.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -53,7 +53,8 @@ class ClassRewriter extends Rewriter {
   }
 
   private decoratorsToLower(n: ts.Node): ts.Decorator[] {
-    return (n.decorators as ts.NodeArray<ts.Decorator>|| []).filter(d => this.shouldLower(d));
+    let decorators = (n.decorators || []) as ts.NodeArray<ts.Decorator>;
+    return decorators.filter(d => this.shouldLower(d));
   }
 
   /**
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -53,8 +53,7 @@ class ClassRewriter extends Rewriter {
   }
 
   private decoratorsToLower(n: ts.Node): ts.Decorator[] {
-    let decorators = (n.decorators || []) as ts.NodeArray<ts.Decorator>;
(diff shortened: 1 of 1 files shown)


src/decorator-annotator.ts
```diff
@@ -53,7 +53,8 @@ class ClassRewriter extends Rewriter {
   }
 
   private decoratorsToLower(n: ts.Node): ts.Decorator[] {
-    return (n.decorators as ts.NodeArray<ts.Decorator>|| []).filter(d => this.shouldLower(d));
+    let decorators = (n.decorators || []) as ts.NodeArray<ts.Decorator>;
```
mprobst: put the type annotation on the LHS?

rkirov:
I have no recollection of why we needed this change, so maybe we don't any longer. Let me try reverting it internally.

src/decorator-annotator.ts
rkirov: Won't work because actual type of `n.decorators || []` is `NodeArray<ts.Decorator> | never[]`, i.e. we are actually performing a downcast here not upcast.

IMHO, the most readable pattern and one that works with and without SNC would be:
```ts
  if (n.decorators) {
    return n.decorators.filter(() => true);
  }
  return [];
```

Also, on to check this as is to match the change in g3 and we can change it later.

rkirov:
I tested internally, and now understand why the change is needed and why travis passes. The major difference in setup is that we run this code without SNC (strictNullChecks) internally, and with SNC on Travis.

It is surprising to have code that fails without SNC but passes with SNC, since SNC are supposed to be stricter. However, consider `let x: C[]; x || []`, with SNC the compiler can reason that x is never falsy, thus what happens on the right hand side of `||` does not affect the type. However, without SNC, x can be null or undefined thus the expression `x || []` is an union type of whats on the left and on the right. Now add into the equation an extension of `Array<T>` like `NodeArray<T>` and you arrive at the issue we are seeing here.

evmar:
PTAL

evmar:
3rd try

rkirov:
LGTM

PR #233, opened 2016-10-03
feat: generate source mappings during rewriting.

mprobst:
This generates basic mappings for the intact pieces of the
original source code, which should be good enough for stack
traces (after combining with TypeScript's source map).


mprobst:
@evmar fixed, though I like the idea of "nam" and maybe also "nom" modules.

PR #237, opened 2016-10-05
add OSPO-required license headers to each src file

appsforartists:
They should also be in test_files, but we'd have to get the formatting correct to avoid breaking the tests.


appsforartists:
Good catch.  I'll fix and resubmit.

Issue #189, opened 2016-08-04
Turn off typechecking of tsickle output


evmar:
Currently tsickle emits TS code that typechecks.  This is nice because the TS compiler helps us spot our mistakes.

But it has some negative consequences:
- Because tsickle's input must be typechecked, using tsickle forces your builds to typecheck your app twice.
- Some Closure patterns are very difficult to express in legal TS.
  - We use a pretty weird hack for fields (the annotations helper static method) because we can't generate idiomatic Closure constructors.
  - When tsickle introduces a new value to capture a TS type, it's hard to make use of this value across modules.

As far as I understand TypeScript, as long as the code is syntactically valid, TS will generate *something* even if it's type-invalid.  And I think that something is predictable (it's a syntactic transformation of the input).

alexeagle:
Another option is to create a `ts.Program` with most tsickle changes included, type-check that, then make the remaining changes and emit without type-checking again. Just mentioning it to be comprehensive.

rkirov:
In an ideal world, tscikle modifications would happen as part of the emit pipeline (that doesn't exist yet), and will not be typechecked either, so I think this is reasonable as it aligns with long term plans.

rkirov:
In a post typechecking scenario, can we still write tests that assert the type errors happen at particular places but no other, so that we don't accidentally have the emit do something crazy and closure just accepting it and breaking code (as we have seen it do).

evmar:
I will leave this open to implement a variant of Rado's idea.

evmar:
I think what we can do is include the error messages (the errors we ignore, I mean) in a golden file somewhere.

PR #238, opened 2016-10-05
fix import error when running tsickle on valid tsc code

appsforartists:
tsickle's current approach is to run TypeScript, modify the output, and pass it through TypeScript again.  This can cause TypeScript to think imports have been declared twice (and fail accordingly).

As @evmar discovered with the internal build of tsickle (b/30708240), this can be worked around by ensuring `getDeclarationDiagnostics` isn't called on the second pass.  A more permanent solution would eliminate the second pass entirely, solving #189.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -208,18 +208,14 @@ function toClosureJS(
   // place of the original source.
   let host = createSourceReplacingCompilerHost(tsickleOutput, ts.createCompilerHost(options));
   program = ts.createProgram(fileNames, options, host);
-  let diagnostics = ts.getPreEmitDiagnostics(program);
-  if (diagnostics.length > 0) {
-    allDiagnostics.push(...diagnostics);
-    return null;
-  }
 
   // Emit, creating a map of fileName => generated JS source.
   const jsFiles = new Map<string, string>();
   function writeFile(fileName: string, data: string): void {
     jsFiles.set(fileName, data);
   }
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -208,7 +208,22 @@ function toClosureJS(
...
+  // through.  Until that issue is resolved (#189), we can work
+  // around it by inlining getPreEmitDiagnostics, and removing the
+  // offending line.
+
+  let diagnostics = [
+    ...program.getOptionsDiagnostics(),
+    ...program.getSyntacticDiagnostics(),
+    ...program.getGlobalDiagnostics(),
+  ];
   if (diagnostics.length > 0) {
```
evmar: Rather than this, we should delete this diagnostics check entirely -- we don't care about any errors that are produced here.

appsforartists: Just this one?  Should the other calls to `allDianostics.push` in toClosureJS be preserved?

evmar: Just this one, your new change looks right.

evmar:
Can you change the description to:

---

tsickle's current approach is to run TypeScript, modify the output, and pass it through TypeScript again. We don't want to type check the code twice, so only check it on the first pass.

Fixes #189.

appsforartists:
Done.

evmar:
Thanks!

PR #246, opened 2016-10-11
Have tsickle output a goog.require for each time a module is referenced.

MatrixFrog:
So the TypeScript code

  import {x} from './y';
  export {x} from './y';
  use(x);

becomes

  var y_1 = goog.require('y');
  var y_2 = goog.require('y');
  exports.x = y_2.x;

  use(y_1.x);

instead of

  var y_1 = goog.require('y');
  var y_2 = y_1;
  exports.x = y_2.x

  use(y_1.x);

This avoids bugs in the Closure Compiler's inlining logic, so that CollapseProperties can be enabled.


alexeagle:
Thank you Tyler!!

PR #247, opened 2016-10-11
add yarn lock file

appsforartists:
Trying out [yarn](https://code.facebook.com/posts/1840075619545360) an npm client that ensures installations are consistently reproducible across machines.  It generates this cache file to confirm that the correct versions of dependencies are downloaded (since `^` versions in dependencies often mean you'll have a different dependency tree depending on when you install).


mprobst:
👍 for using yarn.

PR #253, opened 2016-11-01
Add @unrestricted to all generated Closure classes.

mprobst:
Closure compiler will in some mode complain about properties being defined outside of the constructor. @unrestricted makes sure it does not complain about adding properties outside of the constructor. This would still be an attractive check (adding properties after construction can cause performance issues), but this is better suited to a lint check or similar in tslint.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -654,6 +654,10 @@ class Annotator extends ClosureRewriter {
 
   private visitClassDeclaration(classDecl: ts.ClassDeclaration) {
     let jsDoc = this.getJSDoc(classDecl) || [];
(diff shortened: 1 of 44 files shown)


src/tsickle.ts
```diff
@@ -654,6 +654,9 @@ class Annotator extends ClosureRewriter {
 
   private visitClassDeclaration(classDecl: ts.ClassDeclaration) {
     let jsDoc = this.getJSDoc(classDecl) || [];
+    // Always include @unrestricted so that Closure compiler does not complain if some class fields
+    // are not assigned in the constructor.
```
rkirov: Add that every tsickle output class does that because of '_tsickle_typeAnnotationsHelper'.

PR #254, opened 2016-11-03
Support decorators (not just annotations)

mprobst:
This series of commits was motivated by supporting decorators. It turns out that I had to shave various yaks (most of them self-inflicted) in the process:
* revert the original attempt at fixing
* upgrade Closure Compiler to repro the problem
* change tests not to alias module imports (no longer supported by Closure Compiler?)
* correctly emit for construct type literals such as the value/interface objects (`Document`) (now enforced by Closure Compiler)

Sorry for the great hairy snowball of changes :-(


test_files/jsdoc_types/jsdoc_types.ts
```diff
@@ -4,11 +4,7 @@
  */
 
 import * as module1 from './module1';
-import {ClassOne, value} from './module2';
```
rkirov: Does that mean closure disallows multiple imports from the same goog.module?

mprobst: Yes. We had to fix a bunch of internal uses of that already, we just didn't upgrade CC for outside use here.

src/tsickle.ts
```diff
@@ -744,12 +740,11 @@ class Annotator extends ClosureRewriter {
     if (!classDecl.name) return;
     let className = getIdentifierText(classDecl.name);
 
-    this.emit('\n\n  static _tsickle_typeAnnotationsHelper() {\n');
+    this.emit('\n\n// tsickle type annotations\n');
```

src/tsickle.ts
```diff
@@ -744,12 +740,11 @@ class Annotator extends ClosureRewriter {
     if (!classDecl.name) return;
     let className = getIdentifierText(classDecl.name);
 
-    this.emit('\n\n  static _tsickle_typeAnnotationsHelper() {\n');
```
mprobst: I think we discussed it at the time. I vaguely think we considered it, but then this worked and seemed closer to what ought to happen.

mprobst:
> LGTM, but I didn't see which changes in tsickle.ts enabled properly handling decorators. I see:
> 
> better emitting of field type declarations
> handling constructor signatures
> Was one of those breaking decorators?

Decorators are breaking with Closure Compiler because if we declare fields within a class, TypeScript renames the class reference to e.g. `DeclaredClass_1`. That's fixed by the "move out of class" change. Now with that, decorators still break in this repo due to the old Closure Compiler version, so I ugpraded. But with that, we need to handle the constructor signatures because Closure Compiler now checks something there (o_O). It's a long story full of yak hair :-)

rkirov:
I see, so decorators are still broken for classes with static methods that reference the class name. Of course that is a super rare case.

evmar:
By the way, the reason the properties were in the weird static function on the class was because we previously needed the generated code to pass the TypeScript type check.  Now that we no longer type check the tsickle output, your approach for using `Foo.prototype.bar = ...` is a much better solution.

PR #256, opened 2016-11-10
Do not double-emit JSDoc on classes.

mprobst:


src/tsickle.ts
```diff
@@ -657,8 +657,12 @@ class Annotator extends ClosureRewriter {
     if ((classDecl.flags & ts.NodeFlags.Abstract) !== 0) {
       jsDoc.push({tagName: 'abstract'});
     }
-    this.writeRange(classDecl.getFullStart(), classDecl.getStart());
-    this.emit(jsdoc.toString(jsDoc));
+    if (jsDoc.length === 0) {
+      this.writeRange(classDecl.getFullStart(), classDecl.getStart());
+    } else {
+      this.emit('\n');
```
rkirov: add a comment that a newline somehow prevents double emit

mprobst: The newline doesn't prevent double emit, the if statement does. The newline is just to make the output a bit nicer (not have it all on one line).

PR #257, opened 2016-11-10
Hide property declaring code in a function.

mprobst:
It turns out that a statement like "Foo.prototype.bar" might actually execute code if "bar" on Foo.prototype is a getter. That code then fails because its this pointer is not set in the invocation, breaking load of tsickle coverted code. Some users have an odd build setup where they execute tsickle-converted code in the browser, so we need to hide these statements in a function to avoid breaking code on load.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -740,11 +740,12 @@ class Annotator extends ClosureRewriter {
     if (!classDecl.name) return;
     let className = getIdentifierText(classDecl.name);
 
-    this.emit('\n\n// tsickle -> Closure type declarations\n');
+    this.emit(`\n\nfunction ${className}_tsickle_Closure_declarations() {\n`);
     staticProps.forEach(p => this.visitProperty([className], p));
     let memberNamespace = [className, 'prototype'];
     nonStaticProps.forEach((p) => this.visitProperty(memberNamespace, p));
     paramProps.forEach((p) => this.visitProperty(memberNamespace, p));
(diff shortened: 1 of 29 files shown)


src/tsickle.ts
```diff
@@ -740,11 +740,12 @@ class Annotator extends ClosureRewriter {
     if (!classDecl.name) return;
     let className = getIdentifierText(classDecl.name);
 
-    this.emit('\n\n// tsickle -> Closure type declarations\n');
+    this.emit(`\n\nfunction ${className}_tsickle_Closure_declarations() {\n`);
```
vikerman: nit: Why isn't it ${className}_tsickle_closure_declarations()

src/tsickle.ts
```diff
@@ -740,11 +740,12 @@ class Annotator extends ClosureRewriter {
     if (!classDecl.name) return;
     let className = getIdentifierText(classDecl.name);
 
-    this.emit('\n\n// tsickle -> Closure type declarations\n');
```
rkirov: can you keep the comment? Or you think it adds nothing?

rkirov:
add "@unrestricted" back

src/tsickle.ts
mprobst: Don't think it adds something, we have the method name that says the same thing.

mprobst:
As discussed offline, `@unrestricted` was just a red herring I was trying to follow earlier.

PR #258, opened 2016-11-10
Assign a default export in goog.module()s

mprobst:
Closure Compiler works on improving strictness for goog.module(), in particular enforcing that named exports are immutable. The exports = {} statement indicates to Closure Compiler that there is a default export, so Closure Compiler does not error on mutating/accessing named exports on it.

This is more of a workaround than a fix, but TypeScript users are guaranteed to only ever access named exports through indirection off of the module export object (that is, TypeScript compiler makes sure to implement mutable exports on the usage site), so for TS users the error that Closure Compiler tries to prevent here does not actually occur.

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -54,6 +54,9 @@ class ES5Processor extends Rewriter {
     // so that @suppress statements work.
     const moduleName = this.pathToModuleName('', this.file.fileName);
     // NB: No linebreak after module call so sourcemaps are not offset.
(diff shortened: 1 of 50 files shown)


blickly:
Great, thanks Martin!

blickly:
Thanks Martin. Can you let me know when this change lands internally?

mprobst:
I'm working on syncing it right now, running into some incompatibility
between goog.module and NodeJS module loading. Hopefully will have it in
early next week.
-- 

Martin

PR #259, opened 2016-11-11
Mark TypeScript types as non-nullable

mprobst:
Required some more yaks to loose their fur.

Note that the support for NonNullable expressions is very flakey, due to https://github.com/Microsoft/TypeScript/issues/9873, just as it was for casts before.


test_files/type/type.tsickle.ts
```diff
@@ -9,18 +9,18 @@ interface Array<T> {
...
-let /** @type {Object} */ typeObjectEmpty: {} = null;
+let /** @type {!Object<string,number>} */ typeObject2: {[key:string]: number};
+let /** @type {?} */ typeObject3: {a:number, [key:string]: number};
+let /** @type {!Object} */ typeObjectEmpty: {};
 
 let /** @type {(string|boolean)} */ typeUnion: string|boolean = Math.random() > 0.5 ? false : '';
 let /** @type {(string|boolean)} */ typeUnion2: (string|boolean) = Math.random() > 0.5 ? false : '';
-let /** @type {{optional: (boolean|undefined)}} */ typeOptionalField: {optional?: boolean} = {};
-let /** @type {{optional: ((string|boolean)|undefined)}} */ typeOptionalUnion: {optional?: string|boolean} = {};
+let /** @type {{optional: ((undefined|boolean)|undefined)}} */ typeOptionalField: {optional?: boolean} = {};
```
vikerman: Is there a way to collapse the two undefined?

PR #266, opened 2016-11-16
Make the tsickle ES6 emit CommonJS compatible by excluding the export…

mprobst:
…s object in the expected module.exports location.

This helps if infrastructure ends up loading ES6 code to run in CommonJS.


mprobst:
Actually, I was tracking a red herring here, and this is superseded by a fix to some other infrastructure to never have the ES6 code leak into ES5 mode.

PR #270, opened 2016-11-18
fix(jsdoc): include leading star on each JSDoc line

alexeagle:


alexeagle:
Note, it's a little gross.
Alternative is more complex: instead of jsdoc.parse returning a tag.text, it could be a tag.text[], so the line joining happens later, when we are emitting.

PR #278, opened 2016-11-23
strip @extends, @implements, and @constructor from user jsdoc

### LINKED ISSUE #269: Strip @extends and @constructor from classes
rxjs has code like

```js
/**
 * We need this JSDoc comment for affecting ESDoc.
 * @ignore
 * @extends {Ignored}
 */
export class InnerSubscriber<T, R> extends Subscriber<R> {
```

We need to remove the @extends to make Closure happy.

Also this:
```js
 /**
   * @constructor
   */
  constructor(subscribe?: <R>(subscriber: Subscriber<R>) => TeardownLogic) {
```

evmar:
RxJS uses these and we need to remove them to make Closure happy.

Fixes #269.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -677,12 +677,8 @@ class Annotator extends ClosureRewriter {
     if ((classDecl.flags & ts.NodeFlags.Abstract) !== 0) {
       jsDoc.push({tagName: 'abstract'});
     }
-    if (jsDoc.length === 0) {
-      this.writeRange(classDecl.getFullStart(), classDecl.getStart());
-    } else {
-      this.emit('\n');
-      this.emit(jsdoc.toString(jsDoc));
-    }
+    this.emit('\n');
+    if (jsDoc.length > 0) this.emit(jsdoc.toString(jsDoc));
     if (classDecl.members.length > 0) {
       // We must visit all members individually, to strip out any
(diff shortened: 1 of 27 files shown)


src/tsickle.ts
```diff
@@ -677,12 +677,8 @@ class Annotator extends ClosureRewriter {
     if ((classDecl.flags & ts.NodeFlags.Abstract) !== 0) {
       jsDoc.push({tagName: 'abstract'});
     }
-    if (jsDoc.length === 0) {
-      this.writeRange(classDecl.getFullStart(), classDecl.getStart());
```
mprobst: This was kind of nice to include any comments before the JSDoc. But I guess we don't really care?

test_files/jsdoc/jsdoc.js
```diff
@@ -41,3 +41,11 @@ function JSDocTest_tsickle_Closure_declarations() {
  */
 function x() { }
 ;
+/**
+ *  This class has JSDoc, but some of it should be stripped.
```
mprobst: I always read these first and am confused ;-)

src/tsickle.ts
evmar: We've been gradually going towards "generated files are an internal compiler detail and not human-friendly".  We still preserve any JSDoc though.

Issue #288, opened 2016-12-01
tsickle output preserves shorthand imports


alexeagle:
commonJS allows 
```
folder/
  index.ts
```
to be imported with `import {} from 'folder';`
TypeScript allows the same convention, and in its `--module=es6` output it preserves that import location.

However, this is unspecified and not supported by all bundlers, in particular closure compiler can't accept this ES6 output.

Tsickle is a good place for us to rewrite such imports to come from the canonical location.

alexeagle:
Fixed by that PR.

PR #289, opened 2016-12-01
WIP: re-write shorthand index imports

alexeagle:
addresses https://github.com/angular/tsickle/issues/288


alexeagle:
this is probably not the right way - we should just do this in the existing pass where we annotate the code.
But, our existing test harness doesn't accept multiple files AFAICT
And I kind of like this new testing harness.

evmar:
The existing harness runs all files in a single test_files/foo/ directory together.

evmar:
The reason I like the existing harness is that (1) it tests interactions between these various rewriters, and (2) whenever we change the goldens in these code-based golden tests we need to copy and paste file contents around into tests.

evmar:
I think this rewriting stuff belongs inside processES5.  That code already does a bunch of random unrelated rewrites in a single pass, so this would fit in there fine I think.

alexeagle:
Thanks Evan, I'll look in a few days when I have some spare cycles again.

alexeagle:
PTAL.

ES5Processor isn't the right place for this because that's a goog.module translator, and runs as a post-process.
We really want to pre-process the sources, down-leveling TypeScript module semantics (allow import from index) to ES6 Module semantics, which don't dictate such a thing. So I think this needs to go in the main tsickle rewriter.

The extra module resolution is known to take extra time. Maybe I should hide this behind an option (plus I think we decided not to encourage the 'index.ts' pattern in google3)

src/tsickle.ts
```diff
@@ -630,6 +638,21 @@ class Annotator extends ClosureRewriter {
     return toArray(reexports.keys());
   }
 
+  private writeModuleSpecifier(moduleSpecifier: ts.Expression) {
+    let moduleId = (moduleSpecifier as any).text;
```
evmar: I think you probably need to do some sort of switch on moduleSpecifier.kind here, to make sure we produce a good error message in the case where it's something we don't expect.  See other calls to `this.errorUnimplementedKind`.

src/tsickle.ts
```diff
@@ -630,6 +638,21 @@ class Annotator extends ClosureRewriter {
...
+    if (this.tsOpts && this.host) {
+      const resolved = ts.resolveModuleName(moduleId, this.file.fileName, this.tsOpts, this.host);
```
evmar: Can you throw in a comment in this block about what you're doing (something about "rewrite index references to contain /index" or whatever)

src/tsickle.ts
```diff
@@ -630,6 +638,21 @@ class Annotator extends ClosureRewriter {
...
+      if (resolved && resolved.resolvedModule) {
+        const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(/(\.d)?\.ts$/, '');
+
+        if (resolvedModule.match(/\/index$/) && !moduleId.match(/\/index$/)) {
+          moduleId += '/index';
```
evmar: Does this do the right thing if the person writes
`import * from 'foobar.js'`
?  I can't remember if they are trying to get people to do that now... 

...and do we even care?  Might be fine to just not worry about it for now.

src/tsickle.ts
evmar: https://github.com/angular/tsickle/blob/d3bb6897edddee5df0dcb5b38fbb57b77393d436/src/tsickle.ts#L99 is an example

src/tsickle.ts
alexeagle: good point, I need a test for that, Mohamed says that is a workaround:
https://github.com/Microsoft/TypeScript/issues/12597#issuecomment-264254009

alexeagle:
I don't think we should run this new codepath in google3, because resolving the modules will make annotation slower, and we decided not to provide better support for index imports in google3.
So I added a `convertIndexImportShorthand?: boolean` to the tsickle Options.

src/tsickle.ts
```diff
@@ -592,18 +605,56 @@ class Annotator extends ClosureRewriter {
   }
 
   /**
+   * workaround for https://github.com/Microsoft/TypeScript/issues/12597
```
evmar: Can you make this first say what it does, then mention that it's a workaround?

src/tsickle.ts
```diff
@@ -592,18 +605,56 @@ class Annotator extends ClosureRewriter {
...
+   */
+  private writeModuleSpecifier(moduleSpecifier: ts.Expression) {
+    function getText() {
+      switch (moduleSpecifier.kind) {
+        case ts.SyntaxKind.StringLiteral:
+          return (moduleSpecifier as ts.StringLiteral).text;
+        default:
+          // The above list of kinds should be exhaustive.
+          throw new Error(`unhandled moduleSpecifier kind: ${ts.SyntaxKind[moduleSpecifier.kind]}`);
```
evmar: This could be simplified to something like:

```
if (moduleSpecifier.kind !== ts.SyntaxKind.StringLiteral) throw new Error ...
let moduleId = (moduleSpecifier as ts.SyntaxKind.StringLiteral).text

src/tsickle.ts
alexeagle: done

src/tsickle.ts
alexeagle: done

PR #290, opened 2016-12-01
feat: support including a prelude in processed files.

mprobst:
This allows to e.g. insert an explicit import for the TypeScript helpers library.

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -44,8 +44,8 @@ class ES5Processor extends Rewriter {
   unusedIndex: number = 0;
 
   constructor(
-      file: ts.SourceFile,
-      private pathToModuleName: (context: string, fileName: string) => string) {
+      file: ts.SourceFile, private pathToModuleName: (context: string, fileName: string) => string,
+      private prelude: string) {
     super(file);
   }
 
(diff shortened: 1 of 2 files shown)


src/es5processor.ts
```diff
@@ -44,8 +44,8 @@ class ES5Processor extends Rewriter {
   unusedIndex: number = 0;
 
   constructor(
-      file: ts.SourceFile,
-      private pathToModuleName: (context: string, fileName: string) => string) {
+      file: ts.SourceFile, private pathToModuleName: (context: string, fileName: string) => string,
```
evmar: Wish it would just put each param on its own line :~(

src/es5processor.ts
```diff
@@ -317,11 +318,13 @@ class ES5Processor extends Rewriter {
...
  *     imports with relative paths like "import * as foo from '../foo';".
+ * @param prelude An additional prelude to insert after the `goog.module` call,
+ *     e.g. with additional imports or requires.
  */
 export function processES5(
     fileName: string, moduleId: string, content: string,
-    pathToModuleName: (context: string, fileName: string) => string,
-    isES5 = true): {output: string, referencedModules: string[]} {
+    pathToModuleName: (context: string, fileName: string) => string, isES5 = true,
+    prelude = ''): {output: string, referencedModules: string[]} {
```
evmar: This is getting dangerously close to needing to accept an "options" object with named fields.

PR #292, opened 2016-12-05
obey overloads in extern generation

### LINKED ISSUE #291: Handle function overloads everywhere
Our overload logic is currently only in place on classes, where we first collect all the functions and then unify overloads.

We need to also do the same unification for top-level function declarations and declarations in namespaces, etc.  Basically look at all instances where we wrote `emitFunctionType([f])` instead of passing through an array of functions.

evmar:
When emitting a function in externs, find all overloads of the
function and pass the full set to the function type emitter so
that overloads are properly unioned.

Fixes #291.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -90,19 +90,6 @@ function isValidClosurePropertyName(name: string): boolean {
   return /^[a-zA-Z_][a-zA-Z0-9_]*$/.test(name);
 }
 
-/**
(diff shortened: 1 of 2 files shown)


evmar:
(Post-review commits: I attempted to delete a function in the original CL but accidentally only deleted the JSDoc.  Deleting the function revealed one more caller that I deleted.)

PR #293, opened 2016-12-05
rearrange overload definition merging

evmar:
When merging multiple function overloads, we previously attempted
to collect up multiple parallel arrays of param names and types.
This approach dropped other metadata like rest params and optional
attributes.  It also had separate codepaths for overloaded functions
and non-overloaded functions.

Instead, merge overloaded functions by gathering up sets of JSDoc,
and then merge the JSDoc.  This brings the non-overload and overload
cases together, simplifies the code, and makes optional params
correctly gain the "=" suffix.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -132,7 +132,6 @@ class ClosureRewriter extends Rewriter {
    * Handles emittng the jsdoc for methods, including overloads.
    * If overloaded, merges the signatures in the list of SignatureDeclarations into a single jsdoc.
    * - Total number of parameters will be the maximum count found across all variants.
-   * - Any parameters beyond the minimum count found across all variants; prefix: "opt_"
    * - Different names at the same parameter index will be joined with "_or_"
    * - Variable args (...type[] in TypeScript) will be output as "...type",
    *    except if found at the same index as another argument.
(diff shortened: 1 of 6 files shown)


evmar:
CC @hess-g

src/tsickle.ts
```diff
@@ -163,19 +162,19 @@ class ClosureRewriter extends Rewriter {
       let jsDoc = this.getJSDoc(fnDecl) || [];
 
       // Copy all the tags other than @param/@return into the new
-      // comment without any change; @param/@return are handled later.
-      // TODO: handle for overloads. indexOf on the array doesn't match instances of the Tag type,
-      // so another matching strategy is needed to avoid dupes. Also, there may be problems if
-      // an annotation doesn't apply to all overloads.
+      // JSDoc without any change; @param/@return are handled specially.
+      // TODO: there may be problems if an annotation doesn't apply to all overloads;
```
mprobst: There's also @template T. But I don't know what we'd do in that case, that seems like it just won't fly?

src/tsickle.ts
```diff
@@ -203,87 +202,62 @@ class ClosureRewriter extends Rewriter {
...
+    // Merge the JSDoc tags for each overloaded parameter.
+    for (let i = 0; i < maxArgsCount; i++) {
+      let paramTag = jsdoc.merge(paramTags[i]);
+      // If any overload marks this param optional, mark it optional in the
+      // merged output.
+      const optional = paramTags[i].find(t => t.optional === true) !== undefined;
+      if (optional || i >= minArgsCount) {
+        // TODO(evanm): this successfully renames the parameter in the JSDoc,
+        // but we also need to rename the parameter in the emitted function.
+        // paramTag.parameterName = 'opt_' + paramTag.parameterName;
```
mprobst: I think `opt_` is just a coding convention, not load bearing (and also no longer recommended by style guide). the `=` is the core part. So unless you care about the style, you don't need that here.

evmar: OK, I just removed references to opt_.  I also confirmed they aren't load-bearing in the jscompiler debugger.

PR #296, opened 2016-12-06
fix(test_files): remove committed warnings and error messages

bitjson:
Not sure if these lines were intentionally committed – they seem to be causing Typescript errors in downstream projects that aren't ignoring node_modules. This PR fixes https://github.com/driftyco/ionic-app-scripts/issues/468 for me.

### DIFF
--- a/test_files/declare/declare.tsickle.d.ts
+++ b/test_files/declare/declare.tsickle.d.ts
@@ -1,5 +1,3 @@
-Warning at test_files/declare/declare.d.ts:81:1: anonymous type has no symbol
-====
 declare namespace DeclareTestModule {
   namespace inner {
     var someBool: boolean;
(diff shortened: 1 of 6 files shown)


bitjson:
Ah –

evmar:
I'm not great with npm.  Is it normal to include or exclude the test suite when distributing via npm?

(Note that you cannot in general include node_modules in a single TS project, because each separate module might be built with different incompatible TS flags.)

bitjson:
@evmar – typically npm packages only contain the absolute minimum needed to use the package. (I just looked around at some of the more popular ones to confirm.) I just opened another PR here: #297

(And thanks for the note – I'd not realized that, I'm pretty new to Typescript. Somehow this was the only module causing issues, but it sounds like I need to dig into why my project was also including node_modules.)

PR #297, opened 2016-12-06
build(npmignore): remove tests and vscode config from npm package

bitjson:
This will ignore both `test` and `build/test`, `.vscode`, and `test_files`

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,4 +2,6 @@
 .gitignore
 .idea
 .travis.yml
-build/test
+.vscode
+test
+test_files


evmar:
@alexeagle @mprobst is this the right thing to do?

mprobst:
Yes, this is the right thing to do.

I think we should probably really only include `build/src` (and maybe merge `build/definitions` into it), we're still over-including so to say, but this is a correct step.

That beign said @bitjson, if your project gives you errors from files in `node_modules`, something in your setup is wrong. The TypeScript compiler should not find these files at all for your project. Maybe you have an incorrect `exclude` section in your `tsconfig.json`? You should always exclude `node_modules`.

bitjson:
Thanks for the note @mprobst – I think I have that solved now in my project too.

Issue #263, opened 2016-11-15
Work around destructuring types


evmar:
This code:
```ts
function get({ bar = 0 }) { }
get({});
```

may produce an error like:
```
WARNING - Property bar never defined on $jscomp$destructuring$var1
```

if the name `bar` is never used elsewhere.
We can work around this either by providing a type annotation that mentions `bar` or perhaps by some throwaway code like
```
namespace __unused { var bar; }
```

evmar:
This output has no warning:
```js
/**
 * @param {{bar: (number|undefined)}} __0
 */
function get({ bar = 0 } = {}) { }
get({});
```

mprobst:
`__0` is the parameter index, I presume?

evmar:
I've already forgotten how this works but my recollection is that currently there's no way to name the destructuring parameter, so tsickle just invents the names as "__0", "__1" etc.  I'm not sure how Closure eventually matches them up, though I imagine it probably just matches the indexes in the @params array with the function args.  (Another way of saying this is that I think Closure ignores the variable names in the jsdoc for destructuring params.)

concavelenz:
Thats right, it is ignored by the compiler, the name is only used to complete the JSDoc "form" and to name the parameter for any generated documentation.

mprobst:
@evmar is this done?

evmar:
No.

Given the input:
```js
  declare function destructures({a: number});
```
We were generating:
```js
/** ...closure stuff */
function destructures({a: number});
```
which causes a Closure crash, but now we generate
```js
/** ...closure stuff */
function destructures(__0);
```
which is still wrong (need to put "a" in the jsdoc) but at least avoids the crash.

PR #302, opened 2016-12-09
always enable typed mode when destructuring

evmar:
Even in untyped mode, it's important we provide a type signature
for destructured types, because of the case like:
  function foo({bar}: {bar:number})
we need "bar" to appear in the JSDoc.

Rather than special-case destructuring in untyped mode, let's try
just always providing full types for destructured cases; we'd like
to eliminate untyped mode anyway and hopefully there aren't enough
cases of destructuring where typed mode will fall down.

Fixes issue #263.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -304,7 +304,15 @@ class ClosureRewriter extends Rewriter {
    *     bind patterns.
    */
   typeToClosure(context: ts.Node, type?: ts.Type, destructuring = false): string {
-    if (this.options.untyped) {
+    if (this.options.untyped && !destructuring) {
+      // Note: even in untyped mode, it's important we provide a type signature
+      // for destructured types, because of the case like:
+      //   function foo({bar}: {bar:number})
+      // we need "bar" to appear in the JSDoc.
+      // Rather than special-case destructuring in untyped mode, the above logic
+      // just always provides full types for destructured cases; we'd like
+      // to eliminate untyped mode anyway and hopefully there aren't enough
+      // cases of destructuring where typed mode will fall down.
       return '?';
     }
 
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -270,6 +270,9 @@ export class TypeTranslator {
         typeStr += this.translate(referenceType.target);
       }
       if (referenceType.typeArguments) {
+        // Closure doesn't accept a type like {?<...>}, so if the parameterized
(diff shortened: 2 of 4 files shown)


src/type-translator.ts
```diff
@@ -270,6 +270,9 @@ export class TypeTranslator {
         typeStr += this.translate(referenceType.target);
       }
       if (referenceType.typeArguments) {
+        // Closure doesn't accept a type like {?<...>}, so if the parameterized
+        // type itself is unknown just use that by itself.
```
evmar: This came up in an obscure Angular2 case that I wasn't able to reproduce in a test.  Something about a type-parameterized interface used in destructuring where the interface itself failed to convert to a Closure type...

src/tsickle.ts
```diff
@@ -304,7 +304,15 @@ class ClosureRewriter extends Rewriter {
    *     bind patterns.
    */
   typeToClosure(context: ts.Node, type?: ts.Type, destructuring = false): string {
-    if (this.options.untyped) {
+    if (this.options.untyped && !destructuring) {
+      // Note: even in untyped mode, it's important we provide a type signature
+      // for destructured types, because of the case like:
+      //   function foo({bar}: {bar:number})
```
mprobst: What if we have:

```js
function foo({bar}: {bar: SomeTypeFromTs}) {}
```

Will Closure see `SomeTypeFromTs` and error out because that has never been emitted?

src/type-translator.ts
```diff
@@ -270,6 +270,9 @@ export class TypeTranslator {
         typeStr += this.translate(referenceType.target);
       }
       if (referenceType.typeArguments) {
+        // Closure doesn't accept a type like {?<...>}, so if the parameterized
```
mprobst: I don't understand `type like {?<...>}`

evmar: This code is generating the output for parameterized types e.g. `FooType<string, number>`.  If the `FooType` in particular is `?`, we would generate something like `?<string, number>` which is illegal in Closure.

I think this is a broader tsickle bug in typed mode that this change reveals because it exposes more tsickle logic in untyped mode.

Do you have any better suggestion on how to word this?  I agree it kinda sucks right now.

mprobst:
Ping?

evmar:
I think the last state of this change is that it brings in enough of typed mode into g3 that it doesn't actually work.

PR #304, opened 2016-12-10
properly handle reopened interfaces

evmar:
Interfaces can be reopened, but we should only emit the interface
(and each field) once.

TODO(me): perhaps we should emit the *type* of the interface once
and let TS do the unification for us.

### DIFF
--- a/test_files/interface/interface.js
+++ b/test_files/interface/interface.js
@@ -27,3 +27,9 @@ TrickyInterface.prototype.foo;
 */
 /** @type {(undefined|string)} */
 TrickyInterface.prototype.foobar;
+/** @record */
+function Reopened() { }
+/** @type {number} */
+Reopened.prototype.x;
+/** @type {string} */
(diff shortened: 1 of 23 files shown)


test_files/interface/interface.js
```diff
@@ -27,3 +27,9 @@ TrickyInterface.prototype.foo;
 */
 /** @type {(undefined|string)} */
 TrickyInterface.prototype.foobar;
+/** @record */
+function Reopened() { }
+/** @type {number} */
+Reopened.prototype.x;
+/** @type {string} */
+Reopened.prototype.y;
```
evmar: This is the test case.

mprobst:
Ping?

mprobst:
Ping ping?

evmar:
I wrote a TODO about how we ought to use TS to unify the interface decls into a single unified decl before emitting.  I now think that is right.  I will need to recreate this CL to do this right.

Issue #295, opened 2016-12-06
Migrate to TypeScript 2.1


evmar:
I started work on migrating to the TypeScript 2.1 API.

https://github.com/angular/tsickle/tree/ts2.1

evmar:
I think we'll need to branch tsickle for this; we can't support TS<2.1 and TS2.1 simultaneously.

fxck:
any progress on this?

PR #311, opened 2017-01-09
ts2.1: rearrange type translation, use new ObjectFlags

evmar:
Make the main "translate" function match the enum in the
TypeScript API.  Split out the handling of more complex
Object types into a separate function, and adjust the code
there to use the new objectFlags field.

This actually makes most tests pass(!) but there are still
a few more to track down, and there are still a bunch of
TODOs.

More work on #295.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -51,6 +51,27 @@ export function typeToDebugString(type: ts.Type): string {
     }
   }
 
+  if (type.flags === ts.TypeFlags.Object) {
+    const objType = type as ts.ObjectType;
+    const objectFlags: ts.ObjectFlags[] = [
+      ts.ObjectFlags.Class,
+      ts.ObjectFlags.Interface,
+      ts.ObjectFlags.Reference,
+      ts.ObjectFlags.Tuple,
+      ts.ObjectFlags.Anonymous,
+      ts.ObjectFlags.Mapped,
+      ts.ObjectFlags.Instantiated,
+      ts.ObjectFlags.ObjectLiteral,
+      ts.ObjectFlags.EvolvingArray,
+      ts.ObjectFlags.ObjectLiteralPatternWithComputedProperties,
+    ];
+    for (let flag of objectFlags) {
+      if ((objType.objectFlags & flag) !== 0) {
+        debugString += ` object:${ts.ObjectFlags[flag]}`;
+      }
+    }
+  }
+
   if (type.symbol && type.symbol.name !== '__type') {
     debugString += ` symbol.name:${JSON.stringify(type.symbol.name)}`;
   }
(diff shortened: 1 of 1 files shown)


evmar:
Note: it's still intentional that the tests don't pass.

src/type-translator.ts
```diff
@@ -164,51 +185,78 @@ export class TypeTranslator {
...
-      case ts.TypeFlags.Enum:
-        return 'number';
-      case ts.TypeFlags.StringLiteral:
-        return 'string';
-      case ts.TypeFlags.BooleanLiteral:
-        return 'boolean';
-      case ts.TypeFlags.NumberLiteral:
-        return 'number';
+      case ts.TypeFlags.Never:
+        this.warn(`should not emit a 'never' type`);
```
rkirov: why warning here? I think there are valid programs with the never type. Like users that mistrust the type system and want to augment it with a runtime check

```
switch (x)
 default:
  throw Error(x as any) <- x could be never here 
```

src/type-translator.ts
```diff
@@ -230,34 +278,27 @@ export class TypeTranslator {
...
       // A reference to another type, e.g. Array<number> refers to Array.
       // Emit the referenced type and any type arguments.
       let referenceType = type as ts.TypeReference;
 
       let typeStr = '';
-      // TODO(ts2.1): let isTuple = (referenceType.flags & ts.TypeFlags.Tuple) > 0;
-      let isTuple = true;
-      // For unknown reasons, tuple types can be reference types containing a
-      // reference loop. see Destructuring3 in functions.ts.
-      // TODO(rado): handle tuples in their own branch.
```
rkirov: thanks for taking care of my todo :)

src/type-translator.ts
evmar: The question is what we want to do when we hit one of these.  E.g. if you accidentally make an array of never, what type do we emit for Closure?

These warnings are never shown to a user and only show up inside the test suite, so it's probably harmless to leave it in... (?)

src/type-translator.ts
```diff
@@ -164,51 +185,78 @@ export class TypeTranslator {
...
+      case ts.TypeFlags.NumberLiteral:
         return 'number';
       case ts.TypeFlags.Boolean:
+      case ts.TypeFlags.BooleanLiteral:
         return 'boolean';
+      case ts.TypeFlags.Enum:
+      case ts.TypeFlags.EnumLiteral:
+        return 'number';
+      case ts.TypeFlags.ESSymbol:
+        // NOTE: currently this is just a typedef for {?}, shrug.
```
evmar: Done.

src/type-translator.ts
evmar: It's not done yet!  There's an empty spot where tuples need to be handled.  :P

src/type-translator.ts
rkirov: So it is for inexpressible concepts in Closure. Makes sense, maybe StringLiterals and etc *Literals should also get a warning then? No need to take care in the cl.

PR #312, opened 2017-01-09
Expose the source maps in the decorator-annotator code.

LucasSloan:
Rewriter generates source maps, but we weren't exposing that in DecoratorRewriter.  We need this exposed in order to support source maps going through decorator rewriting as well the type annotation commenting passes.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import {SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
 import {Rewriter} from './rewriter';
@@ -279,7 +280,7 @@ class DecoratorRewriter extends Rewriter {
     super(sourceFile);
   }
 
-  process(): {output: string, diagnostics: ts.Diagnostic[]} {
+  process(): {output: string, diagnostics: ts.Diagnostic[], sourceMap: SourceMapGenerator} {
     this.visit(this.file);
     return this.getOutput();
   }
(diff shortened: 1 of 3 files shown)


test/decorator-annotator_test.ts
```diff
@@ -54,6 +55,21 @@ describe(
...
       });
 
+      it('generates a source map', () => {
+        const {output, sourceMap} = translate(`
+/** @Annotation */ let Test1: Function;
+@Test1
+export class Foo {
+}
+let X = 'a string';`);
+        const rawMap = (sourceMap as any).toJSON();
```
evmar: Why is this coercion needed?

src/decorator-annotator.ts
```diff
@@ -279,7 +280,7 @@ class DecoratorRewriter extends Rewriter {
     super(sourceFile);
   }
 
-  process(): {output: string, diagnostics: ts.Diagnostic[]} {
+  process(): {output: string, diagnostics: ts.Diagnostic[], sourceMap: SourceMapGenerator} {
```
evmar: Why are these returning Generators rather than the SourceMaps themselves?  (Am I just misunderstanding the API?)

LucasSloan: I don't know why.  @mprobst wrote the original feature, let's ask him.

mprobst: I think I wanted users to be able to augment or combine the source map, without having to go through generator -> JSON -> source map again. At least that's what I remember :-)

If it works for you to return the `SourceMap`, go for it.

test/decorator-annotator_test.ts
LucasSloan: It doesn't appear to be.  I copied and pasted from the source map test, I fixed it there as well.

src/decorator-annotator.ts
LucasSloan: I think that for the purpose of munging source maps together, returning a SourceMapGenerator is slightly better than returning a SourceMapConsumer.  I think we should keep the existing API.

LucasSloan:
@evmar 

Responded to comments, PTAL?

PR #314, opened 2017-01-10
ts2.1: handle boolean types correctly

evmar:
Booleans are represented as a union of true|false, but they
also have the boolean flag set so our existing code that handled
unions wasn't firing.

If we just treat them as unions, everything works, including more
complex cases (like boolean|number => true|false|number, without
the boolean flag set).

More work on #295.  With this change, the remaining failures in
tests are mostly whitespace.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -185,8 +185,8 @@ export class TypeTranslator {
   }
 
   translate(type: ts.Type): string {
-    // NOTE: Though type.flags has the name "flags", it can only be one of
-    // the enum options at a time.  This switch handles all the cases in
+    // NOTE: Though type.flags has the name "flags", it usually can only be one
+    // of the enum options at a time.  This switch handles all the cases in
     // the ts.TypeFlags enum in the order they occur.
     switch (type.flags) {
       case ts.TypeFlags.Any:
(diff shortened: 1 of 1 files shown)


src/type-translator.ts
```diff
@@ -223,26 +224,41 @@ export class TypeTranslator {
...
       default:
+        // Handle cases where multiple flags are set.
+
+        // Booleans are represented as
+        //   ts.TypeFlags.Union | ts.TypeFlags.Boolean
+        // where the union is a union of true|false.
+        // Note also that in a more complex union, e.g. boolean|number, then
+        // it's a union of three things (true|false|number) and
+        // ts.TypeFlags.Boolean doesn't show up at all.
+        if (type.flags & ts.TypeFlags.Union) {
```
rkirov: I am confused here. You say that for booleans Union and Boolean flags are on. Since more than one flag is one we are in the 'default' case. Then we are in this case too, and we output 'true | false', but don't we want 'boolean' for booleans?

evmar: The union is true|false in TS, which becomes boolean|boolean in Closure, which we dedup to boolean.

The thing I was trying to point out in this comment is that by handling the union it works everywhere, so we don't need to special-case boolean.
E.g. if the input is boolean|number, then it becomes true|false|number, which we convert to Closure boolean|boolean|number and then dedup to boolean|number.

Do you have advice on how to make the comment better?

rkirov: That makes sense, I was expecting special casing boolean. My reasoning is that if closure ever adds 'true' and 'false', this code will emit 'true|false' for input 'boolean'.

But even with special case boolean, closure adding 'true' would still break on 'boolean | number' because it will emit 'true | false | number'. So I am not going to worry about it.

PR #315, opened 2017-01-10
Use a private TS api to get full comment text.

### LINKED ISSUE #313: getDocumentationComment no longer returns @Annotation for annotations
We rely on sym.getDocumentationComment() to get the JSDoc on decorators to determine that they are annotations.  In TS 2.1 it's now the case that getDocumentationComment() does not return the text of JSDoc tags.  So the @Annotation is dropped.

I traced the code and found inside the compiler:
```js
        return map(getJSDocs(node), doc => doc.comment);
```
where doc.comment is empty; we need doc.tags.

I looked at the API for a while and didn't see an easy way to get at the JSDoc we need.  We might need to change this code to use tsickle's own jsdoc parser.

Blocks #295.

rkirov:
In TS 2.1 the comment text is stripped of jsdoc @ annotations.
To read the full JSDoc text we temporary use a private API.

Passes all tests with 'TEST_FILTER=decorator gulp watch'

Closes #313

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -57,7 +57,19 @@ class ClassRewriter extends Rewriter {
     if (decSym.flags & ts.SymbolFlags.Alias) {
       decSym = this.typeChecker.getAliasedSymbol(decSym);
     }
-    return decSym.getDocumentationComment().some(c => c.text.indexOf('@Annotation') >= 0);
+    for (let d of decSym.getDeclarations()) {
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
+      // For example using '@Annotation' in a true comment.
+      // However, a new TS API would be needed, track at
+      // https://github.com/Microsoft/TypeScript/issues/7393.
+      let range = ts.getLeadingCommentRanges(d.getFullText(), 0);
+      if (!range) return;
+      for (let {pos, end} of range) {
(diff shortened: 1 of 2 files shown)


src/decorator-annotator.ts
```diff
@@ -57,7 +57,19 @@ class ClassRewriter extends Rewriter {
     if (decSym.flags & ts.SymbolFlags.Alias) {
       decSym = this.typeChecker.getAliasedSymbol(decSym);
     }
-    return decSym.getDocumentationComment().some(c => c.text.indexOf('@Annotation') >= 0);
+    for (let d of decSym.getDeclarations()) {
+      // yuk, getLeadingCommentRangesOfNode is a private API, using it until
+      // https://github.com/Microsoft/TypeScript/issues/7393 is resolved.
+      // Switch to the TS JSDoc parser in the future to avoid false positives here.
```
evmar: See getJSDoc() in tsickle.ts for another approach that doesn't use private API.  It's about as long as what you wrote anyway.

rkirov: Good call. I didn't realize that some of the utilities.ts functions were exposed.

I decided against extracting into a comment tsickle helper to return JSDoc text, because it is just two calls into the TS apis and I hope we can switch to using the TS JSDoc parser eventually. WDYT?

PR #316, opened 2017-01-10
ts2.1: only check public type flags

evmar:
NOTE: Some TypeFlags are marked "internal" in the API but still show
up in the value of type.flags.  Use a mask to limit the flag checks to
the ones in the public API.

More work on #295.  This fixes a crash in the "index_import" test.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -188,7 +188,13 @@ export class TypeTranslator {
     // NOTE: Though type.flags has the name "flags", it usually can only be one
     // of the enum options at a time.  This switch handles all the cases in
     // the ts.TypeFlags enum in the order they occur.
-    switch (type.flags) {
+    // NOTE: Some TypeFlags are marked "internal" in the d.ts but still show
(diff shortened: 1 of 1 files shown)


src/type-translator.ts
```diff
@@ -188,7 +188,12 @@ export class TypeTranslator {
     // NOTE: Though type.flags has the name "flags", it usually can only be one
     // of the enum options at a time.  This switch handles all the cases in
     // the ts.TypeFlags enum in the order they occur.
-    switch (type.flags) {
+    // NOTE: Some TypeFlags are marked "internal" in the d.ts but still show
+    // up in the value of type.flags.  This mask limits the flag checks to
+    // the ones in the public API.
```
rkirov: can you add a link to the typescript.d.ts on github, so one can easily access it when lastFlag gets moved and we have to  upgrade this.

evmar: I extended the comment.  Linking on github doesn't work because the github d.ts is different than the version we depend on in this code.

PR #321, opened 2017-01-11
Support for Annotations declared with let bindings.

rkirov:
Also adds comments why property access is not handled yet.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -48,6 +48,9 @@ class ClassRewriter extends Rewriter {
         case ts.SyntaxKind.CallExpression:
           node = (node as ts.CallExpression).expression;
           break;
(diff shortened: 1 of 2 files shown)


src/decorator-annotator.ts
```diff
@@ -62,10 +65,17 @@ class ClassRewriter extends Rewriter {
       // For example using '@Annotation' in a true comment.
       // However, a new TS API would be needed, track at
       // https://github.com/Microsoft/TypeScript/issues/7393.
-      let range = ts.getLeadingCommentRanges(d.getFullText(), 0);
+      let commentNode: ts.Node = d;
+      // Not handling PropertyAccess expressions here, because they are
+      // filtered earlier.
+      if (commentNode.kind == ts.SyntaxKind.VariableDeclaration) {
```
evmar: ===

PR #322, opened 2017-01-11
Extract TsickleCompilerHost from tsc_wrapped.

LucasSloan:
Pulls TsickleCompilerHost from tsc_wrapped, as well as some tests from google3.  Also wires the TsickleCompilerHost into main.ts.


evmar:
Do you know why Travis is failing?

src/modules_manifest.ts
```diff
@@ -0,0 +1,32 @@
+export interface FileMap<T> { [fileName: string]: T; }
+
+/** A class that maintains the module dependency graph of output JS files. */
```
evmar: Is it worth documenting this more?  I never really understood what it did.

src/modules_manifest.ts
```diff
@@ -0,0 +1,32 @@
+export interface FileMap<T> { [fileName: string]: T; }
```
evmar: In tsickle we now can use ES6 Map<> types.  It would require changing the code below to use map.set(key, val) and map.get(key) and so on.  Maybe for another change.

src/tsickle.ts
```diff
@@ -16,6 +16,8 @@ import {toArray} from './util';
 
 export {convertDecorators} from './decorator-annotator';
 export {processES5} from './es5processor';
+export {FileMap, ModulesManifest} from './modules_manifest';
```
evmar: I think there's also a ts.FileMap.  I think since anyone that uses tsickle also must use ts, it might be worth just using their type.  Though maybe the ES6 map thing I just mentioned means we don't want that...

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
+import * as ts from 'typescript';
+
+import {ModulesManifest} from './modules_manifest';
+import * as tsickle from './tsickle';
```
evmar: Is this a circular dependency?  It seems tsickle imports this as well.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
...
+        return this.runTsickle(sourceFile, this.oldProgram, fileName, languageVersion);
+      default:
+        throw new Error('tried to use TsickleCompilerHost with unknown pass enum');
+    }
+  }
+
+  writeFile(
+      fileName: string, content: string, writeByteOrderMark: boolean,
+      onError?: (message: string) => void, sourceFiles?: ts.SourceFile[]): void {
+    // Strip any potential blaze-bin path (e.g. when compiling a generated
```
evmar: We shouldn't mention blaze-bin in the open source code.

LucasSloan:
The tests I added are timing out.  They should be moved into an e2e_test file.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (this.options.googmodule && !fileName.match(/\.d\.ts$/)) {
+      content = this.convertCommonJsToGoogModule(fileName, content);
+    }
+
+    this.delegate.writeFile(fileName, content, writeByteOrderMark, onError, sourceFiles);
+  }
+
+  convertCommonJsToGoogModule(fileName: string, content: string): string {
+    // Strip off the file name extension.
+    fileName = fileName.substring(0, fileName.lastIndexOf('.'));
```
evmar: Do you know why we do this?  Does it make sense in the open source code?

test/tsickle_compiler_host_test.ts
```diff
@@ -0,0 +1,107 @@
...
+  });
+});
+
+export function formatDiagnostics(diagnostics: ts.Diagnostic[]): string {
+  const diagnosticsHost: ts.FormatDiagnosticsHost = {
+    getCurrentDirectory: () => ts.sys.getCurrentDirectory(),
+    getNewLine: () => ts.sys.newLine,
+    // Print filenames including their rootDir, so they can be located on disk
+    getCanonicalFileName: (f: string) => f
+  };
```
evmar: Is this different than tsickle.formatDiagnostics?  How?

src/modules_manifest.ts
LucasSloan: Does that require 2.1?

src/tsickle.ts
LucasSloan: I think we should stick with the wart we know.

src/tsickle_compiler_host.ts
LucasSloan: I don't know.  @mprobst, do you know?

src/tsickle_compiler_host.ts
LucasSloan: It is, but everything seems to be working fine - should I be concerned that nothing's breaking?

src/tsickle_compiler_host.ts
LucasSloan: Done.

test/tsickle_compiler_host_test.ts
LucasSloan: It is different, but we should probably use the tsickle one.

src/modules_manifest.ts
LucasSloan: I don't know how to document it more than it already is.

src/modules_manifest.ts
evmar: No, the Map thing was some months ago.

alexeagle: No, it has been available for a long time

src/modules_manifest.ts
evmar: E.g. what is moduleToFileName?  (Is module the path to the module, its contents, some identifier?)

What is referencedModules?  a map of module to modules, or file to modules

src/tsickle.ts
evmar: SGTM.  Changing this to use an ES6 map is easier with this type declared.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
...
+   */
+  shouldSkipTsickleProcessing: (fileName: string) => boolean;
+  /** Converts a path to a module name */
+  pathToModuleName: (context: string, importPath: string) => string;
+  /** Tsickle treats warnings as errors, if true, ignore warnings */
+  shouldIgnoreWarningsForPath: (filePath: string) => boolean;
+}
+
+/**
+ * TsickleCompilerHost repeats the work of Google3CompilerHost, but with
```
alexeagle: dangling reference to google3

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
...
+  prelude: string;
+}
+
+export interface TsickleEnvironment {
+  /**
+   * If true, tsickle and decorator downlevel processing will be skipped for
+   * that file.
+   */
+  shouldSkipTsickleProcessing: (fileName: string) => boolean;
+  /** Converts a path to a module name */
```
evmar: This description is obvious from the function.  Instead can you say why someone would define it, or the sorts of transformations?  I think this does e.g. "foo/bar" => "foo.bar".  What does the context param do?

src/tsickle_compiler_host.ts
evmar: As I recall one layer expects a filename with a .ts and another one doesn't.  It would be good to pin down which layer requires what (see my other comments about better documenting what the parameters are) to figure out whether this belongs here or inside g3.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,199 @@
...
+    if (isDefinitions && this.environment.shouldSkipTsickleProcessing(fileName)) return sourceFile;
+
+    let {output, externs, diagnostics} =
+        annotate(program, sourceFile, {untyped: !this.options.tsickleTyped});
+    if (externs) {
+      this.externs[fileName] = externs;
+    }
+    if (this.environment.shouldIgnoreWarningsForPath(sourceFile.path)) {
+      // All diagnostics (including warnings) are treated as errors.
+      // But for third-party code we won't fix warnings so just discard them.
```
evmar: Fix "third-party" here (?)

src/util.ts
```diff
@@ -21,3 +23,68 @@ export function toArray<T>(iterator: Iterator<T>): T[] {
...
+    }
+    return delegate.getSourceFile(path, languageVersion, onError);
+  }
+}
+
+/**
+ * Constructs a new ts.CompilerHost that overlays sources in substituteSource
+ * over another ts.CompilerHost.
+ *
+ * @param substituteSource A map of source file name -> overlay source text.
```
alexeagle: name doesn't match the param

src/util.ts
```diff
@@ -21,3 +23,68 @@ export function toArray<T>(iterator: Iterator<T>): T[] {
...
+ */
+export function createOutputRetainingCompilerHost(
+    outputFiles: Map<string, string>, delegate: ts.CompilerHost): ts.CompilerHost {
```
evmar: Out of curiosity, why do this instead of passing the extra arg to program.emit()?  I think they have the extra param exactly for doing the sort of file-writing hook we do.

test/e2e_main_test.ts
```diff
@@ -0,0 +1,44 @@
...
+import {SourceMapConsumer} from 'source-map';
+import * as ts from 'typescript';
+
+import {Settings, toClosureJS} from '../src/main';
+
+describe('main', () => {
+  it('creates externs, adds type comments and rewrites imports', function() {
+    const diagnostics: ts.Diagnostic[] = [];
+
+    const closure = toClosureJS(
```
evmar: It looks like this test is testing toClosureJS, not main (?)

test/e2e_tsickle_compiler_host_test.ts
```diff
@@ -0,0 +1,98 @@
+import {expect} from 'chai';
+import * as ts from 'typescript';
+
+import {pathToModuleName} from '../src/cli_support';
+import {formatDiagnostics} from '../src/tsickle';
+import {Pass, TsickleCompilerHost, TsickleCompilerHostOptions, TsickleEnvironment} from '../src/tsickle_compiler_host';
+import {createSourceReplacingCompilerHost} from '../src/util';
+
+function createTsickleCompilerHostOptions(): TsickleCompilerHostOptions {
```
alexeagle: instead of factory functions, why not just a const value?

src/tsickle_compiler_host.ts
LucasSloan: It looks like this should be google3 only.

src/tsickle_compiler_host.ts
LucasSloan: I think third party is a generic enough term.

src/util.ts
LucasSloan: I maybe should have done it that way, but I was copy and pasting from the other branch where I patched main to use a compiler host, and in that one, the compiler host captured the input itself.

test/e2e_main_test.ts
LucasSloan: That's correct, I've changed the describe call.

src/tsickle_compiler_host.ts
LucasSloan: Good catch.

src/util.ts
LucasSloan: Good catch.

test/e2e_tsickle_compiler_host_test.ts
LucasSloan: Why not indeed!

src/tsickle_compiler_host.ts
LucasSloan: And pulled out.

src/tsickle_compiler_host.ts
LucasSloan: I tried my hand at a rewrite - take another look?

LucasSloan:
@evmar, @alexeagle take another look?

src/tsickle_compiler_host.ts
evmar: I agree it's generic; my objection is that we're talking about third-party code here when the function we're using doesn't say anything about third-party.  (If anything, the mention of third part should be on the definition of the interface as an example of the sorts of code you might want to ignore warnings on)

src/tsickle_compiler_host.ts
```diff
@@ -31,6 +31,8 @@ export interface TsickleEnvironment {
   pathToModuleName: (context: string, importPath: string) => string;
   /** Tsickle treats warnings as errors, if true, ignore warnings */
   shouldIgnoreWarningsForPath: (filePath: string) => boolean;
+  /** Determines what we monkey patch the module.id to be in googmodule mode */
```
evmar: Can you elaborate here?  I don't understand the comment.

src/modules_manifest.ts
evmar: (don't forget about this one)

LucasSloan:
Do those work for you?

src/tsickle_compiler_host.ts
```diff
@@ -29,9 +29,16 @@ export interface TsickleEnvironment {
...
   pathToModuleName: (context: string, importPath: string) => string;
-  /** Tsickle treats warnings as errors, if true, ignore warnings */
+  /**
+   * Tsickle treats warnings as errors, if true, ignore warnings.  This might be
+   * useful for e.g. third party code.
+   */
   shouldIgnoreWarningsForPath: (filePath: string) => boolean;
-  /** Determines what we monkey patch the module.id to be in googmodule mode */
+  /**
+   * If we do googmodule processing, we monkey patch module.id, since that's
```
evmar: Oh one last comment: I'm not sure "monkey patch" is the right term here -- that usually refers to patching out some API, but in this case we're backporting a newer API.  Maybe the term is "polyfill"?

PR #324, opened 2017-01-12
fix build by implementing ts.SymbolWriter

evmar:
This fails on Travis, but how did it build before?
So confused.


test_files/decorator/decorator.js
```diff
@@ -52,8 +52,7 @@ function DecoratorTest_tsickle_Closure_declarations() {
 let DecoratedClass = class DecoratedClass {
 };
 DecoratedClass = __decorate([
-    classDecorator,
-    __metadata("design:paramtypes", [])
```
rkirov: verified this is expected change between TS 2.1.4 and 2.1.5

yarn.lock
```diff
@@ -1865,17 +1859,17 @@ tslint@^3.15.1:
...
-type-detect@^1.0.0:
-  version "1.0.0"
-  resolved "https://registry.yarnpkg.com/type-detect/-/type-detect-1.0.0.tgz#762217cc06db258ec48908a1298e8b95121e8ea2"
-
 type-detect@0.1.1:
   version "0.1.1"
   resolved "https://registry.yarnpkg.com/type-detect/-/type-detect-0.1.1.tgz#0ba5ec2a885640e470ea4e8505971900dac58822"
 
-typescript@^2.0.0:
-  version "2.0.3"
```
rkirov: Since travis is broken doesn't it mean yarn on travis is not using this lock?

evmar: I figured this out, I think.  If you just run "yarn" on ToT with no other changes, it recreates the .lock file (I guess because we modified package.json?).  And that updated .lock file gets the new version.

PR #326, opened 2017-01-13
fix(jsdoc): escape unknown tags rather than omit them

### LINKED ISSUE #325: Upgrading tsickle for angular causes JSDoc to be too escapy
upgrading from tsickle@0.2.2 to 0.2.3 causes undesirable escaping in Angular's JSDoc

eg.
```
/**
 * Construct {\@link CompileDirectiveMetadata} from {\@link ComponentTypeMetadata} and a selector.
 * @param {?} typeReference
 * @param {?} compMeta
 * @return {?}
 */
```

we shouldn't need to escape `@link`

Also the `@stable` annotation is completely dropped.

blocks https://github.com/angular/angular/pull/13471
/cc @IgorMinar

alexeagle:
Fixes #325

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -152,19 +152,6 @@ export function parse(comment: string): {tags: Tag[], warnings?: string[]}|null 
       } else if (arrayIncludes(JSDOC_TAGS_WITH_TYPES, tagName) && text[0] === '{') {
         warnings.push('type annotations (using {...}) are redundant with TypeScript types');
         continue;
-      } else if (!arrayIncludes(JSDOC_TAGS_WHITELIST, tagName)) {
-        // Silently drop tags we don't understand.  This is a subtle
-        // compromise between multiple issues.
-        // 1) If we pass through these non-Closure tags, the user will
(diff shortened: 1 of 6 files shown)


src/jsdoc.ts
```diff
@@ -177,14 +166,29 @@ export function parse(comment: string): {tags: Tag[], warnings?: string[]}|null
...
+        // 1) If we pass through these non-Closure tags, the user will
+        //    get a warning from Closure, and the point of tsickle is
+        //    to insulate the user from Closure.
+        // 2) The output of tsickle is for Closure but also may be read
+        //    by humans, for example non-TypeScript users of Angular.
+        // 3) Finally, we don't want to warn because users should be
+        //    free to add whichever JSDoc they feel like.  If the user
+        //    wants help ensuring they didn't typo a tag, that is the
+        //    responsibility of a linter.
+        tag.escape = true;
```
evmar: Rather than adding an attribute here, you can just look it up down when we call tagToString.

alexeagle: yes, nice simplification, ptal

PR #327, opened 2017-01-13
Tsickle compiler host with source maps

LucasSloan:
Add logic to TsickleCompilerHost for combining the source maps from tsickle transforms with the tsc source map.


src/modules_manifest.ts
```diff
@@ -0,0 +1,34 @@
+export interface FileMap<T> { [fileName: string]: T; }
+
+/** A class that maintains the module dependency graph of output JS files. */
+export class ModulesManifest {
+  // Map of googmodule module name to file name
```
mprobst: Use `/** JSDoc */` for documentation comments.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import * as ts from 'typescript';
+
+import {convertDecorators} from './decorator-annotator';
+import {processES5} from './es5processor';
+import {ModulesManifest} from './modules_manifest';
+import {annotate} from './tsickle';
+
+export enum Pass {
+  None,
```
mprobst: `CONST_CASE` for enum constants.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+  tsickleTyped: boolean;
+  prelude: string;
+}
+
+export interface TsickleEnvironment {
+  /**
+   * If true, tsickle and decorator downlevel processing will be skipped for
+   * that file.
+   */
+  shouldSkipTsickleProcessing: (fileName: string) => boolean;
```
mprobst: Are you intentionally using properties that are function valued instead of methods? I.e. why not `interface X { foo(bar: baz): bam; }`?

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+export interface TsickleCompilerHostOptions {
+  googmodule: boolean;
+  es5Mode: boolean;
+  tsickleTyped: boolean;
+  prelude: string;
+}
+
+export interface TsickleEnvironment {
```
mprobst: I think in the rest of the TS world, this is called a `TsickleHost`.

mprobst: Please add a documentation comment that describes this type.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
+import * as path from 'path';
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import * as ts from 'typescript';
+
+import {convertDecorators} from './decorator-annotator';
+import {processES5} from './es5processor';
+import {ModulesManifest} from './modules_manifest';
+import {annotate} from './tsickle';
+
+export enum Pass {
```
mprobst: Please add docs on the processing model implied by this, it's not immediately clear what a pass is, when it happens, etc.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+  private tsickleSourceMaps = new Map<string, SourceMapGenerator>();
+
+  constructor(
+      private delegate: ts.CompilerHost, private options: TsickleCompilerHostOptions,
+      private environment: TsickleEnvironment, private oldProgram?: ts.Program,
+      private pass = Pass.None) {}
+
+
+  public reconfigureForRun(program: ts.Program, pass: Pass) {
+    this.oldProgram = program;
```
mprobst: oldProgram needs docs.

What's the overall strategy here - we essentially wrap programs in layers, and modify them slightly in subsequent runs?

If so, would this possibly be easier to understand if we exchanged the entire compiler host for subsequent runs, instead of changing state in the existing compiler host? That was the previous design, can you document what this class does and what wouldn't quite work with the previous design?

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  getSourceFile(
+      fileName: string, languageVersion: ts.ScriptTarget,
+      onError?: (message: string) => void): ts.SourceFile {
+    if (this.pass === Pass.None) {
+      return this.delegate.getSourceFile(fileName, languageVersion, onError);
+    }
+
+    if (!this.oldProgram) {
+      throw new Error('tried to run a pass other than None without setting a program');
```
mprobst: It's a bit of a code smell that it's possible to use this code incorrectly like that. Should we maybe not use this compiler host for the situation of the `None` pass?

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (converted.output === fileContent) {
+      // No changes; reuse the existing parse.
+      return sourceFile;
+    }
+    fileContent = converted.output + this.ANNOTATION_SUPPORT;
+    this.decoratorDownlevelSourceMaps.set(fileName, converted.sourceMap);
+    return ts.createSourceFile(fileName, fileContent, languageVersion, true);
+  }
+
+  private runTsickle(
```
mprobst: `run...` sounds like you're executing something for a side effect. Maybe `transformWithTsickle(sourceFile)` is closer to what this code does? Same above.

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+  sourceMapTextToConsumer(sourceMapText: string): SourceMapConsumer {
+    const sourceMapJson: any = sourceMapText;
+    return new SourceMapConsumer(sourceMapJson);
+  }
+
+  combineSourceMaps(tscSourceMapText: string): string {
+    const tscSourceMapConsumer = this.sourceMapTextToConsumer(tscSourceMapText);
+    const tscSourceMapGenerator = this.sourceMapConsumerToGenerator(tscSourceMapConsumer);
+    if (this.tsickleSourceMaps.size > 0) {
+      for (const sourceFileName of (tscSourceMapConsumer as any).sources) {
```
mprobst: can you leave a TODO in places where you cast through any?

src/tsickle_compiler_host.ts
```diff
@@ -0,0 +1,259 @@
...
+  fileNameToModuleId: (fileName: string) => string;
+}
+
+/**
+ * TsickleCompilerHost does tsickle processing of input files, including
+ * closure type annotation processing, decorator downleveling and
+ * require -> googmodule rewriting.
+ */
+export class TsickleCompilerHost implements ts.CompilerHost {
+  private ANNOTATION_SUPPORT = `
```
mprobst: should be `private static` for a deep constant (which `CONST_CASE` implies). But prefer extracting to a local variable in the module, `const ANNOTATION_SUPPORT = ...`, no need to use a class field.

src/main.ts
```diff
@@ -179,58 +149,36 @@ function toClosureJS(
...
-      allDiagnostics.push(...diagnostics);
-      return null;
-    }
-    tsickleOutput.set(ts.sys.resolvePath(fileName), output);
-    if (externs) {
-      tsickleExterns += externs;
-    }
-  }
+  const tsickleEnvironment: tsickle.TsickleEnvironment = {
+    shouldSkipTsickleProcessing: (fileName) => fileNames.indexOf(fileName) === -1,
```
mprobst: do we care that this is O(n^2)? What's the expected cardinality of `fileNames`?

src/modules_manifest.ts
LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: Good to know, thanks!

LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: Nope!  Changed.

src/tsickle_compiler_host.ts
LucasSloan: Good to know, thanks!

src/tsickle_compiler_host.ts
LucasSloan: The real problem here is that I want the constructor to accept either a pass and an oldProgram or neither.  I've found a way to ensure that happens, so I've removed the check and error.

src/tsickle_compiler_host.ts
LucasSloan: I decided to go with downlevelDecorators() and annotateTypes().

src/tsickle_compiler_host.ts
LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: Done.

src/main.ts
LucasSloan: I don't think so - maybe a few hundred?

PR #329, opened 2017-01-17
add @implements tag on "implements" clauses

### LINKED ISSUE #284: Add @implements tag on "implements" clauses
```js
class Foo implements Bar {}
```

becomes just plain `class Foo{}` in ES6 output.  We need to tag it with an `@implements` tag so Closure knows to not erase functions required by the implementation.

evmar:
This is made a bit subtle by the fact that Closure and TypeScript
disagree about how "implements" works around classes.  However,
with the logic implemented here a simple test that verifies
assignment compatibility passes.

Fixes #284.


evmar:
I think it doesn't.

PR #336, opened 2017-01-20
be careful around type aliases in @implements

### LINKED ISSUE #333: @implements error in zone.ts
Got an error like ` ERROR - Could not resolve type in @implements tag of Zone`.

Code looks like:

```
interface Zone { ... }
interface ZoneType { ... }
type AmbientZone = Zone;
const Zone: ZoneType = (function(global: any) {
  class Zone implements AmbientZone {
```

Perhaps the type alias or the intermediate variable name is confusing things.

evmar:
It's legal to @implements a type alias, but we need to be careful
to not @implements a type alias that maps back to a symbol that
we didn't emit a type for due to the symbol being both a type
and a value.  Yikes!

To ensure I got this right, I added both a test for:
- implementing an ordinary interface via an alias
- a more complex reproduction of the zone issue

Fixes #333 harder.

### DIFF
--- a/test_files/class/class.js
+++ b/test_files/class/class.js
@@ -31,18 +31,46 @@ class Extends extends Super {
      */
     interfaceFunc() { }
 }
+/** @typedef {!Interface} */
+var TypeAlias;
+/**
+ * @implements {TypeAlias}
+ * @extends {Super}
+ */
+class ImplementsTypeAlias {
+    /**
+     * @return {void}
+     */
+    interfaceFunc() { }
+    /**
+     * @return {void}
+     */
+    superFunc() { }
+}
 // Verify Closure accepts the various casts.
 let /** @type {!Interface} */ interfaceVar;
 interfaceVar = new Implements();
 interfaceVar = new Extends();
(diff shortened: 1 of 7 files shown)


test_files/class/class.js
```diff
@@ -31,18 +31,46 @@ class Extends extends Super {
      */
     interfaceFunc() { }
 }
+/** @typedef {!Interface} */
+var TypeAlias;
+/**
+ * @implements {TypeAlias}
```
evmar: Note: here is where we successfully implement a type alias.

test_files/class/class.js
```diff
@@ -31,18 +31,46 @@ class Extends extends Super {
...
+function Zone() { }
+class ZoneImplementsInterface {
+}
+function ZoneImplementsInterface_tsickle_Closure_declarations() {
+    /** @type {string} */
+    ZoneImplementsInterface.prototype.zone;
+}
+/** @typedef {?} */
+var ZoneAlias;
+class ZoneImplementsAlias {
```
evmar: And this class (and the one above it) both *don't* @implements anything, due to the interface not existing.

PR #341, opened 2017-01-26
Support angular's tsc_wrapped in with the tsickle_compiler_host.

LucasSloan:
Add an option to TsickleCompilerHostOptions and supporting plumbing to convert index import shorthand, which the angular version of tsc_wrapped needs.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -153,6 +153,7 @@ export function toClosureJS(
     es5Mode: false,
     tsickleTyped: !settings.isUntyped,
     prelude: '',
+    convertIndexImportShorthand: false,
   };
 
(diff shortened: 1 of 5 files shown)


alexeagle:
@evmar - Lucas is plumbing through this one:
https://github.com/angular/tsickle/commit/8bec2e733dd1f1ff9d6e07073b74a810edce0eee
through the new Host pattern

LucasSloan:
@alexeagle @evmar 

I've merged the 2 tsickle options interfaces - does it look better?

evmar:
(Note: hold off on landing anything on master until we get TS2.1 merged into google3 -- I had to revert my other recent change because it broke in subtle ways.)

PR #343, opened 2017-01-26
Add support for ts.SymbolFlags.ObjectLiteral

bendilts:
This allows members of classes that are complex object literals to retain type safety. For example,

```
class Thing {
  public foo = {
    bar:4,
    baz:'asdf'
  };
}
```
emits
```
function Thing_tsickle_Closure_declarations() {
    /** @type {{bar:number,baz:string}} */
    Thing.prototype.foo;
}
```
rather than
```
function Thing_tsickle_Closure_declarations() {
    /** @type {?} */
    Thing.prototype.foo;
}
```

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -339,7 +339,8 @@ export class TypeTranslator {
         return '?';
       }
 
-      if (type.symbol.flags === ts.SymbolFlags.TypeLiteral) {
+      if (type.symbol.flags === ts.SymbolFlags.TypeLiteral ||
+          type.symbol.flags === ts.SymbolFlags.ObjectLiteral) {
         return this.translateTypeLiteral(type);
       } else if (
           type.symbol.flags === ts.SymbolFlags.Function ||
@@ -358,7 +359,6 @@ export class TypeTranslator {
       Tuple
       Mapped
       Instantiated
-      ObjectLiteral
       EvolvingArray
       ObjectLiteralPatternWithComputedProperties
     */
--- a/test_files/class/class.js
+++ b/test_files/class/class.js
@@ -74,3 +74,15 @@ function ZoneImplementsAlias_tsickle_Closure_declarations() {
     /** @type {string} */
     ZoneImplementsAlias.prototype.zone;
 }
+class HasObjectliteral {
+    constructor() {
+        this.foo = {
+            bar: 0,
+            baz: ''
+        };
+    }
+}
+function HasObjectliteral_tsickle_Closure_declarations() {
+    /** @type {{bar: number, baz: string}} */
+    HasObjectliteral.prototype.foo;
+}
--- a/test_files/class/class.ts
+++ b/test_files/class/class.ts
@@ -47,3 +47,10 @@ type ZoneAlias = Zone;
 class ZoneImplementsAlias implements ZoneAlias {
   zone: string;
 }
+
+class HasObjectliteral {
+  public foo = {
+    bar: 0,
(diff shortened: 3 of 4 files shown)


bendilts:
I signed it!

evmar:
Can you add a test?  See test_files/type/type.ts, and just add an example there.
See README.md for how to run the test runner to have it update the files for you.

bendilts:
So, it turns out that `.members` isn't a property of the `ts.ObjectType` interface. My latest push just hard-casts through `any` to get to a type that works correctly. I can't seem to find the right interface to cast it to, any input there would be helpful.

I'll put together a unit test.

evmar:
I think you might be able to add a check

```
    } else if (type.objectFlags & ts.ObjectFlags.ObjectLiteral) {
```
before the line
```
    } else if (type.objectFlags & ts.ObjectFlags.Anonymous) {
```
and then

bendilts:
There's already an object literal test in type.ts, which is confusing given the output and code I was seeing. I'll add one in `class.ts` where there is a member that's an object literal, like the example I gave in this PR's original message.

evmar:
I wonder if there's some way to merge or reuse the code here:
https://github.com/angular/tsickle/blob/5dbc5c1806ff060e662c8be3b8c55f3423ee5030/src/type-translator.ts#L372

I notice that iterates through .members on something.  What do you think?

bendilts:
Good call, I feel much better about this now.

evmar:
I meant, can we reuse or refactor that code?  It seems it's doing the same (?) thing.  I'm not sure, but it's not clear to me right now why this is different than that code.

bendilts:
Oh, man, I think you're right. This is about to turn into a one-line pull request. This function was dealing with `TypeLiteral` but not `ObjectLiteral`, but I think it can just execute the same code for either.

evmar:
In practice, most of my tsickle changes are one line of code and N lines of tests that are carefully reproducing the problems.  See e.g https://github.com/angular/tsickle/commit/667c71b88a9ed56c42f2e2b90f3d6e32334432c1 or https://github.com/angular/tsickle/commit/2255b53724fe38d7bf2f732e56dbee583889b50c  ;)

bendilts:
OK, it's down to one line, and I've left my member-of-class unit test.

bendilts:
My change allows tsickle to now output a type annotation that's not imported into the current module (since you can have a member of a class that implicitly takes on a type from another module that you haven't imported).

I'll have to somehow keep a list of all the types imported into the current module, and emit an `import` statement if something winds up having a type that isn't already imported. This is going to get a little more complex 😞

evmar:
Can you give an example snippet of the problematic code?

evmar:
(In general, if TypeScript is able to compile the code it means all types have been resolved, so you shouldn't need to emit import statements.)

bendilts:
For an example of the problem, consider the factory pattern:
```
import {ThingFactory} from './thing';
export class Foo {
  constructor(private thingFactory:ThingFactory) {}
  public thing = this.thingFactory.create();
}
```
Here, the type annotator puts `/** @type {Thing} */` in front of `Foo.prototype.thing`, but `Thing` isn't actually imported anywhere, so we get Closure compiler errors. I noticed it first with this change, but actually isn't related to this pull request, so we should probably deal with it separately (if at all).

mprobst:
@evmar @bendilts this PR hasn't seen traffic in four months. Please re-open if it is still current.

PR #344, opened 2017-01-26
translate "declare global" properly in externs

### LINKED ISSUE #330: Externs writer doesn't respect namespace global
When generating a externs file from a d.ts file, tsickle tranforms this:

declare global {
  namespace Foo {
    ...
  }
}

into global.Foo.(...) instead of Foo.(...).

evmar:
"declare global" effectively means "reset the namespace to the
global namespace".

As part of this, I fixed the test runner to support multiple
files generating externs.

Fixes #330.


evmar:
cc @LucasSloan

evmar:
@martinprobst can you take a look at this old review, which I've since updated?

I think I convinced myself that the one issue (marked with a TODO here) predates this change and shouldn't hold it up.

PR #346, opened 2017-01-27
Install mocha for development.

mprobst:
It is required by the launch configurations.


evmar:
Why has this always worked for me without this until now?

mprobst:
> Why has this always worked for me without this until now?

Probably because you once in your life installed mocha, and then never cleared your `node_modules` folder? Reproducibility is for suckers :-)

PR #347, opened 2017-01-27
Support exported {interface,class,function,variable} declarations.

mprobst:
These already generated appropriate externs symbols, but did not
generate a symbol on the exports object of the module, leading to
Closure Compiler failures when importing the symbol.

This code interprets exported declared symbols as global symbols that
are accessible under the same name, consistent with previous code.
That's somewhat odd though - we could equally well interpret it as
symbols that the user promises should magically appear on the exports
object. This does not really matter for Closure Compiler compatibility
though.


mprobst:
This might need some more work. Trying it out internally, this will break JS where AJD drops the generated externs. There are also odd issues like:

```ts
export declare interface Foo { x: number; }
export class Foo {}
```

... which leads to the class Foo actually having a `x` defined. People actually do that o_O.

evmar:
Elsewhere in tsickle we do a thing where detect whether there's a type/value conflict and avoid it.
E.g. you can't emit `/** @interface */ var Foo;` on the first Foo because the later class wants to be the value.

mprobst:
Friendly ping.

mprobst:
@evmar is there a way how I can convince you to consider this pull request? Pretty please?

mprobst:
@evmar done wrt copying the folder, PTAL.

PR #348, opened 2017-01-27
Support exported {interface,class,function,variable} declarations.

mprobst:
These already generated appropriate externs symbols, but did not
generate a symbol on the exports object of the module, leading to
Closure Compiler failures when importing the symbol.

This code interprets exported deVclared symbols as global symbols that
are accessible under the same name, consistent with previous code.
That's somewhat odd though - we could equally well interpret it as
symbols that the user promises should magically appear on the exports
object. This does not really matter for Closure Compiler compatibility
though.


src/tsickle.ts
```diff
@@ -374,6 +374,30 @@ class Annotator extends ClosureRewriter {
       sourceMap: annotated.sourceMap,
     };
   }
+
+  getExportDeclarationName(node: ts.Node): string {
+    switch (node.kind) {
+      case ts.SyntaxKind.VariableStatement:
+        const varDecl = node as ts.VariableStatement;
+        if (varDecl.declarationList.declarations.length !== 1) break;
```
evmar: Why not a loop here?
I imagine it's code like:

export let foo = 3, bar = 4;

src/tsickle.ts
```diff
@@ -374,6 +374,30 @@ class Annotator extends ClosureRewriter {
...
+      case ts.SyntaxKind.ClassDeclaration:
+        const decl = node as ts.Declaration;
+        if (!decl.name || decl.name.kind !== ts.SyntaxKind.Identifier) {
+          break;
+        }
+        return getIdentifierText(decl.name as ts.Identifier);
+      default:
+        break;
+    }
+    this.error(
```
evmar: What about "export type X = ..."

src/tsickle.ts
```diff
@@ -386,6 +410,23 @@ class Annotator extends ClosureRewriter {
...
+      // For Closure Compiler, such declarations must still be exported, so that importing code in
+      // other modules can reference them. Because tsickle generates global symbols for such types,
+      // the appropriate semantics are referencing the global name.
+      if (node.flags & ts.NodeFlags.Export) {
+        const declName = this.getExportDeclarationName(node);
+        if (node.kind === ts.SyntaxKind.VariableStatement) {
+          // For variables, TypeScript rewrites every reference to the variable name as an
+          // "exports." access, to maintain mutable ES6 exports semantics. Indirecting through the
+          // window object means we reference the correct global symbol. Closure Compiler does
+          // understand that "var foo" in externs corresponds to "window.foo".
```
evmar: Great comment!

evmar:
Oh also: I haven't been landing patches because we are trying to not break the TS2.1 merge.  I landed an innocuous-seeming change (around @implements) and found it unfixable in g3 so I had to revert it.  So maybe hold off on landing this for another day or two.

src/tsickle.ts

src/tsickle.ts

evmar:
I had hoped we'd have ts2.1 in quicker than this, but it always seems to take weeks.  :~(
If you hold off on this a day or so it might make the migration easier.  (I worry that someone will write code that depends on this behavior in 2.0 and then have it break in 2.1, or that it will change behavior in some way in 2.1.)

mprobst:
SGTM, this can wait a bit.

PR #355, opened 2017-02-01
Pass typeBlackListPaths down from TsickleCompilerHost.

mprobst:
Also adds an e2e test for typed mode.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tsickle",
-  "version": "0.2.4",
+  "version": "0.2.6",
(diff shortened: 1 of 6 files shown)


LucasSloan:
The changes I'm working on will accomplish this result, by combining the TsickleCompilerHostOptions and TsickleOptions interfaces, so I don't think this should get merged.  If it's critical that this happen quickly, I can cherry pick the changes combining the 2 and add your tests.  Otherwise, I'd like to wait.  If that's ok, please remind me to pick up your tests when my PR is opened.

mprobst:
I'm fine with your stuff taking precedence. You might be able to salvage the tests from this?

PR #356, opened 2017-02-02
Handle source maps with relative paths specifying their sources.

LucasSloan:
If the input and output files are in different directories, tsc specifies the source file as a relative path from the output file to the input file.  This wasn't exposed until I tried to incorporate the source map changes into Angular's tsc_wrapped.

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -171,7 +171,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
     const tscSourceMapConsumer = this.sourceMapTextToConsumer(tscSourceMapText);
     const tscSourceMapGenerator = this.sourceMapConsumerToGenerator(tscSourceMapConsumer);
 
-    const fileDir = path.parse(filePath).dir;
+    const fileDir = path.dirname(filePath);
 
     if (this.tsickleSourceMaps.size > 0) {
       // TODO(lucassloan): remove when the .d.ts has the correct types
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -155,7 +155,8 @@ export class TsickleCompilerHost implements ts.CompilerHost {
     return SourceMapGenerator.fromSourceMap(sourceMapConsumer);
   }
 
-  sourceMapGeneratorToConsumerWithFileName(sourceMapGenerator: SourceMapGenerator, fileName: string): SourceMapConsumer {
+  sourceMapGeneratorToConsumerWithFileName(
+      sourceMapGenerator: SourceMapGenerator, fileName: string): SourceMapConsumer {
     const rawSourceMap = sourceMapGenerator.toJSON();
     rawSourceMap.file = fileName;
     return new SourceMapConsumer(rawSourceMap);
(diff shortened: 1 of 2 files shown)


test/e2e_source_map_test.ts
```diff
@@ -73,6 +73,69 @@ describe('source maps', () => {
...
+
+    const {line: stringXLine, column: stringXColumn} = getLineAndColumn(compiledJS, 'a string');
+    const {line: stringBLine, column: stringBColumn} = getLineAndColumn(compiledJS, 'fourth rate');
+
+    expect(sourceMap.originalPositionFor({line: stringXLine, column: stringXColumn}).line)
+        .to.equal(3, 'first string definition');
+    expect(sourceMap.originalPositionFor({line: stringXLine, column: stringXColumn}).source)
+        .to.equal('a/b/input1.ts', 'first input file');
+    expect(sourceMap.originalPositionFor({line: stringBLine, column: stringBColumn}).line)
+        .to.equal(4, 'fouth string definition');
```
alexeagle: fourth spelled wrong - which makes me wonder what's the point of the extra string arg here?
Also it feels like this test asserts some of the same things as the other test. don't you only need to assert that the filename handling is right?

src/tsickle_compiler_host.ts
```diff
@@ -165,25 +167,32 @@ export class TsickleCompilerHost implements ts.CompilerHost {
     return new SourceMapConsumer(sourceMapJson);
   }
 
-  combineSourceMaps(tscSourceMapText: string): string {
+  combineSourceMaps(filePath: string, tscSourceMapText: string): string {
     const tscSourceMapConsumer = this.sourceMapTextToConsumer(tscSourceMapText);
     const tscSourceMapGenerator = this.sourceMapConsumerToGenerator(tscSourceMapConsumer);
+
+    const fileDir = path.parse(filePath).dir;
```
mprobst: path.dirname?

test/e2e_source_map_test.ts
```diff
@@ -73,6 +73,69 @@ describe('source maps', () => {
...
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, 'output.js', new Set<string>(['input2.ts']));
+
+    const {line: methodLine, column: methodColumn} = getLineAndColumn(compiledJS, 'methodName');
+    const {line: stringBLine, column: stringBColumn} = getLineAndColumn(compiledJS, 'fourth rate');
+
+    expect(sourceMap.originalPositionFor({line: methodLine, column: methodColumn}).line)
+        .to.equal(7, 'method definition');
+    expect(sourceMap.originalPositionFor({line: methodLine, column: methodColumn}).source)
```
mprobst: I don't see where this code tests the actual path resolving, it just asserts that all file paths are identical to the original paths. What test would break if we didn't call path.resolve in the implementation?

src/tsickle_compiler_host.ts
LucasSloan: That works better, thanks!

test/e2e_source_map_test.ts
LucasSloan: You're correct.  This is a separate bug that I discovered while working on this.  If you have some files getting decorator downleveled and others not, previously you'd get an undefined val exception when writing out because we'd try to compose decorator downlevel source maps with tsickle source maps when the decorator downlevel source maps weren't there.  I think this test would make more sense if I changed input2.ts to contain some annotations and change some of the comments.

test/e2e_source_map_test.ts
LucasSloan: Good catch on the spelling, thanks!

LucasSloan: I want to make sure that tsickle is still doing its work - otherwise you could make this test pass by having tsickle do no work.

PR #358, opened 2017-02-02
Angular tsickle changes onto g3 tsickle version

LucasSloan:


evmar:
Is it intentional to revert @implements here?

PR #360, opened 2017-02-07
Inline source maps

LucasSloan:
We currently handle source maps that are in separate files from the compiled js tsc outputs, but we don't handle source maps in a comment in the compiled js.  This PR handles extracting the source map from the comment, composing it, and adding a new comment with the composed map.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "typescript": "^2.1.0"
   },
   "devDependencies": {
+    "@types/base-64": "^0.1.0",
     "@types/chai": "^3.4.32",
     "@types/glob": "^5.0.29",
     "@types/google-closure-compiler": "0.0.18",
@@ -28,6 +29,7 @@
     "@types/node": "^6.0.38",
     "@types/source-map": "^0.1.27",
     "@types/source-map-support": "^0.2.27",
+    "base-64": "^0.1.0",
     "chai": "^3.5.0",
     "clang-format": "^1.0.45",
     "glob": "^7.0.0",
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
     "typescript": "^2.1.0"
   },
   "devDependencies": {
-    "@types/base-64": "^0.1.0",
     "@types/chai": "^3.4.32",
     "@types/glob": "^5.0.29",
     "@types/google-closure-compiler": "0.0.18",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,6 @@
     "@types/node": "^6.0.38",
     "@types/source-map": "^0.1.27",
     "@types/source-map-support": "^0.2.27",
-    "base-64": "^0.1.0",
     "chai": "^3.5.0",
     "clang-format": "^1.0.45",
     "glob": "^7.0.0",
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -1,3 +1,4 @@
+import {encode} from 'base-64';
 import * as path from 'path';
 import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
@@ -6,6 +7,7 @@ import {convertDecorators} from './decorator-annotator';
 import {processES5} from './es5processor';
 import {ModulesManifest} from './modules_manifest';
 import {annotate} from './tsickle';
+import {extractInlineSourceMap} from './util';
 
 /**
  * Tsickle can perform 2 different precompilation transforms - decorator downleveling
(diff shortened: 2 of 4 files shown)


src/tsickle_compiler_host.ts
```diff
@@ -209,6 +214,15 @@ export class TsickleCompilerHost implements ts.CompilerHost {
     return tscSourceMapGenerator.toString();
   }
 
+  combineInlineSourceMaps(filePath: string, tscOutputText: string): string {
```
evmar: Can you add some jsdoc to this about the parameters?  (e.g. what is tscOutputText)

src/tsickle_compiler_host.ts
```diff
@@ -209,6 +214,15 @@ export class TsickleCompilerHost implements ts.CompilerHost {
...
+    const sourceMapJson = extractInlineSourceMap(tscOutputText);
+    const composedSourceMap = this.combineSourceMaps(filePath, sourceMapJson);
+    const encodedSourceMap = encode(composedSourceMap);
+    return tscOutputText.replace(
+        new RegExp('//# sourceMappingURL=data:application/json;base64,(.*)'),
```
evmar: Should these regexps be more specific?  For example, should they be anchored to the beginning of the line?  Does "." match newlines?

src/util.ts
```diff
@@ -1,3 +1,4 @@
+import {decode} from 'base-64';
 import * as ts from 'typescript';
```
evmar: Somehow these are ending up above the license header, can you put them below?

LucasSloan: Done.

src/tsickle_compiler_host.ts
LucasSloan: I've anchored them to the beginning and end of line.

src/tsickle_compiler_host.ts
LucasSloan: I've changed tscOutputText to compiledJsWithInlineSourceMap.

src/util.ts
```diff
@@ -91,7 +91,7 @@ export function createOutputRetainingCompilerHost(
 }
 
 export function extractInlineSourceMap(source: string): string {
-  const regex = new RegExp('//# sourceMappingURL=data:application/json;base64,(.*)');
+  const regex = new RegExp('^//# sourceMappingURL=data:application/json;base64,(.*)$', 'm');
   const result = regex.exec(source)!;
```
evmar: What happens if this fails to match?  (Is that possible?)

package.json
```diff
@@ -28,6 +29,7 @@
     "@types/node": "^6.0.38",
     "@types/source-map": "^0.1.27",
     "@types/source-map-support": "^0.2.27",
+    "base-64": "^0.1.0",
```
evmar: Is this library in g3?  We'll have to import it if so.
It might be possible to just use builtin node APIs for base64:
https://nodejs.org/docs/latest/api/buffer.html#buffer_buffers_and_character_encodings

test/e2e_source_map_test.ts
```diff
@@ -160,6 +160,32 @@ describe('source maps', () => {
...
+    sources.set('input.ts', `
+      class X { field: number; }
+      let x : string = 'a string';
+      let y : string = 'another string';
+      let z : string = x + y;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, undefined, undefined, true);
+
+    const {line: stringXLine, column: stringXColumn} = getLineAndColumn(compiledJS, 'a string');
```
evmar: I think this would read better as

```
{
  const {line, column} =  ...
  expect ... sourceMap.originalPositionFor({line, column})...
}
{
  const {line, column} =  ...
  expect ... sourceMap.originalPositionFor({line, column})...
}
```

src/util.ts
LucasSloan: Tsc should always produce the same source map comment, so since we're protected by a check that the proper flag was passed we should be good.

package.json
LucasSloan: That buffer library worked - good catch!

test/e2e_source_map_test.ts
LucasSloan: That pattern is used a lot of places - mind if I do another PR for that?

package.json
```diff
@@ -18,6 +18,7 @@
     "typescript": "^2.1.0"
   },
   "devDependencies": {
+    "@types/base-64": "^0.1.0",
```
evmar: drop this

test/e2e_source_map_test.ts
```diff
@@ -229,21 +255,32 @@ function tsickleCompiler(
...
 
+  let compilerOptions: ts.CompilerOptions;
+  if (inlineSourceMap) {
+    compilerOptions = {
+      inlineSourceMap: inlineSourceMap,
+      outFile: outFile,
+      experimentalDecorators: true
+    };
+  } else {
+    compilerOptions = {sourceMap: true, outFile: outFile, experimentalDecorators: true};
```
evmar: if you write this with an extra comma:
```
experimentalDecorators: true,};
```
clang-format will lay it out linearly which makes it match the line above.

PR #364, opened 2017-02-09
Support preexisisting inline source maps.

LucasSloan:
If tsickle is passed sources that include inline source maps (eg ts 'sources' produced by ngc), compose those source maps with the source maps produced by running tsickle.

### DIFF
--- a/src/source_map_utils.ts
+++ b/src/source_map_utils.ts
@@ -1,0 +1,83 @@
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+
+/**
+ * Return a new RegExp object every time we want one because the
+ * RegExp object has internal state that we don't want to persist
+ * between different logical uses.
+ */
+function getInlineSourceMapRegex(): RegExp {
+  return new RegExp('^//# sourceMappingURL=data:application/json;base64,(.*)$', 'mg');
+}
+
+export function containsInlineSourceMap(source: string): boolean {
+  return getInlineSourceMapCount(source) > 0;
+}
+
+export function getInlineSourceMapCount(source: string): number {
+  const match = source.match(getInlineSourceMapRegex());
+  return match ? match.length : 0;
+}
+
+export function extractInlineSourceMap(source: string): string {
+  const inlineSourceMapRegex = getInlineSourceMapRegex();
+  let previousResult: RegExpExecArray|null = null;
+  let result: RegExpExecArray|null = null;
+  // We want to extract the last source map in the source file
+  // since that's probably the most recent one added.  We keep
+  // matching against the source until we don't get a result,
+  // then we use the previous result.
+  do {
+    previousResult = result;
+    result = inlineSourceMapRegex.exec(source);
+  } while (result !== null);
+  const base64EncodedMap = previousResult![1];
+  return Buffer.from(base64EncodedMap, 'base64').toString('utf8');
+}
+
+export function removeInlineSourceMap(source: string): string {
+  return source.replace(getInlineSourceMapRegex(), '');
+}
+
+/**
+ * Sets the source map inline in the file.  If there's an existing inline source
+ * map, it clobbers it.
+ */
+export function setInlineSourceMap(source: string, sourceMap: string): string {
+  const encodedSourceMap = Buffer.from(sourceMap, 'utf8').toString('base64');
+  if (containsInlineSourceMap(source)) {
+    return source.replace(
+        getInlineSourceMapRegex(),
+        `//# sourceMappingURL=data:application/json;base64,${encodedSourceMap}`);
+  } else {
+    return `${source}\n//# sourceMappingURL=data:application/json;base64,${encodedSourceMap}`;
+  }
+}
+
+export function sourceMapConsumerToGenerator(sourceMapConsumer: SourceMapConsumer):
+    SourceMapGenerator {
+  return SourceMapGenerator.fromSourceMap(sourceMapConsumer);
+}
+
+/**
+ * Tsc identifies source files by their relative path to the output file.  Since
+ * there's no easy way to identify these relative paths when tsickle generates its
+ * own source maps, we patch them with the file name from the tsc source maps
+ * before composing them.
+ */
+export function sourceMapGeneratorToConsumerWithFileName(
+    sourceMapGenerator: SourceMapGenerator, fileName: string): SourceMapConsumer {
+  const rawSourceMap = sourceMapGenerator.toJSON();
+  rawSourceMap.sources = [fileName];
+  rawSourceMap.file = fileName;
+  return new SourceMapConsumer(rawSourceMap);
+}
+
+export function sourceMapTextToConsumer(sourceMapText: string): SourceMapConsumer {
+  const sourceMapJson: any = sourceMapText;
+  return new SourceMapConsumer(sourceMapJson);
+}
+
+export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenerator {
+  const sourceMapJson: any = sourceMapText;
+  return SourceMapGenerator.fromSourceMap(sourceMapTextToConsumer(sourceMapJson));
+}
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -1,12 +1,12 @@
 import * as path from 'path';
-import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import {SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
 import {convertDecorators} from './decorator-annotator';
 import {processES5} from './es5processor';
 import {ModulesManifest} from './modules_manifest';
+import * as sourceMapUtils from './source_map_utils';
 import {annotate, isDtsFileName} from './tsickle';
-import {extractInlineSourceMap} from './util';
 
 /**
  * Tsickle can perform 2 different precompilation transforms - decorator downleveling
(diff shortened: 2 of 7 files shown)


src/tsickle_compiler_host.ts
```diff
@@ -155,71 +157,70 @@ export class TsickleCompilerHost implements ts.CompilerHost {
...
-  getSourceMapKey(outputFilePath: string, sourceFileName: string): string {
-    const fileDir = path.dirname(outputFilePath);
+  stripAndStoreExistingSourceMap(sourceFile: ts.SourceFile): void {
+    if (sm.containsInlineSourceMap(sourceFile.text)) {
+      const sourceMapJson = sm.extractInlineSourceMap(sourceFile.text);
+      const sourceMap = sm.sourceMapTextToGenerator(sourceMapJson);
+      this.preexistingSourceMaps.set(this.getSourceMapKeyForSourceFile(sourceFile), sourceMap);
 
-    return this.getCanonicalFileName(path.resolve(fileDir, sourceFileName));
+      sourceFile.text = sm.removeInlineSourceMap(sourceFile.text);
```
mprobst: It's not generally safe to mutate the TypeScript AST. We've run into very hard to track bugs because of it before. Can you extract the source map on load?

src/tsickle_compiler_host.ts
```diff
@@ -125,6 +126,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
     }
 
     const sourceFile = this.runConfiguration.oldProgram.getSourceFile(fileName);
+    this.stripAndStoreExistingSourceMap(sourceFile);
```
mprobst: does this need to happen on every getSourceFile? Shouldn't we only do this when loading the file from disk the very first time?

Also: how often does `getSourceFile` get called? We shove megabytes through this, running a regexp over and over again on all inputs could get quite costly.

src/source_map_utils.ts
```diff
@@ -0,0 +1,72 @@
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+
+const INLINE_SOURCE_MAP_REGEX =
+    new RegExp('^//# sourceMappingURL=data:application/json;base64,(.*)$', 'm');
+
+export function containsInlineSourceMap(source: string): boolean {
+  return getInlineSourceMapCount(source) > 0;
+}
+
+export function getInlineSourceMapCount(source: string): number {
```
mprobst: this is only used in tests, is it really important to expose this and test it separately?

src/source_map_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (match) {
+    // The INLINE_SOURCE_MAP_REGEX has a capture group, but we only want to count real matches
+    return match.length / 2;
+  } else {
+    return 0;
+  }
+}
+
+export function extractInlineSourceMap(source: string): string {
+  const result = INLINE_SOURCE_MAP_REGEX.exec(source)!;
```
mprobst: we've hard issues before where source code would contain:

```javascript

// #sourceMapURL...
more().code();
// #sourceMap...
```

I.e., you should take the last source map, not the first.

src/source_map_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function containsInlineSourceMap(source: string): boolean {
+  return getInlineSourceMapCount(source) > 0;
+}
+
+export function getInlineSourceMapCount(source: string): number {
+  const match = source.match(INLINE_SOURCE_MAP_REGEX);
+  if (match) {
+    // The INLINE_SOURCE_MAP_REGEX has a capture group, but we only want to count real matches
+    return match.length / 2;
```
mprobst: Are you sure this really counts the matches?

```
> m = 'a a a a a a a'.match(/(a)/m).length
2
```

Maybe you're missing a `g`?

test/e2e_source_map_test.ts
```diff
@@ -196,6 +200,33 @@ describe('source maps', () => {
...
+    decoratorDownlevelSources.set('input.ts', `/** @Annotation */
+        function classAnnotation(t: any) { return t; }
+
+        @classAnnotation
+        class DecoratorTest {
+          public methodName(s: string): string { return s; }
+        }`);
+
+    const closurizeSources = new Map<string, string>();
+    {
```
mprobst: drop the extra block?

src/source_map_utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export function setInlineSourceMap(source: string, sourceMap: string): string {
+  const encodedSourceMap = Buffer.from(sourceMap, 'utf8').toString('base64');
+  if (INLINE_SOURCE_MAP_REGEX.test(source)) {
+    return source.replace(
+        INLINE_SOURCE_MAP_REGEX,
+        `//# sourceMappingURL=data:application/json;base64,${encodedSourceMap}`);
+  } else {
+    return `${source}
+//# sourceMappingURL=data:application/json;base64,${encodedSourceMap}`;
```
mprobst: I think this would be clearer to read if you just used `\n` in the string.

src/source_map_utils.ts
LucasSloan: It isn't part of the public api and I'd like to keep all the source map functions together.

src/source_map_utils.ts
LucasSloan: I was missing a `g`, thanks!

src/source_map_utils.ts
LucasSloan: That is better, yes.

src/tsickle_compiler_host.ts
LucasSloan: Moved into the first load.

src/tsickle_compiler_host.ts
LucasSloan: Moved into the initial load.

src/source_map_utils.ts
LucasSloan: Done.

test/e2e_source_map_test.ts
LucasSloan: Done.

src/tsickle_compiler_host.ts
```diff
@@ -1,12 +1,12 @@
 import * as path from 'path';
-import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import {SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
 import {convertDecorators} from './decorator-annotator';
 import {processES5} from './es5processor';
 import {ModulesManifest} from './modules_manifest';
+import * as sm from './source_map_utils';
```
evmar: The Go rule around short names is that the more local and contextual the name (e.g. the iterator for a "for" loop), the more likely you are to know the name in context (e.g. "i").  For a global name like this that appears across the file, a longer name like `sourceMapUtils` is more appropriate.

(The JS/TS style rule is to just always use long names.)

src/tsickle_compiler_host.ts
```diff
@@ -96,13 +96,16 @@ export class TsickleCompilerHost implements ts.CompilerHost {
   /** externs.js files produced by tsickle, if any. */
   public externs: {[fileName: string]: string} = {};
 
+  private sourceFileToPreexistingSourceMapMap = new Map<ts.SourceFile, SourceMapGenerator>();
```
evmar: You can drop one "map" suffix here, because it's implied in the type.  (E.g. the other maps below don't have Map at the end.)

test/source_map_utils_test.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import {expect} from 'chai';
+
+import * as sm from '../src/source_map_utils';
```
evmar: same comment here about module name, but the short name is okay-er here because it's a short file.

src/source_map_utils.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function sourceMapTextToConsumer(sourceMapText: string): SourceMapConsumer {
+  const sourceMapJson: any = sourceMapText;
+  return new SourceMapConsumer(sourceMapJson);
+}
+
+export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenerator {
+  const sourceMapJson: any = sourceMapText;
+  return SourceMapGenerator.fromSourceMap(this.sourceMapTextToConsumer(sourceMapJson));
```
evmar: What's the idea behind this function and the one above it?  They seem to just wrap another function?

I think the use of `this` here accidentally compiles but it shouldn't.

src/tsickle_compiler_host.ts
LucasSloan: Fair enough, but why is typescript imported as ts?

evmar: Hah, that's a good point!  I blame @mprobst

mprobst:
I'd claim that "ts" is an obvious abbreviation in this context, whereas
"sm" is not?

src/source_map_utils.ts
LucasSloan: You were right!  I was reusing the same regex object, which moves the pointer in the string forward each time, rendering my test that this was happening invalid.

src/source_map_utils.ts
evmar: Which object?  I think this function is a top-level function, not a method (?)

src/source_map_utils.ts
```diff
@@ -0,0 +1,77 @@
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
+
+function getInlineSourceMapRegex(): RegExp {
```
mprobst: document why this is not using a constant.

src/source_map_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+export function getInlineSourceMapCount(source: string): number {
+  const match = source.match(getInlineSourceMapRegex());
+  return match ? match.length : 0;
+}
+
+export function extractInlineSourceMap(source: string): string {
+  const inlineSourceMapRegex = getInlineSourceMapRegex();
+  let previousResult: RegExpExecArray|null = null;
+  let result: RegExpExecArray|null = null;
+  do {
```
mprobst: comment why this is looping

src/source_map_utils.ts
```diff
@@ -0,0 +1,77 @@
...
+    previousResult = result;
+    result = inlineSourceMapRegex.exec(source);
+  } while (result !== null);
+  const base64EncodedMap = previousResult![1];
+  return Buffer.from(base64EncodedMap, 'base64').toString('utf8');
+}
+
+export function removeInlineSourceMap(source: string): string {
+  if (containsInlineSourceMap(source)) {
```
mprobst: can you double check if you need the if() here? and below

src/source_map_utils.ts
LucasSloan: Done.

src/source_map_utils.ts
LucasSloan: Not needed here, needed below.

src/source_map_utils.ts
LucasSloan: Wow, my bad.  I thought this was still where I had originally written it, despite you calling attention to it.  Fixed.

src/source_map_utils.ts
LucasSloan: Done.

PR #366, opened 2017-02-10
Don't canonicalize the fileName in write file.

### LINKED ISSUE #365: writeFile creates lower-case filenames on Mac
Tsickle now converts to canonicalFileName when writing files, here:
https://github.com/angular/tsickle/blob/aae914dc74721f9a62792e9559f70a29669d2245/src/tsickle_compiler_host.ts#L144
I think this behavior was copied from the compilerHost in google3.

This means that when using tsickle on Mac (eg. running `@angular/tsc-wrapped` on rxjs in the new integration test), the outputs for `Observable.ts` are `observable.js`, `observable.d.ts`, and `Observable.js.map` (note how the `.map` file is special-cased in that tsickle code)

This will break any Angular users developing on Macs.

LucasSloan:
We need to canonicalize for g3 support, but we should (and already do) leave that to the host delegate.

Fixes #365

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -141,7 +141,6 @@ export class TsickleCompilerHost implements ts.CompilerHost {
       fileName: string, content: string, writeByteOrderMark: boolean,
       onError?: (message: string) => void, sourceFiles?: ts.SourceFile[]): void {
     if (path.extname(fileName) !== '.map') {
-      fileName = this.delegate.getCanonicalFileName(fileName);
       if (this.options.googmodule && !isDtsFileName(fileName)) {
         content = this.convertCommonJsToGoogModule(fileName, content);
       }


src/tsickle_compiler_host.ts
```diff
@@ -141,7 +141,6 @@ export class TsickleCompilerHost implements ts.CompilerHost {
       fileName: string, content: string, writeByteOrderMark: boolean,
       onError?: (message: string) => void, sourceFiles?: ts.SourceFile[]): void {
     if (path.extname(fileName) !== '.map') {
-      fileName = this.delegate.getCanonicalFileName(fileName);
```
evmar: Is there a test that this affects?

I think you should check this against g3 before you submit, because I think this was important for blaze-bin vs blaze-genfiles

LucasSloan: I took a look at tsc_wrapped, and it already does the canonicalization before writing.  But yes, definitely should test it first.

alexeagle:
I merged this because Angular needs a release that fixes the issue.

LucasSloan:
Got it.

PR #367, opened 2017-02-10
Don't try to compose inline source maps in .d.ts files.

LucasSloan:
.d.ts files don't have any source maps, so trying to extract an inline source map from one results in an error.

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -5,7 +5,7 @@ import * as ts from 'typescript';
 import {convertDecorators} from './decorator-annotator';
 import {processES5} from './es5processor';
 import {ModulesManifest} from './modules_manifest';
-import {annotate} from './tsickle';
+import {annotate, isDtsFileName} from './tsickle';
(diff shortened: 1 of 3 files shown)


src/tsickle_compiler_host.ts
```diff
@@ -145,7 +145,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
       if (this.options.googmodule && !fileName.match(/\.d\.ts$/)) {
         content = this.convertCommonJsToGoogModule(fileName, content);
       }
-      if (this.tscOptions.inlineSourceMap) {
+      if (!/.*\.d\.ts/.test(path.basename(fileName)) && this.tscOptions.inlineSourceMap) {
```
evmar: There's this `isDtsFileName` function in tsickle.ts, maybe we could reuse that?  (I noticed that the regex is unanchored here)

LucasSloan: Changed, and replaced 2 other bare regexes with that function.

PR #370, opened 2017-02-13
Add exporting decorators.

drjanitor:
When a decorator has @ExportsDecoratedItems in its documentation, dump the names of all items it decorates to the externs.


src/tsickle.ts
```diff
@@ -72,6 +72,32 @@ function hasModifierFlag(node: ts.Node, flag: ts.ModifierFlags): boolean {
...
+      node.decorators.some(decorator => isExportingDecorator(decorator, typeChecker));
+}
+
+/**
+ * @return true if the given decorator has an @ExportDecoratedItems directive in its javadoc.
+ */
+function isExportingDecorator(decorator: ts.Decorator, typeChecker: ts.TypeChecker) {
+  return getDecoratorDeclarations(decorator, typeChecker).some(declaration => {
+    let range = ts.getLeadingCommentRanges(declaration.getFullText(), 0);
+    if (range) {
```
mprobst: prefer early exits:

```
if (!range) return false;
...
```

src/tsickle.ts
```diff
@@ -375,21 +406,39 @@ class Annotator extends ClosureRewriter {
       sourceMap: annotated.sourceMap,
     };
   }
+
+  private formatExportingDecoratorExterns() {
+    if (this.exportingDecoratorExterns.length === 0) {
+      return '';
+    } else {
```
mprobst: remove else block, prefer early exits :-)

src/tsickle.ts
```diff
@@ -375,21 +406,39 @@ class Annotator extends ClosureRewriter {
...
+      return '\n' + this.exportingDecoratorExterns.map(name => `var ${name};\n`).join('');
```
mprobst: is `var ${name}` really the right pattern here? Shouldn't these be declared properties on an anonymous object, as we do for templates?

src/util.ts
```diff
@@ -89,6 +89,30 @@ export function createOutputRetainingCompilerHost(
...
+  // Walk down the expression to find the identifier of the decorator function.
+  let node: ts.Node = decorator;
+  while (node.kind !== ts.SyntaxKind.Identifier) {
+    switch (node.kind) {
+      case ts.SyntaxKind.Decorator:
+      case ts.SyntaxKind.CallExpression:
+        node = (node as ts.Decorator | ts.CallExpression).expression;
+        break;
+      default:
+        throw new Error();
```
mprobst: give a message?

src/tsickle.ts
```diff
@@ -359,13 +388,15 @@ class Annotator extends ClosureRewriter {
...
-    if (externs.output) {
+
+    if (externs.output.length > 0 || this.exportingDecoratorExterns.length > 0) {
       externsSource = `/**
  * @externs
  * @suppress {duplicate}
  */
 // NOTE: generated by tsickle, do not edit.
-` + externs.output;
+` + externs.output +
```
mprobst: Should these properties really end up in an externs file? For templates, we add them to a regular source file, and then rely on `@export` annotations on property declarations, which I think is quite different. In particular, I think variables and properties are handled differently. Have you tested this?

mprobst:
Overall approach looks reasonable by the way.

evmar:
If you'd like to add a test, make a new directory that contains a file that exercises this code to `test_files/mydirectory`, then run `UPDATE_GOLDENS=1 TEST_FILTER=mydirectory gulp watch` and look at the output files.

src/util.ts
drjanitor: Could you or Evan clarify what the message that used to be in decorator-annotator.ts meant?

        // PropertyAccessExpression is intentionally missing here,
        // because the rest of the rewriter does not handle such
        // expressions.

Would this refer to decorators of the form `@foo.bar`?

src/tsickle.ts
drjanitor: It seems to me that, whatever I do here, this results in the name to be "blessed" and treated magically by the Closure compiler. At least, I did not see any particular differences there.

src/tsickle.ts
drjanitor: I tested this piece of code, and it seems to do exactly what I mean it to do.

Adding `@export` (manually) on the fields seems to achieve the same result, but I have no clue whether things are handled even remotely similarly.

src/util.ts
evmar: @rkirov wrote this comment

rkirov: This does refer to foo.bar. The api worked as follows, if `shouldDownlevel` returns true, we call `process` on it.

When I last touched this file, I tried to add PAE, but it didn't work because `process` failed to do anything. I found it surprising too and added the comment. Technically we should be able to handle simple propaccess.

I am against throwing an error in decorators we cannot statically analyze. Remember, the original TS team design is to have dynamic decorators (like python). We are doing some static shenanigans to have annotations, but the semantics of annotations are not defined by the language and we keep on iterating on them (your cl adding even more custom semantics). The annotation rewrite should quietly continue if it cannot prove that the decorator comes from an annotation.

```javascript
function getFoo() {
 return isPrime(2^100 + 1) ? anAnnotation : aDecorator;
}

@getFoo()
class X { ... }
```

drjanitor: Your comment makes a lot of sense. I made sure not to throw on decorator types that the code can't (yet) handle.

src/tsickle.ts
drjanitor: Done.

src/tsickle.ts
drjanitor: Done.

src/util.ts
```diff
@@ -88,3 +88,26 @@ export function createOutputRetainingCompilerHost(
     outputFiles.set(fileName, content);
   }
 }
+
+/**
+ * Returns the declarations for the given decorator.
```
evmar: This "util.ts" file is becoming a junk grab-bag.  I think this function belongs in the file that uses it most closely (decorator-annotator).

evmar:
(just in case you missed it in the review comments, please add a test)

src/util.ts
drjanitor: This function is now used by both decorator-annotator.ts and tsickle.ts.
Should tsickle import it from decorator-annotator? It's hard for me to judge which file uses it "most closely", tbh.

drjanitor:
Added the test!

src/tsickle.ts
```diff
@@ -72,6 +72,33 @@ function hasModifierFlag(node: ts.Node, flag: ts.ModifierFlags): boolean {
...
+ * @return true if node has an exporting decorator  (i.e., a decorator with @ExportDecoratedItems
+ * in its javadoc).
+ */
+function hasExportingDecorator(node: ts.Node, typeChecker: ts.TypeChecker) {
+  return node.decorators &&
+      node.decorators.some(decorator => isExportingDecorator(decorator, typeChecker));
+}
+
+/**
+ * @return true if the given decorator has an @ExportDecoratedItems directive in its javadoc.
```
mprobst: Nit: Google style is to prefer "Returns ..." over the `@return` tag.

src/util.ts
mprobst: Could you move it to a file `decorators.ts`? It's better to have it in isolation than in a junkyard ;-)

mprobst:
This looks good to me, I'll defer to @evmar for merging. Can you run `git fetch origin && git rebase -i origin/master && git push -f` to get the git commit history in order?

src/tsickle.ts
```diff
@@ -72,6 +73,33 @@ function hasModifierFlag(node: ts.Node, flag: ts.ModifierFlags): boolean {
 }
 
 /**
+ * @return true if node has an exporting decorator  (i.e., a decorator with @ExportDecoratedItems
+ * in its javadoc).
+ */
+function hasExportingDecorator(node: ts.Node, typeChecker: ts.TypeChecker) {
+  return node.decorators &&
+      node.decorators.some(decorator => isExportingDecorator(decorator, typeChecker));
```
evmar: Maybe all of these functions belong in your decorators module now?

src/tsickle.ts
```diff
@@ -456,6 +496,15 @@ class Annotator extends ClosureRewriter {
       return true;
     }
 
+    if (hasExportingDecorator(node, this.program.getTypeChecker())) {
+      let {name} = node as (
+          ts.ClassDeclaration | ts.MethodDeclaration | ts.PropertyDeclaration |
+          ts.AccessorDeclaration);
+      if (name) {
```
evmar: This is weird.  The "as" expression is a coercion, but we can't guarantee that node is one of these types.  Maybe you meant to check node.kind here?

drjanitor:
my understand is that, if a node has a decorator, it has to be one of those
types?

mprobst:
>
> my understand is that, if a node has a decorator, it has to be one of those
> types?
>

If it serves as a data point, I stumbled over what Evan thought too, and
then figured it's guaranteed by the presence of a decorator. But that
definitely shows it's worth adding a comment.

src/tsickle.ts
drjanitor: Moved.

src/tsickle.ts
drjanitor: Commented.

README.md
```diff
@@ -58,6 +58,24 @@ JSON probably uses the string `'username'` and not whatever name Closure would
 invent for it.  (Note: `declare` on an interface has no additional meaning in
 pure TypeScript.)
 
+#### Exporting decorators
+
+An exporting decorator is a decorator that has `@ExportDecoratedItems` in its
+JSDoc.
+
+The name of the element that have an exporting decorator are preserved through
```
evmar: Typo: "name ... that have"

src/decorators.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  let decSym = typeChecker.getSymbolAtLocation(node);
+  if (decSym.flags & ts.SymbolFlags.Alias) {
+    decSym = typeChecker.getAliasedSymbol(decSym);
+  }
+  return decSym.getDeclarations();
+}
+
+/**
+ * Returns true if node has an exporting decorator  (i.e., a decorator with @ExportDecoratedItems
```
evmar: typo: two spaces

src/tsickle.ts
evmar: Just to make sure I'm following, you have code that ends up like:

.js: class Foo { bar() {} }
.externs.js: var bar;

And bar doesn't get renamed?  I am skeptical but I reproduced it locally so I guess it's right.  I'm not sure why this works though!

README.md
drjanitor: Done.

src/decorators.ts
drjanitor: Done.

test_files/exporting_decorator/exporting.js
```diff
@@ -0,0 +1,97 @@
...
+    set exportThisSetter(x) {
+        console.log(`I don't really care about ${x}.`);
+    }
+};
+__decorate([
+    exportingFieldDecorator(),
+    __metadata("design:type", Function),
+    __metadata("design:paramtypes", []),
+    __metadata("design:returntype", void 0)
+], NonExportedClass.prototype, "exportThisOneToo", null);
```
evmar: By the way, did you verify that this extra output (caused by using a decorator) does what you want in the final output?  Or does it bloat the code?

drjanitor:
The decorators are actually supposed to do stuff themselves, and in order
to do that they need this extra output stuff... so I guess it's unavoidable.

evmar:
Thanks for your patience with the review!

PR #373, opened 2017-02-13
enable more tsc strictness

evmar:
Group the compiler strictness flags together and enable some more.
Fix a few minor bugs they turn up.

(I believe this would've caught the stray "this" seen in a recent CL.)

### DIFF
--- a/test/e2e_main_test.ts
+++ b/test/e2e_main_test.ts
@@ -7,7 +7,6 @@
  */
 
 import {assert, expect} from 'chai';
-import {SourceMapConsumer} from 'source-map';
 import * as ts from 'typescript';
 
 import {Settings, toClosureJS} from '../src/main';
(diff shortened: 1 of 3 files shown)


test/e2e_main_test.ts
```diff
@@ -24,8 +23,7 @@ describe('toClosureJS', () => {
     if (!closure) {
       diagnostics.forEach(v => console.log(JSON.stringify(v)));
       assert.fail();
-      // TODO(lucassloan): remove when the .d.ts has the correct types
-      return {compiledJS: '', sourceMap: new SourceMapConsumer('' as any)};
+      return;
```
LucasSloan: Which of the flags tripped this?

evmar: `noImplicitReturns`.  In JS if a function doesn't return anything it implicitly returns undefined, in the sense that if you write `let x = foo();` then x ends up `undefined` if `foo` doesn't return anything.

The error was:
```
test/e2e_main_test.ts(15,66): error TS7030: Not all code paths return a value.
```
before, the compiler was assuming you wanted to return this object here and then 'undefined' below.

LucasSloan: Got it, this was a bad copy-paste on my part.  Good catch.

evmar: The compiler caught it. My life goal is to replace myself with machine automation.  ;)

PR #375, opened 2017-02-15
Fix source map key generation on windows.

### LINKED ISSUE #371: getCanonicalFileName does not produce consistent value for use in decoratorDownlevelSourceMaps on Windows
Version 0.21.2, TS 2.1.6, Win10

The maps [here](https://github.com/angular/tsickle/blob/5599c188f05d03f46d67673b0ab6927d9445e524/src/tsickle_compiler_host.ts#L99-L100) rely on getCanonicalFileName for the key.  On my machine they get set with '/' for paths, but get looked up with '\\', causing a toJSON to be called on undefined [here](https://github.com/angular/tsickle/blob/5599c188f05d03f46d67673b0ab6927d9445e524/src/tsickle_compiler_host.ts#L169).

Calling getCanonicalFileName changes casing, but does not change the slash.

LucasSloan:
path.resolve() returns paths with '\\' on Windows, but typescript always uses '/' in paths regardless of platform.  If we use path.resolve in one of the sourceMapKey methods, we have to use it in the other.

Fixes #371

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -165,7 +165,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
   }
 
   getSourceMapKeyForSourceFile(sourceFile: ts.SourceFile): string {
-    return this.getCanonicalFileName(sourceFile.path);
+    return this.getCanonicalFileName(path.resolve(sourceFile.path));
   }
 
   stripAndStoreExistingSourceMap(sourceFile: ts.SourceFile): ts.SourceFile {


evmar:
Is there a way to cover this with a test?

LucasSloan:
Alex suggested setting up a CI pipeline on Windows, which sounds like a good idea to me, but sounds like a pretty big investment, so probably shouldn't block this.

mprobst:
I looked at Travis on Windows, it doesn't exist (yet).

Lucas Sloan <notifications@github.com> schrieb am Mi., 15. Feb. 2017 um
13:36 Uhr:

PR #378, opened 2017-02-16
Fix a bug where tsickle would try to extract a non-existent source map.

mprobst:
@LucasSloan it seems that our e2e source map tests are less e2e than we'd like. I ran into this when syncing to g3. Can you take a look at improving our test coverage here, maybe by converting e2e_source_map test to run the fill compiler host?

### DIFF
--- a/src/source_map_utils.ts
+++ b/src/source_map_utils.ts
@@ -18,7 +18,7 @@ export function getInlineSourceMapCount(source: string): number {
   return match ? match.length : 0;
 }
 
-export function extractInlineSourceMap(source: string): string {
+export function extractInlineSourceMap(source: string): string|null {
   const inlineSourceMapRegex = getInlineSourceMapRegex();
   let previousResult: RegExpExecArray|null = null;
   let result: RegExpExecArray|null = null;
@@ -30,7 +30,8 @@ export function extractInlineSourceMap(source: string): string {
     previousResult = result;
     result = inlineSourceMapRegex.exec(source);
   } while (result !== null);
-  const base64EncodedMap = previousResult![1];
+  if (!previousResult) return null;
+  const base64EncodedMap = previousResult[1];
   return Buffer.from(base64EncodedMap, 'base64').toString('utf8');
 }
 
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -171,6 +171,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
   stripAndStoreExistingSourceMap(sourceFile: ts.SourceFile): ts.SourceFile {
     if (sourceMapUtils.containsInlineSourceMap(sourceFile.text)) {
       const sourceMapJson = sourceMapUtils.extractInlineSourceMap(sourceFile.text);
+      if (!sourceMapJson) return sourceFile;
       const sourceMap = sourceMapUtils.sourceMapTextToGenerator(sourceMapJson);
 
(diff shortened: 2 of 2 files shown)


src/source_map_utils.ts
```diff
@@ -30,7 +30,8 @@ export function extractInlineSourceMap(source: string): string {
     previousResult = result;
     result = inlineSourceMapRegex.exec(source);
   } while (result !== null);
-  const base64EncodedMap = previousResult![1];
+  if (!previousResult) return null;
```
LucasSloan: I'm very wary of doing this - if we're mistakenly passing this function a file without a source map, we should fix that, not make this function silently accept bad input.  It's much easier to detect and fix a variable being undefined than a source map not being correct.

src/tsickle_compiler_host.ts
```diff
@@ -238,6 +239,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
 
   combineInlineSourceMaps(filePath: string, compiledJsWithInlineSourceMap: string): string {
     const sourceMapJson = sourceMapUtils.extractInlineSourceMap(compiledJsWithInlineSourceMap);
```
LucasSloan: Did the error occur in this function call?

src/source_map_utils.ts
mprobst: So the problem was that code calling into this has to be stateful - it has to first check if a source map is there, and then call this here. Which is easy to get wrong (qed), and then the failure mode is ugly on top.

I think you could improve the API by making it stateless, i.e. by merging the "has a comment" and "do something with it" calls. That makes calling code impossible to get wrong, at no loss of functionality (and actually a gain of performance, we save some parsing of the file).

The type system allows us to express the optionality, and you could rename the method to express that, e.g. `findInlineSourceMap`.

src/tsickle_compiler_host.ts
mprobst: See the stack I copied to you.

src/source_map_utils.ts
LucasSloan: I agree that it's bad to have an api that forces you to call a guard function before you call another function, but I don't think that what's happening here.  My goal with extractInlineSourceMap is that you only call it when you're sure that the file you're calling it on has an inline source map and that it throws nasty errors if you didn't figure that out correctly.

In the errors that this PR is a response to, there were in fact inline source maps present, but they weren't on their own line.  This PR just ignores them, instead of either fixing the source map regex or fixing whatever put the inline source map on a line with other stuff.

Is there some other way to get what I want (errors when there is in fact a source map, but something went wrong) as well as what you want (a simpler api)?

mprobst: > I agree that it's bad to have an api that forces you to call a guard function before you call another function, but I don't think that what's happening here.
> My goal with extractInlineSourceMap is that you only call it when you're sure that the file you're calling it on has an inline source map and that it throws nasty errors if you didn't figure that out correctly.

This sounds contradictory to me. How can you be sure that you can call this function? By calling the guard function.

I think it's fine to have an "attempt to extract a source map, and if you don't find one, don't". That's what the code would do if it didn't fail, wouldn't it? Is the actual problem just that the two regexs don't match?

LucasSloan: Are you around tomorrow?  We should probably discuss this in person.

mprobst:
>
> Are you around tomorrow? We should probably discuss this in person.
>
Yes, I'm in the office.

>

PR #379, opened 2017-02-16
fix: escape JSDoc in property declarations

### LINKED ISSUE #377: Strip @internal/@experimental from jsdoc
Otherwise you get warnings like:
"Parse error. illegal use of unknown JSDoc tag "internal"; ignoring it"

alexeagle:
Fixes #377


test_files/jsdoc/jsdoc.tsickle.ts
```diff
@@ -30,8 +32,14 @@ function jsDocTestBadDoc(foo: string) {}
  * \@madeUpTag This tag will be escaped, because Closure disallows it.
  */
 class JSDocTest {
-  /** @export */
-  exported: string;
+/**
+ * \@internal
+ */
+static X: string[] = [];
```
alexeagle: note, the indent is getting lost here, maybe help me see why?

evmar: This is expected and fine.

PR #380, opened 2017-02-16
Use main.ts's toClosureJS() for source map tests.

LucasSloan:
Reduce verbosity and improve test coverage of our actual api by using toClosureJS() from main.ts instead of a modified copy in e2e_source_map_test.ts.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -123,39 +123,55 @@ function loadTscConfig(args: string[], allDiagnostics: ts.Diagnostic[]):
   return {options, fileNames};
 }
 
+export interface ClosureJSOptions {
+  tsickleCompilerHostOptions: tsickle.Options;
+  tsickleHost: tsickle.TsickleHost;
+  files: Map<string, string>;
+  tsicklePasses: tsickle.Pass[];
+}
+
+function getDefaultClosureJSOptions(fileNames: string[], settings: Settings): ClosureJSOptions {
+  return {
+    tsickleCompilerHostOptions: {
+      googmodule: true,
+      es5Mode: false,
+      untyped: settings.isUntyped,
+    },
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -123,39 +123,55 @@ function loadTscConfig(args: string[], allDiagnostics: ts.Diagnostic[]):
...
+
 /**
  * Compiles TypeScript code into Closure-compiler-ready JS.
  * Doesn't write any files to disk; all JS content is returned in a map.
  */
 export function toClosureJS(
     options: ts.CompilerOptions, fileNames: string[], settings: Settings,
-    allDiagnostics: ts.Diagnostic[],
-    files = new Map<string, string>()): {jsFiles: Map<string, string>, externs: string}|null {
+    allDiagnostics: ts.Diagnostic[], closureJSOPtions = {} as Partial<ClosureJSOptions>):
```
evmar: Typo: "OPtions"

evmar: Use of Partial here is coooool

src/main.ts
```diff
@@ -123,39 +123,55 @@ function loadTscConfig(args: string[], allDiagnostics: ts.Diagnostic[]):
...
+    {jsFiles: Map<string, string>, externs: string}|null {
+  const fullClosureJSOptions: ClosureJSOptions = {
```
evmar: I don't like the name "full" here because it's meaningless in the rest of the function.
What do you think?  Maybe you have a better idea?

LucasSloan: Do you think naming the parameter `partialClosureJSOptions` and this variable `closureJSOptions` is better?

evmar: Yes, that's much better than what I was thinking

PR #381, opened 2017-02-17
Emit non-value symbols (e.g. interfaces, types) as typedef aliases.

mprobst:
This avoids erroring at runtime because the symbol accessed is not
defined.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -376,7 +376,7 @@ class Annotator extends ClosureRewriter {
     };
   }
 
-  getExportDeclarationNames(node: ts.Node): string[] {
+  getExportDeclarationNames(node: ts.Node): ts.Identifier[] {
     switch (node.kind) {
       case ts.SyntaxKind.VariableStatement:
         const varDecl = node as ts.VariableStatement;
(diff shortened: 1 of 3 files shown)


src/tsickle.ts
```diff
@@ -390,16 +390,49 @@ class Annotator extends ClosureRewriter {
...
       default:
         break;
     }
     this.error(
         node, `unsupported export declaration ${ts.SyntaxKind[node.kind]}: ${node.getText()}`);
-    return ['errorUnhandledDeclarationKind'];
+    return [];
+  }
+
+  emitAmbientDeclaration(node: ts.Node) {
```
evmar: jsdoc plz

src/tsickle.ts
```diff
@@ -390,16 +390,49 @@ class Annotator extends ClosureRewriter {
...
+    this.externsWriter.visit(node);
```
evmar: put this down in the caller, not here

src/tsickle.ts
mprobst: done.

src/tsickle.ts
mprobst: done.

src/tsickle.ts
```diff
@@ -390,16 +390,48 @@ class Annotator extends ClosureRewriter {
...
+  /**
+   * Emits an ES6 export for the ambient declaration behind node, if it is indeed exported.
+   */
+  maybeEmitAmbientDeclarationExport(node: ts.Node) {
+    // In TypeScript, `export declare` simply generates no code in the exporting module, but does
+    // generate a regular import in the importing module.
+    // For Closure Compiler, such declarations must still be exported, so that importing code in
+    // other modules can reference them. Because tsickle generates global symbols for such types,
+    // the appropriate semantics are referencing the global name.
+    if (!this.options.untyped && hasModifierFlag(node, ts.ModifierFlags.Export)) {
```
evmar: Change to early return

PR #382, opened 2017-02-17
Switch to yarn

alexeagle:


LucasSloan:
I don't see any pinning logic here - is that planned for another PR and this one just verifies that yarn duplicates npm properly?

alexeagle:
There is already a `yarn.lock` file checked in. It has the pinned versions of everything. Now travis will install exactly those versions.
You should also use yarn locally now, to make sure you also get the same versions.

PR #384, opened 2017-02-17
emit externs for "declare namespace 'foo'"

evmar:
We must emit *some* externs for these modules so that their fields
aren't renamed, but there's not an obvious namespace to emit them
in.  So just put them in a weird unused namespace.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1021,25 +1021,29 @@ class ExternsWriter extends ClosureRewriter {
       case ts.SyntaxKind.ModuleDeclaration:
         let decl = <ts.ModuleDeclaration>node;
         switch (decl.name.kind) {
-          case ts.SyntaxKind.Identifier:
+          case ts.SyntaxKind.Identifier: {
             // E.g. "declare namespace foo {"
             let name = getIdentifierText(decl.name as ts.Identifier);
             if (name === undefined) break;
-            namespace = namespace.concat(name);
             if (this.isFirstDeclaration(decl)) {
               this.emit('/** @const */\n');
-              if (namespace.length > 1) {
-                this.emit(`${namespace.join('.')} = {};\n`);
-              } else {
-                this.emit(`var ${namespace} = {};\n`);
-              }
+              this.writeExternsVariable(name, namespace, '{}');
             }
-            if (decl.body) this.visit(decl.body, namespace);
-            break;
-          case ts.SyntaxKind.StringLiteral:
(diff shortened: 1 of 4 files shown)


src/tsickle.ts
```diff
@@ -1021,25 +1021,29 @@ class ExternsWriter extends ClosureRewriter {
...
-              } else {
-                this.emit(`var ${namespace} = {};\n`);
-              }
+              this.writeExternsVariable(name, namespace, '{}');
             }
-            if (decl.body) this.visit(decl.body, namespace);
-            break;
-          case ts.SyntaxKind.StringLiteral:
+            if (decl.body) this.visit(decl.body, namespace.concat(name));
+          } break;
```
rkirov: Why not have the `break` inside the block?

src/tsickle.ts
```diff
@@ -1021,25 +1021,29 @@ class ExternsWriter extends ClosureRewriter {
...
+          case ts.SyntaxKind.StringLiteral: {
             // E.g. "declare module 'foo' {" (note the quotes).
-            // Skip it.
-            break;
+            // We still want to emit externs for this module, but
+            // Closure doesn't really provide a mechanism for
+            // module-scoped externs.  For now, ignore the enclosing
+            // namespace (because this is declaring a top-level module)
+            // and emit into a fake namespace.
```
rkirov: internally closure "desugars" module into made up namespaces, so you might not be that wrong here.

src/tsickle.ts
```diff
@@ -1021,25 +1021,29 @@ class ExternsWriter extends ClosureRewriter {
...
+            namespace = ['tsickle_declare_module'];
```
mprobst: Won't this cause many symbol conflicts? Is that OK?

evmar: I'm a bit confused by it.  Using the jscomp debugger it seems to be accepted, but we have code in the other branch that attempts to be careful to not do this.  I think I'll try it out on the g3 bug and see if it helps, and land it if so.

src/tsickle.ts
evmar: If you put the `break` in the block you trip the linter (that wants each block to end in a break).  If you put them outside then the formatter produces this wacky layout.  :(  I might just choose different variable names here.

src/tsickle.ts
evmar: It seems to not break anything.  I am not clear why though.  Perhaps because this pattern (declare module 'foo') is uncommon.

PR #385, opened 2017-02-17
fix: don't emit both @param and @type on a property

alexeagle:
See b/35442358

/cc @dimvar

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -931,7 +931,9 @@ class Annotator extends ClosureRewriter {
 
     let tags = this.getJSDoc(p) || [];
     tags.push({tagName: 'type', type: this.typeToClosure(p)});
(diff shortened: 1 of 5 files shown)


src/tsickle.ts
```diff
@@ -931,7 +931,7 @@ class Annotator extends ClosureRewriter {
 
     let tags = this.getJSDoc(p) || [];
     tags.push({tagName: 'type', type: this.typeToClosure(p)});
-    this.emit(jsdoc.toString(tags));
+    this.emit(jsdoc.toString(tags, ['param']));
```
mprobst: Can you add a comment explaining why we need to filter `@param` here? That'll make it easier to add more when it comes up.

alexeagle: done

PR #388, opened 2017-02-21
Suppress late provide

mprobst:
Add a suppression for lateProvide errors.
    
Tsickle creates aliases for imports that are only used in type declarations, e.g.

      var Foo = foo_mod_1.Foo;
    
This leads Closure Compiler to report a cyclical import error (called late provide) as it thinks the module is actually being dereferenced, whereas it is only an alias for the type declaration. This change suppresses the spurious error.


mprobst:
Actually, this seems to paper over some more fundamental issues, let me see if I can find a better fix.

PR #389, opened 2017-02-21
Use explicit goog.forwardDeclare and goog.require calls to alias JSDoc types

mprobst:
Local variables aliases have a two issues:
- they confuse Closure Compiler into thinking there are cyclical imports/usages of modules
- they don't work for non-value types (i.e. typedefs)

To fix, generate extra `goog.forwardDeclare` calls that create local aliases, which can then be used to refer to imported symbols. Track all known aliases in a map, and use them whenever emitting code.

`goog.forwardDeclare` allows referencing the type without creating a hard dependency, which is required to match TypeScript's module loading semantics: TS does not emit a module load if an import is only used in type definitions. As a special case, if an entire module is only used to declare types (e.g. purely `export interface Foo {}` and type aliases), emit a hard `goog.require`, as that code would otherwise never be imported, which would lead JSCompiler to not understand the type annotations at all.

Beyond fixing above issues, this is also arguably more idiomatic code. Importing from the same module twice seems to be kosher (tested in plain .js files, checked the compiler code for a potential error/warning).


.vscode/launch.json
```diff
@@ -37,7 +37,8 @@
         "--nolazy"
       ],
       "env": {
-        "NODE_ENV": "development"
+        "NODE_ENV": "development",
+        "TEST_FILTER": "^jsdoc_types$"
```
evmar: undo this

src/tsickle.ts
```diff
@@ -126,10 +126,18 @@ const VISIBILITY_FLAGS: ts.ModifierFlags =
  * Another Rewriter subclass manages .ts => externs translation.
  */
 class ClosureRewriter extends Rewriter {
+  symbolsToAliasedNames = new Map<ts.Symbol, string>();
```
evmar: jsdoc

src/tsickle.ts
```diff
@@ -126,10 +126,18 @@ const VISIBILITY_FLAGS: ts.ModifierFlags =
...
+
   constructor(protected program: ts.Program, file: ts.SourceFile, protected options: Options) {
     super(file);
   }
 
+  private modulePrefixCounter = 0;
+
+  nextModulePrefix() {
```
evmar: jsdoc

src/tsickle.ts
```diff
@@ -472,10 +482,18 @@ class Annotator extends ClosureRewriter {
...
+          this.emit(` from '${this.resolveModuleSpecifier(exportDecl.moduleSpecifier)}';`);
+          const symbolPrefix = this.emitGoogRequire(exportDecl.moduleSpecifier);
+          if (exportDecl.exportClause) {
+            this.registerSymbolPrefixes(symbolPrefix, exportDecl.exportClause);
+          }
+        } else {
+          this.emit(';');
+        }
+        if (exportDecl.exportClause) {
+          // TODO(martinprobst): typedefs from export *.
```
evmar: i don't understand this todo

src/tsickle.ts
```diff
@@ -666,16 +684,41 @@ class Annotator extends ClosureRewriter {
...
   /**
+   * Write an `exports.` assignment for each type exported in named exports.
+   * TypeScript by itself does not export non-value symbols (e.g. interfaces, typedefs), as it
+   * expects to remove those entirely for runtime. For Closure, types must be
+   * exported as downstream code will import the type.
+   *
+   * The tsickle pass turns interfaces into values by generating a `function MyInterface() {}` for
+   * them, so in the second conversion pass, TypeScript does export a value for them. However for
+   * pure typedefs, tsickle only generates a property access with a JSDoc comment, so they need to
+   * exported explicitly here.
```
evmar: left out a word here, "need to *be* exported"

src/tsickle.ts
```diff
@@ -666,16 +684,41 @@ class Annotator extends ClosureRewriter {
...
+   */
+  private emitTypeDefExports(exports: ts.NamedExports) {
+    const typeChecker = this.program.getTypeChecker();
+    for (const e of exports.elements) {
+      let sym = typeChecker.getSymbolAtLocation(e.name);
+      if (sym.flags & ts.SymbolFlags.Alias) sym = typeChecker.getAliasedSymbol(sym);
+      const aliased = this.symbolsToAliasedNames.get(sym);
+      const isTypeAlias = sym.flags & ts.SymbolFlags.TypeAlias;
+      const name = getIdentifierText(e.name);
+      // console.log('exported:', name, typeTranslator.symbolToDebugString(sym));
```
evmar: delete this

src/tsickle.ts
```diff
@@ -666,16 +684,41 @@ class Annotator extends ClosureRewriter {
...
+      if (!isTypeAlias) continue;
+      const typeName = aliased || name;
+      this.emit(`\n/** @typdef {${typeName}} */\nexports.${name};`);
```
evmar: typo: "typedef"

Why didn't tests catch this?

src/tsickle.ts
```diff
@@ -731,25 +771,34 @@ class Annotator extends ClosureRewriter {
...
+      // inserts an artificial `const prefix = goog.require` call for the module, and then registers
+      // all symbols from this import to be prefixed.
+      // Subtlety warning: for a default import, the import name is used as the prefix. This means
+      // references to the import do not have to be prefixed, as it already has the correct name.
       if (!this.options.untyped) {
-        let localNames: string[];
-        if (importClause.name) {
+        let bindings = importClause.namedBindings;
+        if (bindings && bindings.kind !== ts.SyntaxKind.NamedImports) {
+          throw new Error(`bindings must be named imports! ${bindings.kind}`);
```
evmar: Can this happen?  Is this an "I haven't implemented this" or a "notreached()"?

It appears this `bindings` is never used so drop this variable?

src/tsickle.ts
```diff
@@ -731,25 +771,34 @@ class Annotator extends ClosureRewriter {
...
+        }
+        let localNames: ts.Identifier[] = [];
+        if (importClause.namedBindings) {
           // import a from ...;
-          localNames = [getIdentifierText(importClause.name)];
-        } else {
-          // import {a as b} from ...;
```
evmar: What about this case?  Do we no longer support it?

src/type-translator.ts
```diff
@@ -157,6 +160,8 @@ export class TypeTranslator {
   public symbolToString(sym: ts.Symbol): string {
     // This follows getSingleLineStringWriter in the TypeScript compiler.
     let str = '';
+    let alias = this.symbolsToAliasedNames.get(sym);
+    if (alias) return alias;
```
evmar: This is great.  I think we even may have a test where one case is commented-out because this logic below gets the wrong name.

test/tsickle_test.ts
```diff
@@ -133,7 +133,13 @@ testFn('golden tests', () => {
         };
         // Run TypeScript through tsickle and compare against goldens.
         let {output, externs, diagnostics} = tsickle.annotate(
-            program, program.getSourceFile(tsPath), options, {
+            program, program.getSourceFile(tsPath),
+            (c, p) => {
```
evmar: variable names are too terse!

test_files/export/export.tsickle.ts
```diff
@@ -16,4 +19,4 @@ let /** @type {number} */ export2 = 3;
 
 // This is just an import, so export5 should still be included.
 import {export5} from './export_helper';
-const export5: NeverTypeCheckMe = export5;  /* local alias for Closure JSDoc */
+const tsickle_module_prefix_4 = goog.require('test_files.export.export_helper');
```
evmar: So we end up importing every module twice?  I agree this is allowed by the Closure semantics but it's a bit disappointing.

test_files/jsdoc_types/jsdoc_types.tsickle.ts
```diff
@@ -5,29 +5,25 @@
...
 // Should be references to the symbols in module2, perhaps via locals.
-let /** @type {!ClassOne} */ useLocalClass = new ClassOne();
-let /** @type {!ClassOne} */ useLocalClassRenamed = new RenamedClassOne();
-let /** @type {!RenamedClassTwo} */ useLocalClassRenamedTwo = new RenamedClassTwo();
-let /** @type {!ClassOne} */ useLocalClassAsTypeRenamed: RenamedClassOne;
-let /** @type {!Interface} */ useLocalInterface: Interface;
-let /** @type {!ClassWithParams<number>} */ useClassWithParams: ClassWithParams<number>;
+let /** @type {!tsickle_module_prefix_1.ClassOne} */ useLocalClass = new ClassOne();
+let /** @type {!tsickle_module_prefix_1.ClassOne} */ useLocalClassRenamed = new RenamedClassOne();
+let /** @type {!tsickle_module_prefix_1.ClassTwo} */ useLocalClassRenamedTwo = new RenamedClassTwo();
```
evmar: Oh, I think this answers my above question about the renaming case.

src/tsickle.ts

src/tsickle.ts
mprobst: Removed.

src/tsickle.ts
mprobst: This code did not special case typedefs exported through `export *`, as those need to be special cased because they don't create a value, just a type. I've fixed this by implementing it.

src/tsickle.ts

src/tsickle.ts

src/tsickle.ts
mprobst: Hahaha. So it turns out, if you write:

```
/** @mopskotze {number} */
exports.Foo;
```

... that actually declares a property `Foo` on the export object, with an unknown type. Same for `@typdef`. So our Closure integration test didn't fail. In g3, it seems like we error on unknown JSDoc tags, but maybe not for comments that could just be comments in the file?

In any case, fixed.

src/tsickle.ts
mprobst: Sorry, indeed refactored away.

src/type-translator.ts
mprobst: Can you point me to it?

test/tsickle_test.ts
mprobst: Fixed.

test_files/export/export.tsickle.ts
mprobst: Actually, fixed.

It turns out if we `goog.require` everything, even if we're just using types from it, we cause AJD to re-order our sources in a way that changes loading semantics. I've sent an email on this, I think it's generally interesting.

The workaround is to just use `goog.forwardDeclare` for everything (this is the intended use case for it).

test_files/jsdoc_types/jsdoc_types.tsickle.ts
mprobst: Ack.

mprobst:
Note: I'll hold off on merging this until tsickle is synced and a bit more settled. I'll also rebase and pull apart the various commits in this.

mprobst:
Note to self: we can probably drop the `lateProvide` thing again due to using `forwardDeclare`.

.vscode/launch.json
evmar: I see your point but I worry that it's really hard to notice -- the test runner prints all of the file names regardless of whether it runs the tests or not.

src/type-translator.ts
evmar: I can't think of it right now, nor how to even find it.  :(  I will find it myself at some point.

src/source_map_utils.ts
```diff
@@ -30,7 +30,8 @@ export function extractInlineSourceMap(source: string): string {
     previousResult = result;
     result = inlineSourceMapRegex.exec(source);
   } while (result !== null);
-  const base64EncodedMap = previousResult![1];
+  if (!previousResult) return null;
+  const base64EncodedMap = previousResult[1];
```
evmar: Something is going wrong in your diff here.

.vscode/launch.json
mprobst: Our test runner prints:

```
[...]
    ✓ import_from_goog (8490ms)
    - index_import
    - interface
    - jsdoc
    - jsdoc_types
    - jsdoc_types.untyped
    - jsx
    - methods
    - nullable
    - optional
    - parameter_properties
    - static
    - structural.untyped
    - super
    - type
    - type_and_value
    - typedef
    - typedef.untyped
    - underscore
    - variables
  1 passing (9s)
  44 pending
```

That seems clear enough? We can remove it again if somebody runs into it and gets confused?

src/source_map_utils.ts
mprobst: I intentionally re-reverted Lucas' commit so I could try this out in google. No longer necessary.

mprobst:
PTAL. I've reworked the code quite a bit, I think it's more readable now, and I had to jump through quite some hoops to get Closure Compiler to accept it. See the note about when we insert `goog.require` calls in `forwardDeclare` in particular.

I've tried this out on the internal code base, tests are passing. I've also tried it on our example Angular 2 YouTube app, it compiles with full type emit with this.

src/tsickle.ts
```diff
@@ -127,6 +128,9 @@ const VISIBILITY_FLAGS: ts.ModifierFlags =
  * Another Rewriter subclass manages .ts => externs translation.
  */
 class ClosureRewriter extends Rewriter {
+  /** A mapping of aliases for symbols in the current file, used when emitting types.  */
```
evmar: Can you elaborate here on what "aliases for symbols" are?  As written this could be a TypeScript type = foo alias, a Closure typedef, or some other thing.

src/tsickle.ts
```diff
@@ -692,18 +709,42 @@ class Annotator extends ClosureRewriter {
...
+   * pure typedefs, tsickle only generates a property access with a JSDoc comment, so they need to
+   * be exported explicitly here.
+   */
+  private emitTypeDefExports(exports: ts.Symbol[]) {
+    if (this.options.untyped) return;
+    const typeChecker = this.program.getTypeChecker();
+    for (let sym of exports) {
+      if (sym.flags & ts.SymbolFlags.Alias) sym = typeChecker.getAliasedSymbol(sym);
+      const isTypeAlias =
+          sym.flags & ts.SymbolFlags.TypeAlias && (sym.flags & ts.SymbolFlags.Value) === 0;
```
evmar: Can you add some more parens here?  I have been bitten so hard by the C precedence bug around & vs == that I am terrified of this code, even if tests do show it works :)

src/tsickle.ts
```diff
@@ -692,18 +709,42 @@ class Annotator extends ClosureRewriter {
...
+      const name = sym.name;
+      if (!isTypeAlias) continue;
```
evmar: Maybe swap this line with the line above, it reads weird otherwise.

src/tsickle.ts
```diff
@@ -759,43 +797,87 @@ class Annotator extends ClosureRewriter {
...
       // Both of these forms create a local name "a", which after
       // TypeScript CommonJS compilation will become some renamed
-      // variable like "module_1.a".  But a user might still use plain
-      // "a" in some JSDoc comment, so gather up these local names for
-      // imports and make an alias for each for JSDoc purposes.
+      // variable like "module_1.default" or "module_1.a" (for default vs named bindings,
+      // respectively).
+      // tsickle references types in JSDoc. Because the module prefixes are not predictable, and
+      // because TypeScript might remove imports entirely if they are only for types, the code below
+      // inserts an artificial `const prefix = goog.require` call for the module, and then registers
```
evmar: Ahh, registering the symbols at import time is a great solution to this problem, I hadn't realized you were taking that approach until now!

src/tsickle.ts
```diff
@@ -759,43 +797,87 @@ class Annotator extends ClosureRewriter {
...
+      // all symbols from this import to be prefixed.
+      // Subtlety warning: for a default import, the import name is used as the prefix. This means
+      // references to the import do not have to be prefixed, as it already has the correct name.
       if (!this.options.untyped) {
-        let localNames: string[];
+        let symbols: ts.Symbol[] = [];
+        const typeChecker = this.program.getTypeChecker();
         if (importClause.name) {
-          // import a from ...;
```
evmar: Can you leave in these comments with the example syntax of which sort of import you're handling?  I found them helpful for understanding the nest of if-elses.  There's one below too.  Like this example here shows it's a default import, which lets me match up why it's fine to only have one symbol here.

src/tsickle.ts
```diff
@@ -759,43 +797,87 @@ class Annotator extends ClosureRewriter {
...
+    // causing a load. See below for the exception to the rule.
+    this.emit(`\nconst ${forwardDeclarePrefix} = goog.forwardDeclare('${moduleNamespace}');`);
+    const hasValues = exports.some(e => (e.flags & ts.SymbolFlags.Value) !== 0);
+    if (!hasValues) {
+      // Closure Compiler's toolchain will drop files that are never goog.require'd *before* type
+      // checking (e.g. when using --closure_entry_point or similar tools). This causes errors
+      // complaining about values not matching 'NoResolvedType', or modules not having a certain
+      // member.
+      // To fix, explicitly goog.require() modules that only export types. This should usually not
+      // cause breakagesdue to load order (as no symbols are accessible from the module - though
```
evmar: typo, "breakages due"

src/tsickle.ts
```diff
@@ -759,43 +797,87 @@ class Annotator extends ClosureRewriter {
...
+    const moduleNamespace =
+        nsImport !== null ? nsImport : this.pathToModuleName(this.file.fileName, importPath);
+    const typeChecker = this.program.getTypeChecker();
+    const exports = typeChecker.getExportsOfModule(typeChecker.getSymbolAtLocation(specifier));
+    // In TypeScript, importing a module for use in a type annotation does not cause a runtime load.
+    // In Closure Compiler, goog.require'ing a module causes a runtime load, so emitting requires
+    // here would cause a change in load order, which is observable (and can lead to errors).
+    // Instead, goog.forwardDeclare types, which allows using them in type annotations without
+    // causing a load. See below for the exception to the rule.
+    this.emit(`\nconst ${forwardDeclarePrefix} = goog.forwardDeclare('${moduleNamespace}');`);
```
evmar: These emits are weird, why have the newline first?  Usually the tsickle pattern is one \n to start a block of emits, and then each emit has its own trailing \n

test/e2e_test_closure_input_Module.js
```diff
@@ -0,0 +1,4 @@
+// Matches test_files/import_from_closure/decls.d.ts
```
evmar: Why not put these in that directory?

test/tsickle_test.ts
evmar: not fixed

test/e2e_test_closure_input_Module.js
mprobst: Done. I was worried they'd interfere with the file based tests.

test/tsickle_test.ts
mprobst: Done!

src/tsickle.ts

src/tsickle.ts
mprobst: Done, also added an explicit comparison to `0`.

src/tsickle.ts

src/tsickle.ts
mprobst: Yeah, I didn't either when I filed the PR, that's what I meant with descriptions etc being a mess.

src/tsickle.ts

src/tsickle.ts

src/tsickle.ts
mprobst: Not sure if that's the case, looking through the various emits in this file, they mostly start with a `\n`. That's also the case for the code that I replaced here (the local name stuff), and generates the correct output. If I emitted with trailing `\n`, I'd end up with incorrect line breaks.

mprobst:
PTAL.

evmar:
Did you still wanna rebase this into multiple commits?  (You mentioned it in a comment.)  I was gonna suggest that this might be better written as multiple reviews but I don't feel strongly about it.  Do you want me to review before or after that?

mprobst:
@evmar ah, misunderstanding. I wanted to keep the commits as is to make it easier for you to review, and then rebase/squash/... once you LGTM.

I've rebased and mostly just squashed. It turns out that the change to use `forwardDeclare` causes the problem with typedefs in our tests, so those have to go together. The `goog.require` for type only modules could go separately in theory, but was hard to pick apart (and it isn't a lot of code).

If it's not too hard to review for you, I think I'd like to keep these together as is.

src/es5processor.ts
```diff
@@ -209,10 +217,11 @@ class ES5Processor extends Rewriter {
   emitGoogRequire(varName: string|null, tsImport: string): string {
     let modName: string;
     let isNamespaceImport = false;
-    if (tsImport.match(/^goog:/)) {
+    const nsImport = extractGoogNamespaceImport(tsImport);
+    if (nsImport !== null) {
       // This is a namespace import, of the form "goog:foo.bar".
       // Fix it to just "foo.bar", and save the variable name.
```
evmar: It looks like this doesn't actually save the variable name, but maybe for some other CL

test/e2e_test_closure_input_Module.js
evmar: Not done?

test/tsickle_test.ts
evmar: not fixed

test_files/import_from_goog/decls.d.ts
```diff
@@ -0,0 +1,11 @@
+// Matches test/e2e_test_closure_input_*.js
```
evmar: fix comment here after you move this

mprobst:
Aha, classic. Worked on my laptop, pushed, then later forgot to fetch, rebased and force pushed from my work PC. Found the commits on my laptop :-)

src/es5processor.ts
mprobst: Done.

PR #390, opened 2017-02-21
Replace inline source maps before goog.module rewriting.

LucasSloan:
Given empty input files and a directive to produce inline source maps, tsc produces output containing only an inline source map.  Since we don't create any new lines when rewriting imports, we have to replace the source map before it gets concatenated with a goog.module() statement.

### DIFF
--- a/test/e2e_source_map_test.ts
+++ b/test/e2e_source_map_test.ts
@@ -14,7 +14,7 @@ import * as ts from 'typescript';
 import * as cliSupport from '../src/cli_support';
 import {ANNOTATION_SUPPORT_CODE, convertDecorators} from '../src/decorator-annotator';
 import {Settings, toClosureJS} from '../src/main';
-import {extractInlineSourceMap, getInlineSourceMapCount, setInlineSourceMap,} from '../src/source_map_utils';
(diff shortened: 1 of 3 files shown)


evmar:
Can you explain more what you mean by "Since we don't create any new lines when rewriting imports, we have to replace the source map before it gets concatenated with a goog.module() statement."
Why are new lines important?

test/e2e_source_map_test.ts
```diff
@@ -350,6 +361,23 @@ function compile(sources: Map<string, string>, partialOptions = {} as Partial<Co
   return {compiledJS, dts, sourceMap};
 }
 
+function extractInlineSourceMap(source: string): string {
+  const inlineSourceMapRegex =
+      new RegExp('//# sourceMappingURL=data:application/json;base64,(.*)$', 'mg');
+  let previousResult: RegExpExecArray|null = null;
+  let result: RegExpExecArray|null = null;
+  // We want to extract the last source map in the source file
+  // since that's probably the most recent one added.  We keep
```
evmar: Can we rely on it just being the last line of the file?

LucasSloan: I don't think so.

PR #391, opened 2017-02-21
fix emitted externs namespace

evmar:
f99e086 translates "declare module 'foo'" into externs declarations,
but it generates a bad variable name if the declared module has a
hyphen in it.  Fix this by whitelisting valid module characters.

Fixes issue #393.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1075,7 +1075,7 @@ class ExternsWriter extends ClosureRewriter {
       case ts.SyntaxKind.ModuleDeclaration:
         let decl = <ts.ModuleDeclaration>node;
         switch (decl.name.kind) {
-          case ts.SyntaxKind.Identifier: {
+          case ts.SyntaxKind.Identifier:
             // E.g. "declare namespace foo {"
             let name = getIdentifierText(decl.name as ts.Identifier);
             if (name === undefined) break;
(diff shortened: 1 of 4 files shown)


src/tsickle.ts
```diff
@@ -1084,20 +1084,23 @@ class ExternsWriter extends ClosureRewriter {
...
             // module-scoped externs.  For now, ignore the enclosing
             // namespace (because this is declaring a top-level module)
             // and emit into a fake namespace.
-            namespace = ['tsickle_declare_module'];
-            let name = (decl.name as ts.StringLiteral).text;
+            namespace
+            = ['tsickle_declare_module'];
+            let importName = (decl.name as ts.StringLiteral).text;
+            // Imports are quoted strings, which allow characters like "-".
+            importName = importName.replace(/-/, '_');
```
rkirov: There are a lot of other cases to cover here '+', '.', etc. I am ok with doing one off as they appear, but maybe add the most common ones - '.' is very likely.

evmar: This is in `declare module "foo"`, so I think + isn't common?  But I bet @ is, ugh.

evmar:
PTAL

src/tsickle.ts
```diff
@@ -1084,20 +1084,32 @@ class ExternsWriter extends ClosureRewriter {
...
+            this.writeExternsVariable('tsickle_declare_module', [], '{}');
             namespace = ['tsickle_declare_module'];
-            let name = (decl.name as ts.StringLiteral).text;
+
+            // Declare the inner "tsickle_declare_module.foo".
+            let importName = (decl.name as ts.StringLiteral).text;
+            this.emit(`// Derived from: declare module "${importName}"\n`);
+            // We also don't care about the actual name of the module ("foo"
+            // in the above example), except that we want it to not conflict.
+            importName = importName.replace(/[^A-Za-z]/g, '_');
```
rkirov: you can keep '$' but I doubt that would be useful.

PR #392, opened 2017-02-21
Fix error when handling input with some files with preexisting source maps, and some without.

LucasSloan:
When composing source maps, we assumed that if any of the source files had a source map for a given transform, they all had them.  While this is correct for decorator downleveling and closurization, we don't have any guarantees for inline source maps in the source files, thus we need to guard against the source map not existing.

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -1,5 +1,5 @@
 import * as path from 'path';
-import {SourceMapGenerator} from 'source-map';
+import {SourceMapConsumer, SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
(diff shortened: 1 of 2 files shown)


src/tsickle_compiler_host.ts
```diff
@@ -221,14 +221,14 @@ export class TsickleCompilerHost implements ts.CompilerHost {
       }
     }
     if (this.preexistingSourceMaps.size > 0) {
-      // TODO(lucassloan): remove when the .d.ts has the correct types
```
evmar: Isn't this TODO still correct?

LucasSloan: Why so it is!  Fixed.

PR #398, opened 2017-02-22
Improve TsickleCompilerHost's implementation of ts.CompilerHost.

LucasSloan:
1. Delegate all methods in the ts.CompilerHost interface, including those that are optional.
2. Handle getSourceFile() returning undefined.

### DIFF
--- a/src/tsickle_compiler_host.ts
+++ b/src/tsickle_compiler_host.ts
@@ -105,7 +105,35 @@ export class TsickleCompilerHost implements ts.CompilerHost {
 
   constructor(
       private delegate: ts.CompilerHost, private tscOptions: ts.CompilerOptions,
-      private options: Options, private environment: TsickleHost) {}
+      private options: Options, private environment: TsickleHost) {
+    // ts.CompilerHost includes a bunch of optional methods.  If they're
+    // present on the delegate host, we want to delegate them.
+    if (this.delegate.getCancellationToken) {
+      this.getCancellationToken = this.delegate.getCancellationToken!.bind(this.delegate);
+    }
+    if (this.delegate.getDefaultLibLocation) {
+      this.getDefaultLibLocation = this.delegate.getDefaultLibLocation!.bind(this.delegate);
+    }
+    if (this.delegate.resolveModuleNames) {
+      this.resolveModuleNames = this.delegate.resolveModuleNames!.bind(this.delegate);
+    }
+    if (this.delegate.resolveTypeReferenceDirectives) {
+      this.resolveTypeReferenceDirectives =
+          this.delegate.resolveTypeReferenceDirectives!.bind(this.delegate);
+    }
+    if (this.delegate.getEnvironmentVariable) {
+      this.getEnvironmentVariable = this.delegate.getEnvironmentVariable!.bind(this.delegate);
+    }
+    if (this.delegate.trace) {
+      this.trace = this.delegate.trace!.bind(this.delegate);
+    }
+    if (this.delegate.directoryExists) {
+      this.directoryExists = this.delegate.directoryExists!.bind(this.delegate);
+    }
+    if (this.delegate.realpath) {
+      this.delegate.realpath = this.delegate.realpath!.bind(this.delegate);
+    }
+  }
 
   /**
    * Tsickle can perform 2 kinds of precompilation source transforms - decorator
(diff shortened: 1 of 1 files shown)


src/tsickle_compiler_host.ts
```diff
@@ -169,7 +170,7 @@ export class TsickleCompilerHost implements ts.CompilerHost {
   }
 
   stripAndStoreExistingSourceMap(sourceFile: ts.SourceFile): ts.SourceFile {
```
evmar: If sourceFile is possibly undefined here, write ts.SourceFile|undefined ?

src/tsickle_compiler_host.ts
```diff
@@ -124,7 +124,8 @@ export class TsickleCompilerHost implements ts.CompilerHost {
       fileName: string, languageVersion: ts.ScriptTarget,
       onError?: (message: string) => void): ts.SourceFile {
     if (this.runConfiguration === undefined || this.runConfiguration.pass === Pass.NONE) {
-      const sourceFile = this.delegate.getSourceFile(fileName, languageVersion, onError);
+      const sourceFile: ts.SourceFile =
```
evmar: Why the annotation here?

src/tsickle_compiler_host.ts
```diff
@@ -318,36 +319,20 @@ export class TsickleCompilerHost implements ts.CompilerHost {
...
-  }
-
-  getDefaultLibFileName(options: ts.CompilerOptions): string {
-    return this.delegate.getDefaultLibFileName(options);
-  }
-
-  getCanonicalFileName(fileName: string): string {
-    return this.delegate.getCanonicalFileName(fileName);
-  }
+  fileExists = this.delegate.fileExists;
```
evmar: It would be neat if this works!  But I'm not sure -- does this.delegate get set up first, or do these run first?

LucasSloan: I've tested it and it seems to work fine.

evmar: I tried it in the playground and it does the right thing!

src/tsickle_compiler_host.ts
LucasSloan: Leftover from when I was trying to annotate as ts.SourceFile | undefined.  Removed.

src/tsickle_compiler_host.ts
LucasSloan: We can't, because then the return type has to be ts.SourceFile|undefined and that has bad effects higher up.  I've added a comment explaining the weirdness.

evmar:
Here's an alternative idea:

```
class TsickleHost { // does *not* implement compiler host
  functionsAsBefore() {}
  asCompilerHost(): ts.CompilerHost {
    return {
      // we implement some functions
      getSourceFile: (...) => this.getSourceFile(...),
      // others we forward to delegate
      otherRandomFunc: this.delegate.otherRandomFunc,
  }
}
```

LucasSloan:
I think that has the same problems as the old approach that the this isn't bound to the right place.  And fixing that puts us right back here.

evmar:
What do you mean by "bound to the right place"?

evmar:
Ah, in this one, right.
```
      otherRandomFunc: this.delegate.otherRandomFunc,
```

LucasSloan:
Yep.

evmar:
I guess it's maybe a case for using .bind?
```
otherRandomFunc: this.delegate.otherRandomFunc.bind(this.delegate),
```

The nice thing about this pattern is you don't have to write out the full signature of each method.

evmar:
Oh wait, that is still wrong, because .bind() can't be called on undefined.

src/tsickle_compiler_host.ts
```diff
@@ -105,7 +105,37 @@ export class TsickleCompilerHost implements ts.CompilerHost {
...
+    // present on the delegate host, we want to delegate them.
+    if (this.delegate.getCancellationToken) {
+      this.getCancellationToken = () => this.delegate.getCancellationToken!();
+    }
+    if (this.delegate.getDefaultLibLocation) {
+      this.getDefaultLibLocation = () => this.delegate.getDefaultLibLocation!();
+    }
+    if (this.delegate.resolveModuleNames) {
+      this.resolveModuleNames = (moduleNames, containingFile) =>
+          this.delegate.resolveModuleNames!(moduleNames, containingFile);
```
evmar: I think in each of these you can write it like

this.resolveModuleNames = this.delegate.resolveModuleNames!.bind(this.delegate);

that will let you omit all the parameter passing around.

LucasSloan: Good suggestion - done!

PR #404, opened 2017-02-27
default to untyped mode in CLI

### LINKED ISSUE #394: Make `--untyped` the default
We should put the typed mode behind a flag until we're sure it works.

evmar:
We want to eliminate untyped mode, but until we do, the open source
tsickle behavior should match the behavior we're using in production.

Fixes #394.

### DIFF
--- a/test/e2e_main_test.ts
+++ b/test/e2e_main_test.ts
@@ -9,7 +9,7 @@
 import {assert, expect} from 'chai';
 import * as ts from 'typescript';
 
-import {Settings, toClosureJS} from '../src/main';
+import {toClosureJS} from '../src/main';
 
 describe('toClosureJS', () => {
   it('creates externs, adds type comments and rewrites imports', function() {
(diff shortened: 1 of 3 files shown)


test/e2e_main_test.ts
```diff
@@ -18,7 +18,7 @@ describe('toClosureJS', () => {
     const closure = toClosureJS(
         {sourceMap: true, experimentalDecorators: true} as ts.CompilerOptions,
         ['test_files/underscore/export_underscore.ts', 'test_files/underscore/underscore.ts'],
-        {isUntyped: false} as Settings, diagnostics);
```
LucasSloan: I think the unnecessary as was me - is that something we could get the linter to complain about?

evmar: I don't think so.  Sometimes using "as" is legit.  I blame your reviewer (probably me). :P 

I wish they had made the syntax for coercions a bit heavier so it'd stand out more.  Like in C++ you have to write `reinterpret_cast<Settings>(...)` and it's more obvious at review time that it's not what you want except in exceptional scenarios.

PR #405, opened 2017-02-28
Ban more redundant Closure tags.

mprobst:
@enum, @typedef, @const etc.
    
Also improves the error message to be a bit more actionable.

This is based on going through https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler.

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -113,14 +113,35 @@ const JSDOC_TAGS_WHITELIST = [
 ];
 
 /**
- * A list of JSDoc @tags that are never allowed in TypeScript source.
- * These are Closure tags that can be expressed in the TypeScript surface
- * syntax.
+ * A list of JSDoc @tags that are never allowed in TypeScript source. These are Closure tags that
+ * can be expressed in the TypeScript surface syntax. As tsickle's emit will mangle type names,
+ * these will cause Closure Compiler issues and should not be used.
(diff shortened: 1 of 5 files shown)


src/jsdoc.ts
```diff
@@ -150,7 +171,11 @@ export function parse(comment: string): {tags: Tag[], warnings?: string[]}|null
         warnings.push(`@${tagName} annotations are redundant with TypeScript equivalents`);
         continue;  // Drop the tag so Closure won't process it.
       } else if (arrayIncludes(JSDOC_TAGS_WITH_TYPES, tagName) && text[0] === '{') {
-        warnings.push('type annotations (using {...}) are redundant with TypeScript types');
+        warnings.push(`the type annotation on @${tagName
+                      } is redundant with its TypeScript type, remove the {...} part`);
+                      continue;
```
evmar: That clang-format mangling ewww, especially the 'continue'.
Maybe break this into multiple strings or something?

mprobst: This is fixed in the newer `clang-format`, just approve that and I'll fixup the code here.

evmar: Approve what?

mprobst: https://github.com/angular/tsickle/pull/414

PR #414, opened 2017-02-28
Upgrade clang-format to 1.0.48

mprobst:


mprobst:
@evmar I did, that's the source of the changes in `yarn.lock`. github doesn't display them because the file is too large.

Issue #410, opened 2017-02-28
class implements (abstract) class leads to @extends annotation


mprobst:
```ts
abstract class ComponentFactoryResolver {
...
}

class CodegenComponentFactoryResolver implements ComponentFactoryResolver {
...
}
```

Yields:

```js
/**
 * @abstract
 */
class ComponentFactoryResolver {
...
}
/**
 * @extends {ComponentFactoryResolver}
 */
class CodegenComponentFactoryResolver {
...
}
```

mprobst:
Maybe `@extends` is right, but then it should be a `class X extends Y`, not an `@extends`.

evmar:
Hm, what happens if you "implements" *two* superclasses?  We can't @extends both (I tried, the second one produces "WARNING - Bad type annotation. type annotation incompatible with other annotations. See https://github.com/google/closure-compiler/wiki/Bad-Type-Annotation for more information.").  I guess this is an argument for making everything structural or whatever.  Maybe I'll just reject that for now.

mprobst:
`@extends` is wrong, the complete fix is probably to generate an interface for each class that can be used when it's used in a type location:

```ts
class Foo {}
class Bar implements Foo {}
```

```js
/** @record */
function Foo_Interface() {}
/** @constructor @implements {Foo_Interface} */
function Foo() {}
/** @record @implements {Foo_Interface}*/
function Bar_Interface() {}
/** @constructor @implements {Bar_Interface} */
function Bar() {}
```

... and then use `Foo_Interface` and `Bar_Interface` whenever they are referred to in a type position. That also fixes the structural vs nominal typing aspect of classes in TS vs Closure.

evmar:
Thinking about this more: I think that change (introducing the interface) can preserve the semantics correctly, but it adds a ton of code everywhere (each class gets all its declarations doubled) for a very corner case feature (using "implements" against a class, which doesn't even work if the class has any 'private' members).

In particular I worry that we ship the generated tsickle JS as part of the open source Angular bundle (which is the reason we preserve JSDoc comments, for example).  And then users of open source Angular might not have a minifier that can remove these.

Is it possible for us get a 99% fix without uglying up all the code?  Perhaps this falls into the "not supported well by Closure" code pattern bucket that we should reject?  It feels like this goog.log bundle thing a bit.

mprobst:
It'd also fix structural typing issues.

Evan Martin <notifications@github.com> schrieb am Mi., 1. März 2017, 20:43:

[...]

PR #415, opened 2017-02-28
add abstract classes to the "class" test

evmar:
In preparation for #410, add some more test cases and comments to the "class" test.

No functionality change, just comments and renames.

### DIFF
--- a/test_files/class/class.ts
+++ b/test_files/class/class.ts
@@ -32,12 +32,15 @@ abstract class AbstractClass {
 interface InterfaceExtendsInterface extends Interface {
   interfaceFunc2(): void;
 }
+let interfaceExtendsInterfaceValue = {} as InterfaceExtendsInterface;
 interface InterfaceExtendsClass extends Class {
   interfaceFunc2(): void;
 }
+let interfaceExtendsClassValue: InterfaceExtendsClass = {} as InterfaceExtendsClass;
 interface InterfaceExtendsAbstractClass extends AbstractClass {
   interfaceFunc2(): void;
 }
+let interfaceExtendsAbstractClassValue: InterfaceExtendsAbstractClass = {} as any;
 
 // 3) class implements.
 class ClassImplementsInterface implements Interface {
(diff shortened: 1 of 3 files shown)


test_files/class/class.ts
```diff
@@ -1,24 +1,52 @@
...
+// This tests "implements" against an abstract class.
+// Note: in Closure you cannot currently "implements" two classes, see #410.
+class ImplementsAbstract implements AbstractClass {
+  abstractFunc(): void {}
+  // Note: because this class *implements* AbstractClass, it must also implement
+  // nonAbstractFunc despite that already having an implementation.
+  nonAbstractFunc(): void {}
+}
+
+// This tests "extends" against the class.
```
mprobst: ... and implements?

test_files/class/class.ts
```diff
@@ -1,24 +1,52 @@
...
+class Extends extends Class implements Interface {
   interfaceFunc(): void {}
 }
 
+// This tests "extends" against the abstract class.
+class ExtendsAbstract extends AbstractClass {
```
mprobst: should these be `ClassExtendsAbstract`, so we can also have `InterfaceExtendsInterface` vs `InterfaceImplementsInterface`? That is, put the Cartesian product into the type names to describe the case it's exercising?

evmar: I didn't realize you could "extends" an interface!  I will fix.

evmar:
PTAL (just review the .ts test)

test_files/class/class.ts
```diff
@@ -1,38 +1,107 @@
...
+// 1) interface implements
+// 2) interface extends
+// 3) class implements
+// 4) class extends
 
-class Implements implements Interface, Super {
-  interfaceFunc(): void {}
-  superFunc(): void {}
+// 1) interface implements.
+// No examples; this is not legal TypeScript.
```
mprobst: I don't understand what this comment refers to. Should this have a number from above?

test_files/class/class.ts
```diff
@@ -1,38 +1,107 @@
...
+abstract class AbstractClass {
+  abstract abstractFunc(): void;
+  nonAbstractFunc(): void { }
 }
 
+// Write out all permutations:
+// 1) interface implements
+// 2) interface extends
+// 3) class implements
+// 4) class extends
```
mprobst: does it make any difference whether the class extending/implementing is abstract? If not, document?

test_files/class/class.ts
evmar: Maybe it's not worth numbering these.  I will remove.

test_files/class/class.ts
evmar: It probably can matter.  I'll add some more sections.

evmar:
PTAL

PR #418, opened 2017-03-07
make readme say more about what tsickle does

evmar:
We're no longer just a type-annotator, we do a lot of transformation!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 # Tsickle - TypeScript to Closure Translator [![Build Status](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle)
 
-Tsickle converts TypeScript code into a form acceptible to the [Closure Compiler].
-This allows using TypeScript to transpile your sources, and then using
-Closure Compiler to bundle and optimize them, while taking advantage of type
-information in Closure Compiler.
+Tsickle converts TypeScript code into a form acceptable to the [Closure
+Compiler].  This allows using TypeScript to transpile your sources, and then
+using Closure Compiler to bundle and optimize them, while taking advantage of
+type information in Closure Compiler.
 
 [Closure Compiler]: https://github.com/google/closure-compiler/
 
@@ -12,7 +12,7 @@ information in Closure Compiler.
 A (non-exhaustive) list of the sorts of transformations Tsickle applies:
 
 - inserts Closure-compatible JSDoc annotations on functions/classes/etc
-- converts ES6 modules into `goog.module`/`goog.provide`
+- converts ES6 modules into `goog.module` modules
 - generates externs.js from TypeScript d.ts (and `declare`, see below)
 - declares types for class member variables
 - translates `export * from ...` into a form Closure accepts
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,12 +1,27 @@
-# Tsickle - TypeScript to Closure Annotator [![Build Status](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle)
+# Tsickle - TypeScript to Closure Translator [![Build Status](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle)
 
-Tsickle processes TypeScript and adds [Closure Compiler]-compatible JSDoc
-annotations. This allows using TypeScript to transpile your sources, and then
+Tsickle converts TypeScript code into a form acceptible to the [Closure Compiler].
```
kylecordes: acceptible -> acceptable

README.md
```diff
@@ -1,12 +1,27 @@
...
 information in Closure Compiler.
 
 [Closure Compiler]: https://github.com/google/closure-compiler/
 
+## What conversion means
+
+A (non-exhaustive) list of the sorts of transformations Tsickle applies:
+
+- inserts Closure-compatible JSDoc annotations on functions/classes/etc
+- converts ES6 modules into `goog.module`/`goog.provide`
```
mprobst: nit: / `goog.forwardDeclare`

kylecordes: How about mentioning here, that it performs this translation by means of common JS modules, hence it requires module:commonjs in the TSConfig?  I am working on a change to stop within explanatory error if that configuration is missing or incorrect.

README.md
evmar: Thanks!

README.md
evmar: I'm trying to stay high-level here so I will drop the goog.provide part

evmar: I'm trying to stay high-level here (and I think we can just fix this bug anyway)

kylecordes: @evmar For the sake of people coming in from outside, I think that sentence mentioning goog.* is fantastic. It is a critical thing to understand about what tsickle does and does not do.  If I were to give somebody a super concise high-level description of tsickle, I would probably say:

Tsickle extends the TypeScript compiler to emit JavaScript code containing Closure-Compiler-compatible JSDoc-like comment annotations translated from the original types. It  can emit either ES5 or ES2016 JavaScript output, but only in Closure "goog.module" modules, not in ES2016 modules.

evmar: Part of the point of tsickle is for users to need to know about goog.module modules. :)  (In principle if Closure Compiler improved its ES6 support we may not even do this transformation.)

The point of this wall of text is instead just to say "tsickle does some invasive code transformations, it doesn't just insert jsdoc" (which is what outsiders often think when they're not familiar with the guts).

kylecordes: So far in my experience (keeping in mind that I'm probably doing many things wrong, but in my defense there is not an abundance of documentation...) I have needed to be aware of how these module systems interoperate in Closure. By the way the documentation on that last bit is not particularly encouraging:

https://github.com/google/closure-compiler/wiki/JS-Modules#module-interoperation

However I am completely thrilled that making it moot is a goal - In this readme is a great improvement regardless of my nitpick :-)

PR #419, opened 2017-03-07
Add module: commonjs preflight check

kylecordes:
There did not appear to be any test coverage for the outer "main" CLI driver, I could add that if necessary but this change seemed straightforward and safe without it.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -218,8 +218,8 @@ function main(args: string[]): number {
     // This is not an upstream TypeScript diagnostic, therefore it does not go
     // through the diagnostics array mechanism.
     console.error(
-        'tsickle converts TypeScript modules to Closure modules, via CommonJS internally. Set tsconfig.js "module": "commonjs"');
-    return 2;
+        'tsickle converts TypeScript modules to Closure modules via CommonJS internally. Set tsconfig.js "module": "commonjs"');
+    return 1;
   }
 
   // Run tsickle+TSC to convert inputs to Closure JS files.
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -214,6 +214,14 @@ function main(args: string[]): number {
     return 1;
   }
 
+  if (config.options.module !== ts.ModuleKind.CommonJS) {
+    // This is not an upstream TypeScript diagnostic, therefore it does not go
+    // through the diagnostics array mechanism.
+    console.error(
+        'tsickle converts TypeScript modules to Closure modules, via CommonJS internally. Set tsconfig.js "module": "commonjs"');
```
mprobst: nit: drop the `,` in `, via CommonJS`.

src/main.ts
```diff
@@ -214,6 +214,14 @@ function main(args: string[]): number {
...
+    return 2;
```
mprobst: nit: as long as we don't have any other program specifically check for this return value, I'd just always return `1`, less confusing that way.

src/main.ts
kylecordes: Sure, commit coming in a moment.

src/main.ts
kylecordes: Also done, see commit. I have not the ground the source code of very many Google projects, there seems to be disagreement among experts as to whether it is best to consolidate to a single error exit return value, or to have as many different values as possible for different things that went wrong. I'm equally happy with either, I have found that scripts rarely care why command failed.

PR #422, opened 2017-03-07
Filter implemented interfaces through the type blacklist.

### LINKED ISSUE #421: @implements ignores the type blacklist
Because it doesn't use `type-translator.ts`. That's a bit of a code smell, but probably still better than trying to use it and ending up with something like `@implements {?}`. But maybe we should still centralize the code doing type emits into `type-translator.ts`.

mprobst:
Fixes #421.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -328,10 +328,15 @@ class ClosureRewriter extends Rewriter {
     if (!type) {
       type = typeChecker.getTypeAtLocation(context);
     }
-    let translator = new typeTranslator.TypeTranslator(
-        typeChecker, context, this.options.typeBlackListPaths, this.symbolsToAliasedNames);
+    return this.newTypeTranslator(context).translate(type);
+  }
+
+  newTypeTranslator(context: ts.Node) {
+    const translator = new typeTranslator.TypeTranslator(
+        this.program.getTypeChecker(), context, this.options.typeBlackListPaths,
+        this.symbolsToAliasedNames);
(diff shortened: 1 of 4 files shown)


test_files/jsdoc_types/jsdoc_types.ts
```diff
@@ -31,3 +31,7 @@ let useDefaultClassAsType: DefaultClass;
 // NeverTyped should be {?}, even in typed mode.
 let useNeverTyped: NeverTyped;
 let useNeverTyped2: string|NeverTyped;
+/** Note: no implements JSDoc clause because the type is blacklisted. */
+class ExtendsNeverTyped implements NeverTyped {
```
evmar: I guess ImplementsNeverTyped, not Extends.

src/tsickle.ts
```diff
@@ -920,6 +925,9 @@ class Annotator extends ClosureRewriter {
             if (sym.flags & ts.SymbolFlags.Alias) {
               sym = typeChecker.getAliasedSymbol(sym);
             }
+            if (this.newTypeTranslator(heritage).isBlackListed(sym)) {
```
evmar: Is it important that this use a different context parameter than all the other uses of typetranslator?

mprobst: Nope, I slipped in finding the closest context node (not that it matters). Fixed.

test_files/jsdoc_types/jsdoc_types.ts
mprobst: Done.

PR #424, opened 2017-03-09
Improve type generation for interfaces

mprobst:
Support interface extension.
Support generic type arguments (@template tags).

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -166,18 +166,19 @@ class ClosureRewriter extends Rewriter {
     // then paramTags[0] = [info about x, info about y].
     const paramTags: jsdoc.Tag[][] = [];
     const returnTags: jsdoc.Tag[] = [];
+    const typeParameterNames = new Set<string>();
 
     for (let fnDecl of fnDecls) {
       // Construct the JSDoc comment by reading the existing JSDoc, if
       // any, and merging it with the known types of the function
       // parameters and return type.
-      let jsDoc = this.getJSDoc(fnDecl) || [];
+      let docTags = this.getJSDoc(fnDecl) || [];
 
       // Copy all the tags other than @param/@return into the new
       // JSDoc without any change; @param/@return are handled specially.
       // TODO: there may be problems if an annotation doesn't apply to all overloads;
       // is it worth checking for this and erroring?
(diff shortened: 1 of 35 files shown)


src/tsickle.ts
```diff
@@ -187,6 +188,14 @@ class ClosureRewriter extends Rewriter {
         newDoc.push({tagName: 'abstract'});
       }
 
+      // Add any @template tags.
+      // This will break assuming multiple declarations with conflicting
+      // template variable names.
```
evmar: I don't get this comment.

src/tsickle.ts
```diff
@@ -166,18 +166,19 @@ class ClosureRewriter extends Rewriter {
     // then paramTags[0] = [info about x, info about y].
     const paramTags: jsdoc.Tag[][] = [];
     const returnTags: jsdoc.Tag[] = [];
+    const typeParameterNames = new Set<string>();
```
evmar: Why a set and not an array here?

src/type-translator.ts
```diff
@@ -231,8 +231,11 @@ export class TypeTranslator {
         return '?';
       case ts.TypeFlags.TypeParameter:
         // This is e.g. the T in a type like Foo<T>.
-        this.warn(`unhandled type flags: ${ts.TypeFlags[type.flags]}`);
-        return '?';
+        if (!type.symbol) {
+          this.warn(`TypeParameter without a symbol`);
```
evmar: Do you know when this can occur?  Can you leave a comment here about it?

src/tsickle.ts
mprobst: Imagine you have three overloads of your function, they might be using the same template type parameter multiple times. I've added a test for that.

src/type-translator.ts
mprobst: I think it cannot happen. Added a comment.

src/tsickle.ts
mprobst: I didn't really get it either. It turns out it's not a problem (I think). I was wondering what happens if you have multiple overloads that use the same template arg name (eg. `T`) in different locations and meanings. But I think it's fine because we just make a huge union type mess, so it should all work out.

PR #425, opened 2017-03-10
factor out the shared header of externs generation

evmar:
Rather than having annotate() return externs with a header (which means multiple
calls to annotate() can't be concatenated), instead expose an EXTERNS_HEADER
constant.

Use this in the tests (causing no test change).

Also fix the tests so that if a test *did* have multiple externs files, it wouldn't cause
a duplicated header.


src/tsickle.ts
```diff
@@ -34,6 +34,18 @@ export interface Output {
 }
 
 /**
+ * The header to be used in generated externs.  This is not included in the
+ * output of annotate() because annotate() works one file at a time, and
+ * typically you create one externs file from the entire compilation.
```
mprobst: nit: entire compilation unit, as in one bazel target, right?

src/tsickle_compiler_host.ts
```diff
@@ -6,7 +6,8 @@ import {convertDecorators} from './decorator-annotator';
 import {processES5} from './es5processor';
 import {ModulesManifest} from './modules_manifest';
 import * as sourceMapUtils from './source_map_utils';
-import {annotate, isDtsFileName} from './tsickle';
+import * as tsickle from './tsickle';
+import {isDtsFileName} from './tsickle';
```
mprobst: nit: use `tsickle.isDtsFileName` for consistency?

evmar: Yeah I struggled on this one.  I figured isDtsFileName itself is a pretty specific name (it doesn't even really belong in the tsickle module maybe?) but I guess it's trivial to change it later so I'll fix it.

mprobst: Ack, makes sense.

PR #426, opened 2017-03-13
Strip Clutz' look of disapproval namespace (ಠ_ಠ.clutz)

mprobst:
Most code seen by Tsickle will only ever see local aliases, Clutz
symbols can be written by users directly in code, and they can appear by
dereferencing TypeAliases. The code below simply strips the prefix, the
remaining type name then matches Closure's type.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -187,6 +187,12 @@ export class TypeTranslator {
       reportInaccessibleThisError: doNothing,
     };
     builder.buildSymbolDisplay(sym, writer, this.node);
+    // Clutz (https://github.com/angular/clutz) emits global type symbols hidden in a special
+    // ಠ_ಠ.clutz namespace. While most code seen by Tsickle will only ever see local aliases, Clutz
+    // symbols can be written by users directly in code, and they can appear by dereferencing
+    // TypeAliases. The code below simply strips the prefix, the remaining type name then matches
+    // Closure's type.
+    if (str.startsWith('ಠ_ಠ.clutz.')) str = str.substring('ಠ_ಠ.clutz.'.length);
     return str;
   }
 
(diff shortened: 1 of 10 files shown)


mprobst:
/CC @rkirov 

This will continue to work (or start being harmless) even if we remove the look of disapproval. Do you have a different plan on how to prevent people from seeing these types/getting auto complete on them in their IDE?

rkirov:
I would have only added, pointing straight to the Java class and constant in the comment, can same one of us some searching in the future. No big deal.

About removing LOD - it does what it is supposed to do, but has two big disadvantages - IDEs show it on hover (the module reexport is transparent) and 

A alternative line of thinking goes like this: 

There are two code structuring mechanisms - modules and namespaces. Both exists in Closure and TS. To TS users we expose all Closure as-if they were modules. Clutz meanwhile lays out all Closure as namespace (even goog.modules) and does 1:1 reexporting through modules. We are always making up something, either some modules get turned in namespaces or vice versa in Clutz.

What if we flip the lie - lay out everything as modules in the .d.ts (maybe except externs). The hard part is that we need to write out the imports:

```javascript
declare module 'goog:mods.A' {
  import * as B from 'goog:mods.B';
  class A {
    foo: B.B;
  }
}

declare module 'goog:mods.B' {
  class B {}
}
```

Any obvious flaws with this argument? If none, I will move this to an issue on clutz.

mprobst:
@rkirov I think your analysis is spot on, agreed. When we initially wrote Clutz, we shied away from using modules only, as our impression was it'd be too hard to import & rename all the symbols we need to emit (you also need to be mindful of scoping, e.g. when the same name `Foo` exists in multiple packages and the local scope). But we never tried it, it might actually run out to be cleaner and less hard than we expect.

PR #428, opened 2017-03-14
Retain @fileoverview comments preceding non-value imports.

mprobst:
TypeScript drops imports from the output if none of the imported symbols
are ever used as values (as opposed to types). In Closure, those
comments can have semantic meaning attached to them. Make sure the
comment gets retained by splitting it off the import, through a single
line break between the comment and the import.


src/tsickle.ts
```diff
@@ -880,6 +880,14 @@ class Annotator extends ClosureRewriter {
...
   private emitImportDeclaration(decl: ts.ImportDeclaration): boolean {
     this.writeRange(decl.getFullStart(), decl.getStart());
+    // This is a load-bearing line break. TypeScript only emits imports that provide symbols used in
+    // value positions (as opposed to type positions). When TypeScript drops an import, it drops the
+    // comment together with it. However in Closure, some comments have semantic meaning, and must
+    // be retained. In particular file level comments are associated with the first import, so they
+    // would be dropped by TS if the import is type only. The additional line break before the
+    // import separates a potential comment from the import and makes sure it gets emitted.
+    // See also the "fileoverview" test.
+    if (decl.getFullStart() !== decl.getStart()) this.emit('\n');
```
evmar: Is it worth checking that this extra trivia is non-whitespace?  It'd make the test output a bit cleaner (less random newlines) at the cost of a bit of predictability (but it's already random-feeling whether this newline shows up or not).

src/tsickle.ts
```diff
@@ -887,7 +887,11 @@ class Annotator extends ClosureRewriter {
     // would be dropped by TS if the import is type only. The additional line break before the
     // import separates a potential comment from the import and makes sure it gets emitted.
     // See also the "fileoverview" test.
-    if (decl.getFullStart() !== decl.getStart()) this.emit('\n');
+    if (decl.getFullStart() !== decl.getStart() &&
+        // Only emit if the trivia contain non-whitespace, i.e. a comment.
+        !this.file.text.slice(decl.getFullStart(), decl.getStart()).match(/^\s*$/)) {
```
evmar: alternatively, drop the ! and match(/\S/)

mprobst: Good point, that should also be faster (depending on VM opts, I guess).

PR #430, opened 2017-03-15
Support renaming exported typedefs.

### LINKED ISSUE #429: tsickle uses the target symbols name for TypeAliases
```ts
// file1:
export type Foo = number;
// file2:
export {Foo as Bar} from './file1';
```

Causes file2.js to contain `exports.Bar;`, but usage sites write `file2_forward_declare.Foo`.

mprobst:
Fixes #429.


src/tsickle.ts
```diff
@@ -609,8 +609,12 @@ class Annotator extends ClosureRewriter {
           this.emit(` {${exportedSymbols.map(e => unescapeName(e.name)).join(',')}}`);
         } else {
           if (exportDecl.exportClause) {
-            exportedSymbols =
-                exportDecl.exportClause.elements.map(e => typeChecker.getSymbolAtLocation(e.name));
+            exportedSymbols = exportDecl.exportClause.elements.map(e => {
+              return {
+                name: getIdentifierText(e.name),
```
evmar: Isn't this just sym.name?

src/tsickle.ts
```diff
@@ -815,7 +820,9 @@ class Annotator extends ClosureRewriter {
       this.generatedExports.add(name);
       reexports.add(sym);
     }
-    return toArray(reexports.keys());
+    return toArray(reexports.keys()).map(s => {
```
evmar: If you name this var 'sym' then the next line gets cuter.

src/tsickle.ts
```diff
@@ -919,19 +926,26 @@ class Annotator extends ClosureRewriter {
...
-        let symbols: ts.Symbol[] = [];
+        let symbols: {name: string, sym: ts.Symbol}[] = [];
         const typeChecker = this.program.getTypeChecker();
         if (importClause.name) {
           // import a from ...;
-          symbols = [typeChecker.getSymbolAtLocation(importClause.name)];
+          symbols = [{
+            name: getIdentifierText(importClause.name),
+            sym: typeChecker.getSymbolAtLocation(importClause.name)
+          }];
```
evmar: This code feels weirdly close to the earlier code, is it useful to refactor?

src/tsickle.ts
evmar: Ohh I see below it's about renaming exports.  Maybe dump a comment here about how/why sym.name and e.name will be different.

src/tsickle.ts
```diff
@@ -919,19 +926,26 @@ class Annotator extends ClosureRewriter {
...
       if (!this.options.untyped) {
-        let symbols: ts.Symbol[] = [];
+        let symbols: {name: string, sym: ts.Symbol}[] = [];
         const typeChecker = this.program.getTypeChecker();
         if (importClause.name) {
           // import a from ...;
-          symbols = [typeChecker.getSymbolAtLocation(importClause.name)];
+          symbols = [{
+            name: getIdentifierText(importClause.name),
+            sym: typeChecker.getSymbolAtLocation(importClause.name)
```
evmar: For renaming imports, I vaguely recall importClause has two members, one which is the name and one which is the rename.  Though maybe the symbol serves for one of those.

src/tsickle.ts
mprobst: I've added a comment and extracted a type.

src/tsickle.ts

src/tsickle.ts
mprobst: it has `propertyName?` and `name`. Confusingly, name is always the local name, and propertyName takes the role of the original name if present.

src/tsickle.ts
mprobst: I've extracted a method, it's useful together with the comment. The code overall is eerily similar between exports and imports, but different enough that it's annoying to fully unify the larger block (export * vs import * vs import Name).

PR #431, opened 2017-03-15
add gulp task to format sources

alexeagle:


alexeagle:
you know, I have this in the Angular repo but I don't think everyone knows about it.
I should change https://github.com/angular/gulp-clang-format/blob/master/index.js#L70 so it could print a custom message, then we'd just print "run gulp format"

PR #433, opened 2017-03-15
Support input source maps that say their file name is different from what tsc was told.

LucasSloan:
Motivating example is that ngc produces fully specified source map file names, but we compile them by their bare names.  (e.g. /foo/bar/baz.ngfactory.ts vs. baz.ngfactory.ts).

### DIFF
--- a/test/e2e_source_map_test.ts
+++ b/test/e2e_source_map_test.ts
@@ -242,6 +242,22 @@ describe('source maps', () => {
     expect(sourceMap.originalPositionFor({line, column}).line).to.equal(6, 'method position');
   });
 
+  it('handles input source maps with different file names than supplied to tsc', function() {
+    const sources = new Map<string, string>();
+    const inputSourceMap =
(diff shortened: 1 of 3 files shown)


test/e2e_source_map_test.ts
```diff
@@ -242,6 +242,26 @@ describe('source maps', () => {
...
 
+  it('handles input source maps with different file names than supplied to tsc', function() {
+    const sources = new Map<string, string>();
+    const inputSourceMap =
+        `{"version":3,"sources":["original.ts"],"names":[],"mappings":"AAAA,MAAM,EAAE,EAAE,CAAC","file":"foo/bar/intermediate.ts","sourceRoot":""}`;
+    sources.set(
+        'intermediate.ts',
+        `const x = 3;
+//# sourceMappingURL=data:application/json;base64,${
+                                                    Buffer.from(inputSourceMap, 'utf8')
```
alexeagle: this formatting looks wacky, can be fixed with extract variable

evmar:
Is it possible to fix the paths so that the input is already correct?  It would be nice if we didn't need to get so fancy here.

PR #435, opened 2017-03-16
Make tests runnable on windows

b-strauss:
Removes the custom host.directoryExists implementation in favor of an explicit empty types compiler option.
Normalizes both fileName and cachedLibPath to pass equality check on windows.

This PR only makes the tests loadable/runnable on windows. There are still tests that fail.

### DIFF
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -26,6 +26,9 @@ export const compilerOptions: ts.CompilerOptions = {
   jsx: ts.JsxEmit.React,
   // Flags below are needed to make sure source paths are correctly set on write calls.
   rootDir: path.resolve(process.cwd()),
+  // Disable searching for @types typings. This prevents TS from looking
+  // around for a node_modules directory.
+  types: [],
   outDir: '.',
   strictNullChecks: true,
 };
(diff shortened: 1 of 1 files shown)


test/test_support.ts
```diff
@@ -26,6 +26,9 @@ export const compilerOptions: ts.CompilerOptions = {
   jsx: ts.JsxEmit.React,
   // Flags below are needed to make sure source paths are correctly set on write calls.
   rootDir: path.resolve(process.cwd()),
+  // Disable searching for @types typings. This prevents TS from looking
```
evmar: Does the comment on line 27 need fixing now?

test/test_support.ts
```diff
@@ -34,26 +37,20 @@ const {cachedLibPath, cachedLib} = (function() {
...
-      let relName = path.relative(process.cwd(), dirName);
-      if (relName === 'node_modules/@types') return false;
-    }
-    return realDirectoryExists ? realDirectoryExists(dirName) : false;
-  };
-
   host.getSourceFile = function(
                            fileName: string, languageVersion: ts.ScriptTarget,
                            onError?: (msg: string) => void): ts.SourceFile {
+    fileName = path.normalize(fileName);
```
evmar: Why are these normalize() calls necessary?  Is it worth explaining in a comment?  (I'm not sure, what do you think?)

test/test_support.ts
b-strauss: I will move `types` up, so the comment can remain unchanged.

test/test_support.ts
b-strauss: Yes, `cachedLibPath` has mixed directory seperators on windows, e.g. `tsickle\node_modules\typescript\lib/lib.es6.d.ts`, and `fileName` has only unix like directory seperators `tsickle/node_modules/typescript/lib/lib.es6.d.ts`. That's why the equality check in the next line fails.

I think it's worth explaining and add a comment.

b-strauss:
cc @evmar

evmar:
Thanks!

orenmizr:
question, there is still that path bug in windows with _ instead of . but tests are passing?

b-strauss:
@orenmizr I guess you should open up a new issue. This is probably the wrong place for such a request.

orenmizr:
was just stating that for a wider usage of this package - it is best that window users will get this fix 435. can't open a issue for that. it does not stand by its own.

b-strauss:
cc @evmar

evmar:
@alexeagle does releases, but I think we shouldn't go to the effort of making a release for windows until we're sure windows works.

b-strauss:
Yeah, i think #446 should be done first.

PR #438, opened 2017-03-19
Fix line endings mismatch in tests and pathToModuleName on Windows

### LINKED ISSUE #400: pathToModuleName wrong on Windows
On #374:

```
the root cause is the regex in pathToModuleName function.
"build/testDep.".replace(///g, '.').replace(/^[^a-zA-Z$]/, '').replace(/[^a-zA-Z0-9.$]/g, '_');
```

We need to handle backslashes in paths on Windows.

### LINKED ISSUE #436: golden file tests fail with mismatching line endings on Windows
There are `\n`s all over the codebase. This leads to most of the failed golden tests on windows.

I will add an OS check and, if we are on windows, normalize the line endings.

b-strauss:
Fixes #436
Fixes #400

### DIFF
--- a/test/tsickle_test.ts
+++ b/test/tsickle_test.ts
@@ -62,6 +62,9 @@ function compareAgainstGolden(output: string|null, path: string) {
       }
     }
   } else {
+    // Make sure we have proper line endings when testing on Windows.
+    if (golden != null) golden = golden.replace(/\r\n/g, '\n');
+    if (output != null) output = output.replace(/\r\n/g, '\n');
     expect(output).to.equal(golden, path);
   }
 }
--- a/test/tsickle_test.ts
+++ b/test/tsickle_test.ts
@@ -48,6 +48,10 @@ function compareAgainstGolden(output: string|null, path: string) {
     }
   }
 
+  // Make sure we have proper line endings when testing on Windows.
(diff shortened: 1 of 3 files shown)


test/tsickle_test.ts
```diff
@@ -62,6 +62,9 @@ function compareAgainstGolden(output: string|null, path: string) {
       }
     }
   } else {
+    // Make sure we have proper line endings when testing on Windows.
+    if (golden != null) golden = golden.replace(/\r\n/g, '\n');
+    if (output != null) output = output.replace(/\r\n/g, '\n');
```
evmar: This seems like the wrong place for this.  What about the above check on line 51 for output !== golden?  I think this logic should move up to before line 51 perhaps.

b-strauss: Line 54 writes the file on disk. On Windows this would mean `\n` in written files ... but now that I think about it, `output !== golden` in line 51 is always false on Windows. I'll change it.

b-strauss:
cc @evmar 

This can be merged if you approve.

b-strauss:
@evmar hang on a minute I think I can also fix the remaining tests with the regexp.

b-strauss:
No, sorry. This can be merged. :D

PR #441, opened 2017-03-20
fix jsdoc parsing on Windows

### LINKED ISSUE #440: Tsickle fails to emit multi-line docs on Windows
We have four remaining tests on Windows that fail. Three of them fail because of multi-line comments:

- golden tests comments
- golden tests interface
- golden tests jsdoc

I suspect line endings again.

Example (screenshot for color format):

errors

b-strauss:
Fixes #440


evmar:
Thanks for debugging this!

PR #444, opened 2017-03-21
fix isBlackListed check on Windows

### LINKED ISSUE #443: test 'tsickle compiler host passes blacklisted paths' fails on Windows
Last remaining test that fails on Windows:

```
1) tsickle compiler host passes blacklisted paths:
   AssertionError: expected 'let /** @type {!Banned} */ b: Banned = {b: "a"};' to include '/** @type {?} */ b: Banned'
   at Context.<anonymous> (D:\workspace\b-strauss\test\e2e_tsickle_compiler_host_test.ts:83:23)
```

b-strauss:
Rearranged the code a bit, so the normalization is not done if the method returns early.

Fixes #443

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -511,13 +511,14 @@ export class TypeTranslator {
   /** @return true if sym should always have type {?}. */
   isBlackListed(symbol: ts.Symbol): boolean {
     if (this.pathBlackList === undefined) return false;
+    const pathBlackList = this.pathBlackList;
     if (symbol.declarations === undefined) {
       this.warn('symbol has no declarations');
       return true;
     }
     return symbol.declarations.every(n => {
       const fileName = path.normalize(n.getSourceFile().fileName);
-      return this.pathBlackList != null && this.pathBlackList.has(fileName);
+      return pathBlackList.has(fileName);
     });
   }
 }
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as nodePath from 'path';
 import * as ts from 'typescript';
 
 export function assertTypeChecked(sourceFile: ts.SourceFile) {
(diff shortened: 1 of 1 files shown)


src/type-translator.ts
```diff
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as nodePath from 'path';
```
mprobst: nit: use `pathlib` as the alias (we have other places doing that, too), or, preferably rename the local variables to avoid the collision?

src/type-translator.ts
```diff
@@ -503,13 +504,15 @@ export class TypeTranslator {
   /** @return true if sym should always have type {?}. */
   isBlackListed(symbol: ts.Symbol): boolean {
     if (this.pathBlackList === undefined) return false;
-    const pathBlackList = this.pathBlackList;
     if (symbol.declarations === undefined) {
       this.warn('symbol has no declarations');
       return true;
     }
+    // Normalize paths to not break checks on Windows.
+    const pathBlackList = new Set<string>(
```
mprobst: can we arrange the code such that we only do this once, e.g. in the constructor, instead of on every check? This method is called frequently, doing the work over and over again will likely be a performance issue.

b-strauss: Should I keep the auto assignment of `pathBlackList` and just overwrite the property like this in the constructor?

```typescript
this.pathBlackList = new Set<string>(
        Array.from(this.pathBlackList.values()).map(path => nodePath.normalize(path)));
```

src/type-translator.ts
b-strauss: I'll rename the variable to `fileName`.

src/type-translator.ts
mprobst: Yes, that sounds about right. Bonus points for making the property `readonly` :-)

b-strauss: Maybe we could change everything in the constructor to readonly? Seems like nothing gets overwritten later.

mprobst:
SGTM.

Benjamin Strauß <notifications@github.com> schrieb am Di., 21. März 2017 um
11:14 Uhr:

b-strauss:
I'll apply the changes later when I get home from work.

mprobst:
Thanks!

Benjamin Strauß <notifications@github.com> schrieb am Di., 21. März 2017 um
11:19 Uhr:

src/type-translator.ts
```diff
@@ -503,14 +511,13 @@ export class TypeTranslator {
   /** @return true if sym should always have type {?}. */
   isBlackListed(symbol: ts.Symbol): boolean {
     if (this.pathBlackList === undefined) return false;
-    const pathBlackList = this.pathBlackList;
```
evmar: The reason for this line was to save the non-null pathBlackList.

Issue #427, opened 2017-03-13
TS 2.2 causes static methods to get removed by the closure compiler


rubenlg:
File a.ts:
```typescript
export class A {
  static print() {
    console.log('A');
  }
}
```

File b.ts:
```typescript
import {A} from './a';
A.print();
```

File tsconfig.json:
```json
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es6"
    },
    "include": ["*.ts"]
}
```

Output of running tsickle + closure:
```javascript
$ tsickle 
$ closure-compiler -O ADVANCED *.js
var a={};Object.defineProperty(a,"__esModule",{value:!0});a.a=function(){};Object.defineProperty({},"__esModule",{value:!0});a.a.b();
```

That code fails because `a.a.b()` is not defined (`b` should be the static method `print`, which is gone).

rubenlg:
In case it is relevant, I installed tsickle earlier today with npm, and I'm using the closure compiler version v20170218.

The problem seems to be `Object.defineProperty(exports, "__esModule", { value: true });`. If I remove that line from the transpiled javascript, then the static method gets emitted by closure and the code runs as expected.

evmar:
This bug suggests this might be a breaking change in TS 2.2:
https://github.com/Microsoft/TypeScript/issues/14351

Can you try TS 2.1 and let me know if it works?  We haven't yet updated to 2.2.

rubenlg:
Thanks for the quick reply! I just tested TS 2.1 and it does work. I was using TS 2.2 since that's what you get when you do:
  npm install tsickle typescript

It might be worth a warning on the frontpage that this doesn't work yet with TS 2.2.

evmar:
Sorry about this.  You are apparently the first person to encounter this.  We typically update to the current TS version within a few weeks of the release so we don't call out which version we support anywhere.  Maybe we can specify this in our package.json somewhere?

mprobst:
Maybe there's some "ts.version" that we could check at runtime?

evmar:
I attempted to reproduce this in a tsickle test case via 

```
export default class Foo {
  static bar() { return 3; }
}
```

(using "default" because that triggers this `__esmodule` thing in ts2.1) and it compiled fine (newlines added for readability):

```
module$contents$test_files$export_default$default_Foo=function(){};
module$contents$test_files$export_default$default_Foo.bar=function(){return 3};
[...]

PR #447, opened 2017-03-21
Upgrade to TypeScript 2.2

alexeagle:
I created a ts2.1 branch on the angular/tsickle repo, which I propose we use as a release branch for any changes that need to land in a TS 2.1 dependency.

So this change would go on master now, and master becomes the ts2.2 branch.


src/type-translator.ts
```diff
@@ -111,13 +133,43 @@ export function symbolToDebugString(sym: ts.Symbol): string {
...
+    ts.SymbolFlags.EnumMemberExcludes,
+    ts.SymbolFlags.FunctionExcludes,
+    ts.SymbolFlags.ClassExcludes,
+    ts.SymbolFlags.InterfaceExcludes,
+    ts.SymbolFlags.RegularEnumExcludes,
+    ts.SymbolFlags.ConstEnumExcludes,
+    ts.SymbolFlags.ValueModuleExcludes,
+    ts.SymbolFlags.NamespaceModuleExcludes,
+    ts.SymbolFlags.MethodExcludes,
+    ts.SymbolFlags.GetAccessorExcludes,
```
evmar: This list is a bit subtle and maybe not worth keeping right, but it's trying to just be the unique values of the enum.  They have a bunch of stuff in the enum that is just the union of other values in the enum and it seems you've copied them here.

Basically everything up to this line but nothing after
https://github.com/Microsoft/TypeScript/blob/00abefbb683a9317c7f61b11ca7760a938ac199c/src/compiler/types.ts#L2725

evmar: [same in the earlier block too]

test_files/clutz.no_externs/strip_clutz_type.js
```diff
@@ -1,4 +1,5 @@
 goog.module('test_files.clutz.no_externs.strip_clutz_type');var module = module || {id: 'test_files/clutz.no_externs/strip_clutz_type.js'};
+Object.defineProperty(exports, "__esModule", { value: true });
```
evmar: I think this breaks Closure.
https://github.com/angular/tsickle/issues/427

test_files/comments/comments.tsickle.ts
```diff
@@ -14,13 +14,15 @@ export1: string;
...
   nodoc3: number;
-  /** inline jsdoc comment without type annotation */
-  jsdoc1: number;
-  /**
-   * multi-line jsdoc comment without
-   * type annotation.
-   */
-  jsdoc2: number;
+/**
+ * inline jsdoc comment without type annotation
```
evmar: I wonder what happened here?

src/type-translator.ts
```diff
@@ -37,13 +38,40 @@ export function typeToDebugString(type: ts.Type): string {
...
+    ts.TypeFlags.StringLike,
+    ts.TypeFlags.NumberLike,
+    ts.TypeFlags.BooleanLike,
+    ts.TypeFlags.EnumLike,
+    ts.TypeFlags.UnionOrIntersection,
+    ts.TypeFlags.StructuredType,
+    ts.TypeFlags.StructuredOrTypeVariable,
+    ts.TypeFlags.TypeVariable,
+    ts.TypeFlags.Narrowable,
+    ts.TypeFlags.NotUnionOrUnit,
```
rkirov: OOC, whats the clang-format logic here. I feel it keeps swapping back and forth.

package.json
```diff
@@ -43,7 +43,7 @@
     "mocha": "^3.2.0",
     "temp": "^0.8.1",
     "tslint": "^3.15.1",
-    "typescript": "2.1.5"
+    "typescript": "2.2"
```
rkirov: fix to 2.2.1

src/type-translator.ts
alexeagle: I believe if all the entries are short enough to fit within max/3 cols, it does three per line, otherwise it puts them one per line

package.json
alexeagle: done

test_files/comments/comments.tsickle.ts
alexeagle: Something in TS changed their comment parsing? Might be interesting to track down what the change was, I recall you had some requests for them to improve it

evmar: I remember it taking some effort to figure out the right order of regexes/parsing to get whitespace right in comments, but I think it's probably not worth worrying about here.

src/type-translator.ts
alexeagle: Done, added comment for the next maintainer.

test_files/clutz.no_externs/strip_clutz_type.js
alexeagle: Interesting, we did have some of those before, right? I'll verify this change through global presubmit before merging.

evmar: From the above link you can go to the TS bug where they say:
"The behavior was there for modules with export default. now it is for all modules."
So I think google3 was ok because we disallow export default.

test_files/comments/comments.tsickle.ts
alexeagle: okay, your call, I can dig in and find the delta if you like.

alexeagle:
Rebased, ready to merge PTAL

PR #450, opened 2017-03-22
Generate @export instead of externs for exporting decorators.

drjanitor:
Avoids compiler explosions when multiple files export symbols with the same name. Also, the closure compiler documentation really, really wants you to use exports instead of externs whenever possible.


drjanitor:
Done, and done.

evmar:
I guess this talks about this issue:
https://developers.google.com/closure/compiler/docs/api-tutorial3#no

test_files/exporting_decorator/exporting.ts
```diff
@@ -1,50 +1,47 @@
...
+function exportingDecorator() {
   return function(target: any, name: string|symbol) {}
 }
 
-function nonExportingFieldDecorator() {
+function nonExportingDecorator() {
   return function(target: any, name: string|symbol) {}
 }
 
-@exportingClassDecorator()
```
evmar: Why did you remove this?  Is it related?

drjanitor:
turns out you can't @export classes inside a goog.module, so I removed that
from tests.

evmar:
Does that mean you can't use this for classes?  Doesn't that break the use case?

drjanitor:
the use case I'm after is "do not remove this field from the class", which
is indeed working. I don't think I'm actually afraid of anyone
instantiating my class from a template or something else.

drjanitor:
Ping?

evmar:
Oops, sorry, meant to approve before.  I am a bit worried that this feature is very particular to your use case and not generally reusable by other users.

PR #454, opened 2017-03-28
strip __esModule properties from emitted output

### LINKED ISSUE #427: TS 2.2 causes static methods to get removed by the closure compiler

evmar:
When we convert to `goog.module`, we no longer want the 
  Object.defineProperty(exports, "__esModule", ...);
bit that TS generates for marking the module as an ES6 module.

Fixes #427.


evmar:
I'm not sure the extra test is needed given that this touches the other tests, what do you think?

src/es5processor.ts
```diff
@@ -284,6 +289,11 @@ class ES5Processor extends Rewriter {
     return this.isRequire(arg as ts.CallExpression);
   }
 
+  isEsModuleProperty(expr: ts.ExpressionStatement): boolean {
+    // We're matching the explicit source text generated by the TS compiler.
+    return expr.getText() === 'Object.defineProperty(exports, "__esModule", { value: true });';
```
alexeagle: did you consider matching the tree a bit more deeply? What if they change the whitespace around the brackets later? I'm on the fence about whether it would be better to check eg.
```
expr.LHS = 'Object'
expr.RHS = 'defineProperty'
expr.RHS.arguments = ['exports', '__esModule', *anything*]
```

evmar: I considered it and am also on the fence but decided this was simpler.  I think if they change the output maybe it's worth having our tests notice it?

alexeagle:
SGTM. I rebased my PR on this one.

PR #459, opened 2017-04-10
Fix semantic closure comments being lost when not outputing "use strict";

LucasSloan:
TypeScript's AST attaches comments to statements, so when we don't output a statement, we need to make sure to output its attached comment(s).


LucasSloan:
I've updated the Golden Test run code to use the noImplicitUseStrict flag, which would cause this problem to manifest (see this [file](https://github.com/angular/tsickle/blob/master/test_files/file_comment/file_comment.ts)), but the fix means no change to the goldens.

PR #462, opened 2017-04-11
fix: re-write index imports using TypeScript's resolved module path

### LINKED ISSUE #458: Index import rewriting broken
Two scenarios:
package.json has a main field which points to another spot.

`node_modules/foo/package.json`:
```
"main": "dist/index.js",
"typings": "dist/index.d.ts",
```
`user.ts`:
```
import * as f from 'foo';
```

rewrites as 
```
import * as f from 'foo/index';
```
it should instead be 
```
import * as f from 'foo/dist/index';
```

Also, 
```
import {a} from './';
```
with an index.ts in the same directory produces
```
import {a} from './/';
```

/cc @frozenpandaz

alexeagle:
fixes #458

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+import * as path from 'path';
 import {SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
@@ -470,6 +471,9 @@ class ClosureRewriter extends Rewriter {
 type HasTypeParameters =
     ts.InterfaceDeclaration|ts.ClassLikeDeclaration|ts.TypeAliasDeclaration|ts.SignatureDeclaration;
 
+// Matches common extensions of TypeScript input filenames
+const extension = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
+
 /** Annotator translates a .ts to a .ts with Closure annotations. */
 class Annotator extends ClosureRewriter {
   /**
@@ -854,13 +858,14 @@ class Annotator extends ClosureRewriter {
       }
       const resolved = ts.resolveModuleName(moduleId, this.file.fileName, this.tsOpts, this.host);
       if (resolved && resolved.resolvedModule) {
-        const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(/(\.d)?\.ts$/, '');
-        const requestedModule = moduleId.replace(/\.js$/, '');
-        // If the imported module resolves to foo/index, but the specified module was foo, then we
-        // append the /index.
-        if (resolvedModule.substr(resolvedModule.length - 6) === '/index' &&
-            requestedModule.substr(requestedModule.length - 6) !== '/index') {
-          moduleId += '/index';
+        const requestedModule = moduleId.replace(extension, '');
+        const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(extension, '');
+        if (requestedModule.substr(requestedModule.lastIndexOf('/')) !==
+            resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
+          const relative = './' + path.relative(path.dirname(this.file.fileName), resolvedModule);
+          // moduleResolution=node support: if importing from under
+          // node_modules, omit the leading part of the path.
+          moduleId = relative.replace(/^(\.+\/)+node_modules\//, '');
         }
       }
     }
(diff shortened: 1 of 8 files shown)


src/tsickle.ts
```diff
@@ -469,6 +470,8 @@ class ClosureRewriter extends Rewriter {
 type HasTypeParameters =
     ts.InterfaceDeclaration|ts.ClassLikeDeclaration|ts.TypeAliasDeclaration|ts.SignatureDeclaration;
 
+const EXT = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
```
evmar: Doc or give a better variable name?  Not sure what this is for.

src/tsickle.ts
```diff
@@ -853,13 +856,20 @@ class Annotator extends ClosureRewriter {
       }
       const resolved = ts.resolveModuleName(moduleId, this.file.fileName, this.tsOpts, this.host);
       if (resolved && resolved.resolvedModule) {
-        const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(/(\.d)?\.ts$/, '');
-        const requestedModule = moduleId.replace(/\.js$/, '');
+        const requestedModule = moduleId.replace(EXT, '');
+        const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(EXT, '');
         // If the imported module resolves to foo/index, but the specified module was foo, then we
         // append the /index.
```
evmar: I think this comment is obsolete now?

src/tsickle.ts
```diff
@@ -853,13 +856,20 @@ class Annotator extends ClosureRewriter {
...
-          moduleId += '/index';
+        if (requestedModule.substr(requestedModule.lastIndexOf('/')) !==
+            resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
+          moduleId = path.relative(this.file.fileName, resolvedModule)
+                         // path.relative starts from the parent directory (../self) but ES6
+                         // imports are relative to the current directory (./self) so we strip
+                         // one dot.
+                         .replace(/^\.\./, '.')
+                         // moduleResolution=node support: if importing from under
+                         // node_modules, omit the leading part of the path.
```
evmar: Will this get confused by the real node_modules dir in g3?

src/tsickle.ts
```diff
@@ -853,13 +856,20 @@ class Annotator extends ClosureRewriter {
...
         // append the /index.
-        if (resolvedModule.substr(resolvedModule.length - 6) === '/index' &&
-            requestedModule.substr(requestedModule.length - 6) !== '/index') {
-          moduleId += '/index';
+        if (requestedModule.substr(requestedModule.lastIndexOf('/')) !==
+            resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
+          moduleId = path.relative(this.file.fileName, resolvedModule)
+                         // path.relative starts from the parent directory (../self) but ES6
+                         // imports are relative to the current directory (./self) so we strip
+                         // one dot.
```
evmar: I don't quite follow this.  Why does it start from the parent directory?

alexeagle:
```
$ node
> path.relative('a/b', 'a/c')
'../c'
```
but we would like to import `./c` from `a/b`
How could I explain it better?

src/tsickle.ts
alexeagle: We don't even run this code in google3 - the `convertIndexImportShorthand` option is set to false.

This only matches strings where TypeScript actually resolved a module. If google3 TypeScript were to resolve a file under a node_modules directory, we'd be in trouble already.

src/tsickle.ts
evmar: Oh I see, it treats the first argument as if it's a directory.  Maybe it'd be more readable as
```js
path.relative(path.dirname(this.file.fileName), resolveModule)
```
and then drop the regex??

alexeagle: Sure, though I still need to add a `./` on the front to keep it relative
```
> path.relative(path.dirname('a/b'), 'a/c');
'c'
```

alexeagle:
Turns out this was the wrong fix.

Closure recently supports reading the `package.json` file so we don't need to compensate for node_modules which have a `main` field.
At the same time, this does the wrong thing - it uses the layout of files on the type side (where is the .d.ts file) instead of the value side (where is the .js file). These layouts are typically different, eg. the Jscookie example linked above.

The only case where we should do index rewrites is when the import is implicitly from `/index.js`.

alexeagle:
reverting in https://github.com/angular/tsickle/pull/472

PR #466, opened 2017-04-13
Support heritance clauses in externs.

### LINKED ISSUE #464: tsickle doesn't generate an @extends for externs
```ts
// foo.d.ts
namespace a {
  class X {}
  class Y extends X {}
}
```

```js
/** @constructor */
a.X = function() {};
/** @constructor */  // NB: missing @extends here
a.Y = function() {};
```

mprobst:
Fixes #464.


evmar:
This change ended up not working, causing problems like

node.externs.js:7766: WARNING - tsickle_declare_module._debugger.Request @extends non-object type string
tsickle_declare_module._debugger.Request = function() {};
                                           ^^^^^^^^^^^^^

PR #467, opened 2017-04-13
Do not emit duplicate @deprecated tags.

### LINKED ISSUE #449: Don't duplicate @deprecated tags when unifying overloads
When given input like:
```js
/** @deprecated foo */
function a(x: string);
/** @deprecated bar */
function a(x: number);
```

we should include only one @deprecated tag in the output.

mprobst:
Also consistently use sets for the include operations.

Fixes #449.

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -13,9 +13,9 @@
  * to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
  */
 export interface Tag {
-  // tagName is e.g. "param" in an @param declaration.  It's absent
+  // tagName is e.g. "param" in an @param declaration.  It is the empty string
   // for the plain text documentation that occurs before any @foo lines.
-  tagName?: string;
+  tagName: string;
   // parameterName is the the name of the function parameter, e.g. "foo"
   // in
   //   @param foo The foo param.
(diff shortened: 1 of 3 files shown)


src/jsdoc.ts
```diff
@@ -32,20 +32,13 @@ export interface Tag {
   text?: string;
 }
 
-function arrayIncludes<T>(array: T[], key: T): boolean {
```
evmar: Ah, I think this code predated the use of ES6 in tsickle. :D

src/jsdoc.ts
```diff
@@ -255,8 +247,10 @@ function tagToString(tag: Tag, escapeExtraTags: string[] = []): string {
   return out;
 }
 
+const SINGLETON_TAGS = new Set(['deprecated']);
```
evmar: Comment/docstring

src/jsdoc.ts
```diff
@@ -270,7 +264,12 @@ export function toString(tags: Tag[], escapeExtraTags: string[] = []): string {
 
   let out = '';
   out += '/**\n';
+  const emitted = new Set<string>();
   for (let tag of tags) {
+    if (emitted.has(tag.tagName) && SINGLETON_TAGS.has(tag.tagName)) {
```
evmar: This vaguely feels like it might belong in the merge() function, which handles multiple copies of other tags as well.  Another way of saying this is that this toString function is careful to emit exactly what it was given, rather than trying to fix up its input, so the fixing maybe belongs at a different layer (not necessarily merge). 

Maybe for now split this into a cleanup() function that toString() always calls, and cleanup() does the singleton-filtering?  Or maybe I am worrying too much about purity here and what you have is fine.  Your call.

mprobst: When trying to fix this, my impression was that overall our approach to JSDoc is too simplistic. We should have a proper model class with an appropriate data structure backing it. I've added this comment on parse above, which seems the closest place:

```javascript
// TODO(martinprobst): representing JSDoc as a list of tags is too simplistic. We need functionality
// such as merging (below), de-duplicating certain tags (@deprecated), and special treatment for
// others (e.g. @suppress). We should introduce a proper model class with a more suitable data
// strucure (e.g. a Map<TagName, Values[]>).
```

Issue #463, opened 2017-04-13
Handle types from TypeScript base .d.ts libraries that are missing in Closure


mprobst:
E.g. 

```
ArrayLike
HTMLCollectionOf
HTMLTableDataCellElement
HTMLTableHeaderCellElement
NodeListOf
RegExpExecArray
```

Possibly using a `.js` file for Closure that aliases the types?

PR #469, opened 2017-04-14
@suppress {checkTypes}

### LINKED ISSUE #461: Initial comments (e.g. @fileOverview) lost in ES5 mode
In https://github.com/angular/tsickle/commit/ffcf75844026024b753e78dde0eadbfe77ba92a9 we insert a newline to make top-level comments preserve in the ES6 output.  But we also want these comments to stick around in ES5 mode, for tools that work in dev mode.

mprobst:
Emits an `@suppress {checkTypes}` for all Closure code. This is a partial fix for #463.

Also undoes the attempt to preserve @fileoverview tags that only works in ES6 mode, and with that fixes #461.


mprobst:
@alexeagle sorry, the issue reporting is a bit messy here. We had multiple problems:

1. tsickled code references symbols unknown to Closure, such as `NodeListOf<...>`
2. type decisions in tsickled code might not match Closure's view of the world, e.g. whether some return value is assignable to some variable's type
3. `@fileoverview` comments would get dropped, which caused various problems.

We fixed 1., but we still need to fix 2.

For 3., we had a fix that only works in prod mode in g3. I revert that here because I'm too lazy to create a separate PR. The Real Fix (tm) is using a lint message to teach people to add a line break.

Does that make sense?

mprobst:
Friendly ping @evmar @alexeagle.

I've verified this to work internally.

src/rewriter.ts
```diff
@@ -92,10 +92,23 @@ export abstract class Rewriter {
...
-  // Write a span of the input file as expressed by absolute offsets.
-  // These offsets are found in attributes like node.getFullStart() and
-  // node.getEnd().
+  private skipToOffset = 0;
+
+  /**
+   * Skip emitting any code before the given offset. Used to avoid emitting @fileoverview comments
+   * twice.
+   */
+  skipUpToOffset(offset: number) {
```
alexeagle: nit: I'd name the property and method the same. any particular reason it's not just a setter?

src/tsickle.ts
```diff
@@ -761,6 +764,53 @@ class Annotator extends ClosureRewriter {
...
   }
 
+  private handleSourceFile(sf: ts.SourceFile) {
+    const comments = ts.getLeadingCommentRanges(sf.getFullText(), 0);
+    // Add a @suppress {checkTypes} tag to each source file's JSDoc comment.
+    // Be careful to retain an existing comment and its @suppress'ions
+    let fileoverviewTags: jsdoc.Tag[]|null = null;
+    if (comments && comments.length) {
+      // Find any existing @fileoverview comment and merge our @suppress tag with it.
+      // JSCompiler considers the last comment with @fileoverview (or @externs or @nocompile)
```
alexeagle: nit: use the external name "closure compiler"

src/tsickle.ts
```diff
@@ -761,6 +764,53 @@ class Annotator extends ClosureRewriter {
...
+      const parseResult = jsdoc.parse(sf.getFullText().substring(comments[0].pos, comments[0].end));
+      if (parseResult && parseResult.tags.some(t => t.tagName === 'fileoverview')) {
+        fileoverviewTags = parseResult.tags;
+        // Do not emit the fileoverview comment again later, otherwise we'll end up with duplicated
+        // @fileoverview, @modName etc comments.
+        this.skipUpToOffset(comments[0].end);
+      }
+    }
+    if (fileoverviewTags) {
+      // Add @suppress {checkTypes}, or add to the list in an existing @suppress tag.
```
alexeagle: does it matter if there's an existing one? doesn't it also work to
```
/**
 * @fileoverview file is nice
 * @suppress{problemA,problemB}
 * @suppress{problemC}
 */
```

If so maybe the complexity here isn't worth the trouble

test_files/arrow_fn.untyped/arrow_fn.untyped.js
```diff
@@ -1 +1,5 @@
-goog.module('test_files.arrow_fn.untyped.arrow_fn.untyped');var module = module || {id: 'test_files/arrow_fn.untyped/arrow_fn.untyped.js'};var /** @type {?} */ fn3 = (a) => 12;
+goog.module('test_files.arrow_fn.untyped.arrow_fn.untyped');var module = module || {id: 'test_files/arrow_fn.untyped/arrow_fn.untyped.js'};/**
```
alexeagle: this stuff was all jammed on one line to avoid changing the sourcemap offsets, right? Is there any reason we still have to do that? If there is a reason, does this change cause a problem with it?

src/rewriter.ts
mprobst: It cannot have the same name – in JS, properties and functions are in the same namespace. I didn't want to just assign into a field across two files and a class hierarchy, but then again this is all internal to the library, and we don't really have proper separation here anyway.

So changed to a simple field.

src/tsickle.ts

src/tsickle.ts
mprobst: Added a comment, Closure Compiler fails fatally if there are multiple `@suppress` tags in a file :-/

test_files/arrow_fn.untyped/arrow_fn.untyped.js
mprobst: I think now that we presumably have working source maps, this should be fine. I'm not sure if we properly compose them in dev mode though?

PR #471, opened 2017-04-18
Upgrade to TypeScript 2.3.

voithos:
The `.path` property of SourceFile has now been marked @internal, but it
seems we can just use `.fileName` to the same effect.

Note that before this is merged, we should create a ts22 branch in tsickle for future commits that need to target TS 2.2.


package.json
```diff
@@ -43,7 +43,7 @@
     "mocha": "^3.2.0",
     "temp": "^0.8.1",
     "tslint": "^3.15.1",
-    "typescript": "2.2.1"
+    "typescript": "2.3.0"
```
voithos: Note that this file also references `typescript` under "peerDependencies", but I wasn't sure if that should also be modified, since it wasn't changed when upgrading to 2.2 (it is currently set to `^2.1.0`).

voithos:
Done - ran `yarn` and the TS version was updated in the lock file (It did re-format a bunch of other things in the lock file, though). I suppose we'll probably want to do this again later once TS 2.3.0 finalizes (it's currently RC)?

package.json
evmar: I think we should change both.  The peerDependencies are only used from npm, not g3, which is why we didn't notice, but they should be the same value.

(And then run yarn again just to check)

voithos: Done. I removed the leading `^`, as it didn't make sense to make it different from the other version specifier.
Ran yarn, and nothing was modified.

evmar:
Note that we might need to rollback the other change that you are CC'd on about the bug.

PR #472, opened 2017-04-19
Fix index imports take 2

alexeagle:
index imports should not apply to node_modules

See notes at https://github.com/angular/tsickle/pull/462#issuecomment-295433380
Closure understands package.json, and third-party libraries may have a different shape
when TypeScript resolves modules (.d.ts layout is different from .js layout)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Tsickle - TypeScript to Closure Translator [![Build Status](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle)
+# Tsickle - TypeScript to Closure Translator [![Linux build](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle) [![Windows build](https://ci.appveyor.com/api/projects/status/puxdblmlqbofqqt1/branch/master?svg=true)](https://ci.appveyor.com/project/alexeagle/tsickle/branch/master)
 
 Tsickle converts TypeScript code into a form acceptable to the [Closure
 Compiler].  This allows using TypeScript to transpile your sources, and then
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 build
 
 # Dependency directory
-./node_modules
+/node_modules
 
 # WebStorm/IntelliJ directory
 .idea
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -860,12 +860,12 @@ class Annotator extends ClosureRewriter {
       if (resolved && resolved.resolvedModule) {
         const requestedModule = moduleId.replace(extension, '');
         const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(extension, '');
-        if (requestedModule.substr(requestedModule.lastIndexOf('/')) !==
-            resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
-          const relative = './' + path.relative(path.dirname(this.file.fileName), resolvedModule);
-          // moduleResolution=node support: if importing from under
-          // node_modules, omit the leading part of the path.
-          moduleId = relative.replace(/^(\.+\/)+node_modules\//, '');
+        if (resolvedModule.indexOf('node_modules') === -1 &&
+            requestedModule.substr(requestedModule.lastIndexOf('/')) !==
+                resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
+          moduleId = './' +
+              path.relative(path.dirname(this.file.fileName), resolvedModule)
+                  .replace(path.sep, '/');
         }
       }
     }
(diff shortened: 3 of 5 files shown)


b-strauss:
I do not know if that was the reason, but the change in the .gitignore makes git think I have 4000 unversioned files.

alexeagle:
@b-strauss - no that wasn't the reason to revert. Do you have a `node_modules` folder somewhere other than next to the .gitignore in the project root?

b-strauss:
@alexeagle no, I just deleted everything and checked out clean again. It is the node_modules in the project root that makes trouble. My IDE (Intellij IDEA) even says this line is problematic `./node_modules`. Maybe this problem is only visible on Windows? Does the path need to be relative? I thought paths inside .gitignore are already relative to the location of the file?

evmar:
I remember seeing the `./` in the .gitignore and wondering to myself whether it worked or not.  I think what you really want is `/node_modules`, which from the docs:

       ·   A leading slash matches the beginning of the pathname. For example, "/*.c" matches
           "cat-file.c" but not "mozilla-sha1/sha1.c".

alexeagle:
Thanks Evan, I was just checking that doc as well. Updated in this PR.

alexeagle:
note that this is still broken if you have
`node_modules/f/path/index.js`
and the user
`import {} from 'f/path'`

I don't see a way to fix this in tsickle, we should just convince Closure compiler to look for a module `xx$index` if they fail to find module `xx`.

b-strauss:
@alexeagle FYI - the test for this fails on Windows.

```
1) golden tests index_import:

    test_files\index_import\user.tsickle.ts
    + expected - actual

     /// <ref './library.d.ts'>
     
    -import {a} from './has_index\index';
    +import {a} from './has_index/index';
     const tsickle_forward_declare_1 = goog.forwardDeclare('test_files.index_import.has_index.index');
    -export {a} from './has_index\index';
    +export {a} from './has_index/index';
     const tsickle_forward_declare_2 = goog.forwardDeclare('test_files.index_import.has_index.index');
    -export {a} from './has_index\index';
    +export {a} from './has_index/index';
     const tsickle_forward_declare_3 = goog.forwardDeclare('test_files.index_import.has_index.index');
     import {a as a2} from './has_index/index';
     const tsickle_forward_declare_4 = goog.forwardDeclare('test_files.index_import.has_index.index');
     import {a as a3} from './has_index/index.js';
     const tsickle_forward_declare_5 = goog.forwardDeclare('test_files.index_import.has_index.index');
     import {b} from './lib';
     const tsickle_forward_declare_6 = goog.forwardDeclare('test_files.index_import.lib');
    -import * as foo from './node_modules\foo\subdir\foo';
    +import * as foo from 'foo/subdir/foo';
```

alexeagle:
@b-strauss okay let's do that appveyor thing :)
I think I can get the permission...

alexeagle:
This is green on Appveyor now (it's not reporting a status here, maybe that's because the PR was created before Appveyor was set up with webhooks? green run is here: https://ci.appveyor.com/project/alexeagle/tsickle )

b-strauss:
@alexeagle nice! I am not sure about the github reporting, maybe it needs to be configured somewhere on github? I guess we could just create a test pull request after this is merged.

Two more things:

We should use the master branch for the badge `https://ci.appveyor.com/api/projects/status/5dbo0affcbyxce55/branch/master?svg=true`
and the build version format in the appveyor project settings should be changed from `1.0.{build}` to just `{build}`, so people do not think tsickle is on version 1.0.

Thanks!

alexeagle:
Updated those things. It sucks that appveyor assigns the URL
"/project/alexeagle/tsickle" even though the project being built is
"angular/tsickle". I deleted and recreated the projects and that's still
the case. Any idea if I can get their URL structure to match the github
org/repo structure?

b-strauss:
AFAIK this is always `username/project`. If you want `angular/tsickle`, you would have to create an account for that. Maybe there is already one at google? The Angular-CLI seems to have created one specifically for the project (https://ci.appveyor.com/project/AngularCLI/angular-cli).

README.md
```diff
@@ -1,4 +1,4 @@
-# Tsickle - TypeScript to Closure Translator [![Linux build](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle) [![Windows build](https://ci.appveyor.com/api/projects/status/5dbo0affcbyxce55?svg=true)](https://ci.appveyor.com/project/alexeagle/tsickle-5vglt)
+# Tsickle - TypeScript to Closure Translator [![Linux build](https://travis-ci.org/angular/tsickle.svg?branch=master)](https://travis-ci.org/angular/tsickle) [![Windows build](https://ci.appveyor.com/api/projects/status/puxdblmlqbofqqt1?svg=true)](https://ci.appveyor.com/project/alexeagle/tsickle)
```
b-strauss: Missing `/branch/master` ;)

PR #482, opened 2017-04-27
Upgrade to TypeScript 2.3.1.

voithos:
The underlying implementation of getTypeOfSymbolAtLocation hasn't
changed, but a slightly different call chain causes the compiler to
reject `undefined` arguments.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -439,7 +439,7 @@ export class TypeTranslator {
           indexable = true;
           break;
         default:
-          let member = type.symbol.members.get(field);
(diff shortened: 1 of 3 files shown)


src/type-translator.ts
```diff
@@ -442,7 +442,7 @@ export class TypeTranslator {
           let member = type.symbol.members.get(field);
```
evmar: Better: make this `.get(field)!`.
We know member is never-null because we are iterating the keys of the map.

voithos: True, done.

PR #488, opened 2017-05-04
always emit interfaces, even in untyped mode

evmar:
This changes inches us closer to fully typed mode, by making us
always emit a type definition even in untyped mode.

We accomplish this by moving around a few if statements.  The
resulting interfaces don't have heritance clauses (that broke
in google3) and the fields don't have types, but it means that
at least a type that is declared in TypeScript is visible (albeit
slightly mangled) to Closure code.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1038,8 +1038,6 @@ class Annotator extends ClosureRewriter {
   }
 
   private emitInterface(iface: ts.InterfaceDeclaration) {
-    if (this.options.untyped) return;
-
     // If this symbol is both a type and a value, we cannot emit both into Closure's
     // single namespace.
     let sym = this.program.getTypeChecker().getSymbolAtLocation(iface.name);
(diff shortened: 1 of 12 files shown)


src/tsickle.ts
```diff
@@ -1120,7 +1120,9 @@ class Annotator extends ClosureRewriter {
       case ts.SyntaxKind.StringLiteral:
         // E.g. interface Foo { 'bar': number; }
         // If 'bar' is a name that is not valid in Closure then there's nothing we can do.
-        return (prop.name as ts.StringLiteral).text;
+        let text = (prop.name as ts.StringLiteral).text;
+        if (!isValidClosurePropertyName(text)) return null;
```
rkirov: This might need a test too.

src/tsickle.ts
```diff
@@ -1038,17 +1038,17 @@ class Annotator extends ClosureRewriter {
...
     // single namespace.
     let sym = this.program.getTypeChecker().getSymbolAtLocation(iface.name);
     if (sym.flags & ts.SymbolFlags.Value) return;
 
     let docTags = this.getJSDoc(iface) || [];
     docTags.push({tagName: 'record'});
-    this.maybeAddTemplateClause(docTags, iface);
-    this.maybeAddHeritageClauses(docTags, iface);
+    if (!this.options.untyped) {
+      this.maybeAddTemplateClause(docTags, iface);
```
rkirov: I guess it makes sense for templates to get dropped, because all keys are '?', maybe add a comment because the reason for maybeAddTemplateClause is different than maybeAddHeritageClauses.

src/tsickle.ts
evmar: Done.

src/tsickle.ts
evmar: This is the same logic as the other user of maybeAddTemplateClause, I just copied it blindly.  I assume it's here for a good reason.

PR #491, opened 2017-05-10
Qualify type names using their global name. 

mprobst:
Types declared in namespaces must be qualified in Closure Compiler. While
Closure Compiler supports introducing local aliases using imports, its safer to
always emit global types (where the FQN is a dotted path).


src/type-translator.ts
```diff
@@ -164,12 +164,24 @@ export class TypeTranslator {
...
    */
-  public symbolToString(sym: ts.Symbol): string {
+  public symbolToString(sym: ts.Symbol, useFqn: boolean): string {
     // This follows getSingleLineStringWriter in the TypeScript compiler.
     let str = '';
     let alias = this.symbolsToAliasedNames.get(sym);
     if (alias) return alias;
+    if (useFqn) {
+      const fqn = this.typeChecker.getFullyQualifiedName(sym);
+      if (!fqn.startsWith(`"`) && !fqn.startsWith(`'`)) {
```
evmar: Can you comment here about what it means for a fqn to be quoted?  I don't quite follow.

src/type-translator.ts
```diff
@@ -164,12 +164,24 @@ export class TypeTranslator {
...
+        return this.stripClutzNamespace(fqn);
+      } else {
+        // TODO(martinprobst): only warn in externs, this is OK for
+        // non-ambient contexts.
```
evmar: I don't get this part either.

src/type-translator.ts
```diff
@@ -196,13 +208,18 @@ export class TypeTranslator {
       reportIllegalExtends: doNothing,
     };
     builder.buildSymbolDisplay(sym, writer, this.node);
-    // Clutz (https://github.com/angular/clutz) emits global type symbols hidden in a special
-    // ಠ_ಠ.clutz namespace. While most code seen by Tsickle will only ever see local aliases, Clutz
-    // symbols can be written by users directly in code, and they can appear by dereferencing
-    // TypeAliases. The code below simply strips the prefix, the remaining type name then matches
-    // Closure's type.
     if (str.startsWith('ಠ_ಠ.clutz.')) str = str.substring('ಠ_ಠ.clutz.'.length);
```
evmar: It looks like you meant to delete this line

src/type-translator.ts
```diff
@@ -164,12 +164,24 @@ export class TypeTranslator {
    * - TypeChecker.typeToString translates Array as T[].
    * - TypeChecker.symbolToString emits types without their namespace,
    *   and doesn't let you pass the flag to control that.
+   * @param useFqn whether to scope the name using its fully qualified name.
    */
-  public symbolToString(sym: ts.Symbol): string {
+  public symbolToString(sym: ts.Symbol, useFqn: boolean): string {
```
evmar: Would it make sense to default this to true?  It seems you only want it off in one special case.

mprobst: I've recently preferred to avoid optional arguments. I often have the impression that it's better to be explicit at callsites as to what functionality you want, rather than have things work some way by default.

E.g. here, I think it makes sense that you need to think about whether a name should be qualified or not if you're adding new code calling into it.

src/type-translator.ts
mprobst: Added docs.

src/type-translator.ts
mprobst: More docs added.

src/type-translator.ts
mprobst: I did, in a different git branch...

PR #492, opened 2017-05-10
Simplify types in decorator downlevel.

alexeagle:
This is needed to allow declarations:true on the resulting TS code, because the DecoratorInvocation type
was part of the public API but not exported.


alexeagle:
Build is fixed now PTAL

PR #495, opened 2017-05-15
Correctly handle files with multiple leading comments.

mprobst:
The `@fileoverview` comment can actually follow any number of comments,
including `@license` comments, and, surprisingly, even other `@fileoverview`
comments, even ones with `@suppress` tags (which Closure Compiler silently
ignores).

This means our parsing logic has to consider *all* comments, and augment the
last one that has an `@fileoverview`.


evmar:
Should we disallow multiple @fileoverview comments so that this doesn't come up?
We also could consider escaping them, like we do for most other @foo comments because we're just generating code for closure.

mprobst:
@evmar not sure. I hope we don't have any, OTOH you'll always find odd things in `third_party` code. It wouldn't change the code substantially, so we might as well do what works?

It's a Closure Compiler issue that this is so ambiguous, but it might not be ours to fix. I'm a bit weary of fixing such things in tsickle - it seems like the wrong layer.

src/tsickle.ts
```diff
@@ -765,51 +765,68 @@ class Annotator extends ClosureRewriter {
...
-    this.emit(jsdoc.toString(fileoverviewTags));
-    this.emit('\n\n');  // separate from file body to avoid being dropped.
-    for (const stmt of sf.statements) {
-      this.visit(stmt);
+    const comment = comments[fileoverviewIdx];
+    this.writeRange(0, comment.pos);
+    this.skipUpToOffset = comment.end;
+
+    const parsed = jsdoc.parse(sf.getFullText().substring(comment.pos, comment.end));
+    if (!parsed) throw new Error('parsed just before?');
```
alexeagle: give a better error message

mprobst: This is a "this cannot happen" case, the code above has just validate that the thing at the index can be parsed. I've made it a bit more clear this is an internal error.

b-strauss:
FYI: Windows output

```
103 passing (14s)
2 failing
1) golden tests file_comment:
    test_files\file_comment\fileoverview_comment_add_suppress.tsickle.ts
    + expected - actual
      * @suppress {checkTypes}
      */
     
     
    -
    -
     // here comes code.
     export let /** @type {number} */ x = 1;
    
    at compareAgainstGolden (D:\workspace\_external\test\tsickle_test.ts:68:23)
    at _loop_1 (D:\workspace\_external\test\tsickle_test.ts:169:9)
    at Context.<anonymous> (build\test\tsickle_test.js:169:17)
2) golden tests use_closure_externs:
    test_files\use_closure_externs\use_closure_externs.tsickle.ts
    + expected - actual
      * @suppress {checkTypes}
      */
     
     
    -
    -
     console.log('work around TS dropping consecutive comments');
     
     let /** @type {!NodeListOf<!HTMLParagraphElement>} */ x: NodeListOf<HTMLParagraph
     console.log(x);
    
    at compareAgainstGolden (D:\workspace\_external\test\tsickle_test.ts:68:23)
    at _loop_1 (D:\workspace\_external\test\tsickle_test.ts:169:9)
    at Context.<anonymous> (build\test\tsickle_test.js:169:17)
```

mprobst:
Is that a \r\n vs \n issue? I thought we normalize the line breaks after
test execution?

b-strauss:
Problems on Windows are either line separator or path separator problems.

We normalize line breaks before test execution here:
https://github.com/angular/tsickle/blob/master/test/tsickle_test.ts#L51-L52

I'm not sure why this is failing and I currently don't have time to debug this, but I would be happy if we finally had working Windows CI tests. @alexeagle 😆

b-strauss:
The problem seems to be in the condition here:

https://github.com/angular/tsickle/blob/8ebfd011e9386997ca94001210596c5fd75cd17d/src/tsickle.ts#L836-L838

PR #498, opened 2017-05-16
More `@fileoverview` comments.

mprobst:
It turns out that various tooling treats comments containing `@modName`,
`@mods`, and `@pintomodule` as a file overview comment.

Also adds some more allowed JSDoc comments I found in Closure Compiler's
source code.

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -13,22 +13,32 @@
  * to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
  */
 export interface Tag {
-  // tagName is e.g. "param" in an @param declaration.  It is the empty string
-  // for the plain text documentation that occurs before any @foo lines.
+  /**
+   * tagName is e.g. "param" in an @param declaration.  It is the empty string
+   * for the plain text documentation that occurs before any @foo lines.
+   */
   tagName: string;
-  // parameterName is the the name of the function parameter, e.g. "foo"
-  // in
-  //   @param foo The foo param.
+  /**
+   * parameterName is the the name of the function parameter, e.g. "foo"
+   * in `\@param foo The foo param`
+   */
   parameterName?: string;
+  /**
+   * The type of a JSDoc \@param, \@type etc tag, rendered in curly braces.
+   * Can also hold the type of an \@suppress.
+   */
   type?: string;
-  // optional is true for optional function parameters.
+  /** optional is true for optional function parameters. */
   optional?: boolean;
-  // restParam is true for "...x: foo[]" function parameters.
+  /** restParam is true for "...x: foo[]" function parameters. */
   restParam?: boolean;
-  // destructuring is true for destructuring bind parameters, which require
-  // non-null arguments on the Closure side.  Can likely remove this
-  // once TypeScript nullable types are available.
+  /**
+   * destructuring is true for destructuring bind parameters, which require
+   * non-null arguments on the Closure side.  Can likely remove this
+   * once TypeScript nullable types are available.
+   */
   destructuring?: boolean;
+  /** Any remaining text on the tag, e.g. the description. */
   text?: string;
 }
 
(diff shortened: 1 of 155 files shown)


src/jsdoc.ts
```diff
@@ -37,44 +37,66 @@ export interface Tag {
  * The public Closure docs don't list all the tags it allows; this list comes
  * from the compiler source itself.
  * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/Annotation.java
+ * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
```
evmar: This seems a bit risky -- it means we pass through all these comments verbatim.  Is that your intent?

src/jsdoc.ts
```diff
@@ -111,19 +141,10 @@ const JSDOC_TAGS_WHITELIST = new Set([
...
-  'public',
-  'record',
-  'template',
-  'this',
-  'type',
-  'typedef',
+  'augments', 'class',     'constructs', 'constructor', 'enhance',    'enhanceable',
+  'enum',     'extends',   'field',      'function',    'implements', 'interface',
+  'lends',    'namespace', 'private',    'public',      'record',     'static',
+  'template', 'this',      'type',       'typedef',
```
evmar: I think the vertical layout diffs better, shrug

src/tsickle.ts
```diff
@@ -774,15 +774,21 @@ class Annotator extends ClosureRewriter {
     }
   }
 
+  static readonly FILEOVERVIEW_COMMENTS =
+      new Set(['fileoverview', 'externs', 'modName', 'mods', 'pintomodule']);
```
evmar: File-level constant

src/tsickle.ts
```diff
@@ -774,15 +774,21 @@ class Annotator extends ClosureRewriter {
...
+
   /**
    * Emits an \@suppress {checkTypes} fileoverview comment.
    */
   private emitSuppressChecktypes(sf: ts.SourceFile) {
     const comments = ts.getLeadingCommentRanges(sf.getFullText(), 0) || [];
+    // comments.push(sf.getChildAt(0).getLeadingTrivia)
```
evmar: xxx

src/tsickle.ts
```diff
@@ -815,12 +821,17 @@ class Annotator extends ClosureRewriter {
     // only appear once and be merged.
     const suppressIdx = tags.findIndex(t => t.tagName === 'suppress');
     if (suppressIdx !== -1) {
-      const suppressions =
-          (tags[suppressIdx].text || '').replace(/^\{(.*)\}$/, '$1').split(',').map(s => s.trim());
-      if (suppressions.indexOf('checkTypes') === -1) {
-        suppressions.push('checkTypes');
+      const match = (tags[suppressIdx].text || '').match(/^\{(.*)\}(.*)$/);
+      if (!match) {
+        this.error(sf, `Failed to parse @suppress tag ${tags[suppressIdx].text}`);
```
evmar: I wonder if this belongs in jsdoc.parse().  It already has some logic around parsing {foo} for types.

src/tsickle.ts
```diff
@@ -815,12 +821,17 @@ class Annotator extends ClosureRewriter {
...
+        return;
+      }
+      const [, suppressions, text] = match;
+      const suppressionsList = suppressions.split(',').map(s => s.trim());
+      if (suppressionsList.indexOf('checkTypes') === -1) {
+        suppressionsList.push('checkTypes');
       }
-      tags[suppressIdx].text = `{${suppressions.join(',')}}`;
+      tags[suppressIdx].text = `{${suppressionsList.join(',')}}${text}`;
```
evmar: and then this would belong in jsdoc.tagToString()

src/tsickle.ts
alexeagle: interesting that `@license` is not one of them

src/tsickle.ts
alexeagle: remove stray comment?

test_files/file_comment/other_fileoverview_comments.js
```diff
@@ -0,0 +1,6 @@
+goog.module('test_files.file_comment.other_fileoverview_comments');var module = module || {id: 'test_files/file_comment/other_fileoverview_comments.js'};/**
+ * @modName {some_mod}
+ * @suppress {checkTypes}
+ */
+// @modName also bleongs in a fileoverview comment, so must be merged.
```
alexeagle: spelling: belongs

src/jsdoc.ts
mprobst: clang-format decided on this. I tend to agree, but actually if you have a ws ignoring differ, it's fine.

src/tsickle.ts
mprobst: Yeah, only comments containing `@fileoverview` have `@suppress` parsed in them, and of those, only the last. `@mods` is only parsed out of the first (?) `@fileoverview` by some tooling, but other tooling considers all. So we have every flavour, covering all bases ;-)

mprobst: You're the boss. I did like it being closer to the code though.

src/tsickle.ts

src/jsdoc.ts
```diff
@@ -13,22 +13,32 @@
  * to their API will be easier.  See e.g. ts.JSDocTag and ts.JSDocComment.
  */
 export interface Tag {
-  // tagName is e.g. "param" in an @param declaration.  It is the empty string
-  // for the plain text documentation that occurs before any @foo lines.
+  /**
```
evmar: Too much switching from TS to Go  :(

src/tsickle.ts
mprobst: Actually that's much nicer, done. I'm overloading the "type" field to be anything that comes in curlies after the tag. I've poked at adding another tag, but I cannot come up with a better name for it ("kind"? "variant"? "flavour"? "args"?), so I added docs and think it's best to reuse the field.

src/tsickle.ts
mprobst: done, mostly. I'm keeping the {...} part around as a string.

test_files/file_comment/other_fileoverview_comments.js
mprobst: done.

src/jsdoc.ts
mprobst: As discussed, I've reverted to just the ones we know are load bearing and probably something we want to emit.

PR #501, opened 2017-05-23
Do not unquote declared properties, but emit a warning.

mprobst:
It turns out users do rely on manually quoted properties not to be
renamed, even if there is a separate declaration for the property - at
least we discovered one instance in Angular doing this.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -764,9 +764,11 @@ class Annotator extends ClosureRewriter {
         }
         break;
       case ts.SyntaxKind.ElementAccessExpression:
-        // Convert quoted accesses to properties that have a symbol to dotted accesses, to get
-        // consistent Closure renaming. This can happen because TS allows quoted access with literal
-        // strings to properties.
+        // Warn for quoted accesses to properties that have a symbol declared.
+        // Mixing quoted and non-quoted access to a symbol (x['foo'] and x.foo) risks breaking
+        // Closure Compiler renaming. Quoted access is more cumbersome to write than dotted access
+        // though, so chances are users did intend to avoid renaming. The better fix is to use
+        // `declare interface` though.
         const eae = node as ts.ElementAccessExpression;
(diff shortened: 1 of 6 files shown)


src/tsickle.ts
```diff
@@ -776,11 +778,18 @@ class Annotator extends ClosureRewriter {
...
-        this.writeNode(eae.expression);
-        this.emit(`.${propName}`);
-        return true;
+        const symName = this.typeChecker.symbolToString(quotedPropSym);
+        this.debugWarn(
+            eae,
+            `Declared property ${symName} accessed with quotes. ` +
+                `This can lead to renaming bugs. A better fix is to use 'declare interface' ` +
+                `on the declaration.`);
+        // Previously, the code blow changed the quoted into a non-quoted access.
```
evmar: below

src/tsickle.ts
```diff
@@ -776,11 +778,18 @@ class Annotator extends ClosureRewriter {
...
         if (!quotedPropSym) return false;
         const propName = (eae.argumentExpression as ts.StringLiteral).text;
-        // Properties containing non-JS identifier names must not be unquoted.
+        // Properties containing non-JS identifier names can only be accessed with quotes.
         if (!isValidClosurePropertyName(propName)) return false;
-        this.writeNode(eae.expression);
-        this.emit(`.${propName}`);
-        return true;
+        const symName = this.typeChecker.symbolToString(quotedPropSym);
+        this.debugWarn(
```
alexeagle: I think we have to move it to a linter if we actually want developers to be aware of the issue.
I can get the type-aware linting back up and running today.

src/tsickle.ts

src/tsickle.ts
mprobst: Yes, this doesn't actually get emitted to users. I think it's still useful to keep the code to track the issue, also in our tests.

PR #502, opened 2017-05-24
Add Bazel build rules

alexeagle:
They are not used yet.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,19 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/tsickle
+    docker:
+      - image: alexeagle/ngcontainer
+    steps:
+      - checkout
+      - restore_cache:
+          key: tsickle-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      # Build twice, workaround for
+      # https://github.com/bazelbuild/bazel/issues/3114
+      # Turn off sandboxing, workaround for
+      # https://github.com/alexeagle/ngcontainer/issues/1
+      - run: bazel build --spawn_strategy=standalone ... || bazel build --spawn_strategy=standalone ...
+      - save_cache:
+          key: tsickle-{{ .Branch }}-{{ checksum "yarn.lock" }}
+          paths:
+            - "node_modules"
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 # Compiled JS
-build
+/built
 
 # Dependency directory
 /node_modules
@@ -9,3 +9,6 @@ build
 
 # VSCode directory
 .vscode
+
+# Bazel working directories
+/bazel-*
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,0 +1,17 @@
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "io_bazel_rules_typescript",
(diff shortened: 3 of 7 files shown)


.bazelrc
```diff
@@ -0,0 +1,2 @@
+# Performance: avoid stat'ing input files
+build --watchfs
```
evmar: Does this matter?  tsickle is tiny and stat is fast.

evmar: https://github.com/ninja-build/ninja/wiki/Timing-Numbers

.circleci/config.yml
```diff
@@ -0,0 +1,16 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/tsickle
+    docker:
+      - image: alexeagle/ngcontainer
+    steps:
+      - checkout
+      - restore_cache:
+          key: tsickle-{{ .Branch }}-{{ checksum "yarn.lock" }}
```
evmar: sha1sum instead of checksum (?)
I think checksums aren't intended to be used for for content id

.gitignore
```diff
@@ -1,5 +1,5 @@
 # Compiled JS
-build
+/built
```
evmar: Why change this?

WORKSPACE
```diff
@@ -0,0 +1,17 @@
...
+
+load("@io_bazel_rules_typescript//:defs.bzl", "node_repositories", "yarn_install")
+
+# Install a hermetic version of node
+# After this is run, a label @io_bazel_typescript_node//:bin/node will exist
+node_repositories()
+
+# Install yarn, and run yarn install to create node_modules.
+# Also creates the BUILD file under node_modules.
+yarn_install(package_json = "//:package.json")
```
evmar: Can this be :package.json or do you need the full path?

src/BUILD
```diff
@@ -0,0 +1,9 @@
+package(default_visibility=["//visibility:public"])
+load("@io_bazel_rules_typescript//:defs.bzl", "ts_library")
+
+ts_library(
+    name = "src",
+    srcs = glob(["*.ts"]),
```
evmar: shouldn't this exclude tests?  or we have all tests in a different dir?

src/BUILD
```diff
@@ -0,0 +1,9 @@
...
+    deps = [],
+    tsconfig = "//:tsconfig.json",
```
evmar: weird to see library-specific tsconfigs!

.bazelrc
alexeagle: No I was just cargo-culting from the Angular setup I'm working on, removed.

.circleci/config.yml
alexeagle: It's a SHA256: https://circleci.com/docs/2.0/configuration-reference/#save_cache
(it's not an arbitrary unix command, it's part of their config DSL)

.gitignore
alexeagle: surprising things happen on Mac when you have both BUILD and build/
surprising in a bad way

WORKSPACE
alexeagle: it totally can, since there's a BUILD file in this directory, it's in the current package

src/BUILD
alexeagle: yeah, it's different

src/BUILD
alexeagle: tests are in /test/

.gitignore
evmar: hahahaha

Issue #505, opened 2017-05-30
Better support for abstract classes which does not generate dead code.


mhevery:
## Overview

Currently abstract classes produce a lot of runtime dead code which can not be tree shaken by uglify. The core of the issue is that tsickle converts abstract methods into concrete methods for closure which are than retained by uglify.

Here is a proposal for new way of doing it which does not produce dead code.

Given
```
abstract class Base {
  abstract methodA();
}
```

Tsickle currently generates
```
class Base {
  /**
   * @abstract
   * @return {void}
   */
  methodA() {}
}
 ```

which than downlevels to:
```
var Base=(function(){
  function Base(){}
  Base.prototoype.methodA = function () {};
  return Base;
}());
```
 
Notice that the abstract method now is no longer abstract and it has a method body. This creates an issue because the method is now retained at runtime even though it is never called (it was abstract)
 
## Proposal
[...]

PR #506, opened 2017-05-31
Emit abstract methods as pure declarations in a helper function.

mprobst:
This should help Uglify and similar simple optimizers to remove the declarations entirely.

I verified that the emit is actually checked by Closure - it does understand abstract function declarations in this form.

This also fixes some lint hinting to use "const" instead of "let", and the more modern cast form.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -72,7 +72,7 @@ export function formatDiagnostics(diags: ts.Diagnostic[]): string {
         let res = ts.DiagnosticCategory[d.category];
         if (d.file) {
           res += ' at ' + d.file.fileName + ':';
-          let {line, character} = d.file.getLineAndCharacterOfPosition(d.start);
+          const {line, character} = d.file.getLineAndCharacterOfPosition(d.start);
           res += (line + 1) + ':' + (character + 1) + ':';
         }
         res += ' ' + ts.flattenDiagnosticMessageText(d.messageText, '\n');
@@ -202,7 +202,7 @@ class ClosureRewriter extends Rewriter {
       // JSDoc without any change; @param/@return are handled specially.
       // TODO: there may be problems if an annotation doesn't apply to all overloads;
       // is it worth checking for this and erroring?
-      for (let tag of docTags) {
+      for (const tag of docTags) {
         if (tag.tagName === 'param' || tag.tagName === 'return') continue;
         newDoc.push(tag);
       }
@@ -229,7 +229,7 @@ class ClosureRewriter extends Rewriter {
         const name = getParameterName(paramNode, i);
         const isThisParam = name === 'this';
 
-        let newTag: jsdoc.Tag = {
+        const newTag: jsdoc.Tag = {
           tagName: isThisParam ? 'this' : 'param',
           optional: paramNode.initializer !== undefined || paramNode.questionToken !== undefined,
           parameterName: isThisParam ? undefined : name,
@@ -245,7 +245,7 @@ class ClosureRewriter extends Rewriter {
         }
         newTag.type = this.typeToClosure(fnDecl, type);
 
-        for (let {tagName, parameterName, text} of docTags) {
+        for (const {tagName, parameterName, text} of docTags) {
           if (tagName === 'param' && parameterName === newTag.parameterName) {
             newTag.text = text;
             break;
@@ -257,10 +257,10 @@ class ClosureRewriter extends Rewriter {
 
       // Return type.
       if (!isConstructor) {
-        let retType = typeChecker.getReturnTypeOfSignature(sig);
-        let retTypeString: string = this.typeToClosure(fnDecl, retType);
+        const retType = typeChecker.getReturnTypeOfSignature(sig);
+        const retTypeString: string = this.typeToClosure(fnDecl, retType);
         let returnDoc: string|undefined;
-        for (let {tagName, text} of docTags) {
+        for (const {tagName, text} of docTags) {
           if (tagName === 'return') {
             returnDoc = text;
             break;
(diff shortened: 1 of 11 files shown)


evmar:
This change does X, and also does totally unrelated thing Y.  ಠ_ಠ

src/tsickle.ts
```diff
@@ -678,32 +678,17 @@ class Annotator extends ClosureRewriter {
...
-              // Can you even have an unnamed abstract function?
-              this.error(fnDecl, 'anonymous abstract function');
-              return false;
-            }
-            this.writeRange(fnDecl.name.getStart(), fnDecl.parameters.end);
-            this.emit(') {}');
-            return true;
+            // Emit normally. Declaring the type is handled in visitClassDeclaration below.
+            return false;
           }
```
evmar: It appears you can delete this 'if' statement because it always returns false after anyway.

src/tsickle.ts
```diff
@@ -1171,60 +1156,82 @@ class Annotator extends ClosureRewriter {
...
     staticProps.forEach(p => this.visitProperty([className], p));
-    let memberNamespace = [className, 'prototype'];
+    const memberNamespace = [className, 'prototype'];
     nonStaticProps.forEach((p) => this.visitProperty(memberNamespace, p));
     paramProps.forEach((p) => this.visitProperty(memberNamespace, p));
+
+    for (const fnDecl of abstractMethods) {
+      const name = this.propertyName(fnDecl);
+      if (!name) {
+        this.error(fnDecl, 'anonymous abstract function');
```
evmar: Is it worth not dying here?  I forget whether this.error() aborts the compilation or not.  Might be worth just emitting some `/* TODO: handle ... */`.

test_files/abstract/abstract.js
```diff
@@ -7,41 +7,43 @@ goog.module('test_files.abstract.abstract');var module = module || {id: 'test_fi
...
+    Base.prototype.simple;
     /**
      * @abstract
      * @return {void}
      */
-    publicAbstract() { }
+    Base.prototype.publicAbstract;
     /**
      * @abstract
      * @param {!Array<number>} x
```
evmar: Huh, didn't think you could @param on a property, neat.

test_files/class/class.tsickle.ts
```diff
@@ -35,16 +35,22 @@ classFunc(): void {}
...
+  abstract abstractFunc(): void;
 /**
- * @abstract
  * @return {void}
  */
-abstractFunc() {}
+nonAbstractFunc(): void { }
+}
+
+function AbstractClass_tsickle_Closure_declarations() {
```
mhevery: Can we emit 
```
/** @__PURE__ */
function AbstractClass_tsickle_Closure_declarations() {
```

This would greatly help uglify.

test_files/class/class.js
```diff
@@ -19,14 +19,16 @@ class Class {
  */
 class AbstractClass {
     /**
-     * @abstract
      * @return {void}
      */
-    abstractFunc() { }
+    nonAbstractFunc() { }
+}
+function AbstractClass_tsickle_Closure_declarations() {
```
mhevery: Can we emit 
```
/** @__PURE__ */
function AbstractClass_tsickle_Closure_declarations() {
```

This would greatly help uglify.

test_files/abstract/abstract.tsickle.ts
```diff
@@ -8,43 +8,61 @@
...
+bar() {
+    this.simple();
+    this.publicAbstract();
+    this.params([]);
+    this.noReturnType();
+    this.hasReturnType();
+  }
+}
+
+function Base_tsickle_Closure_declarations() {
```
mhevery: Can we emit 
```
/** @__PURE__ */
function AbstractClass_tsickle_Closure_declarations() {
```

This would greatly help uglify.

test_files/abstract/abstract.js
```diff
@@ -7,41 +7,43 @@ goog.module('test_files.abstract.abstract');var module = module || {id: 'test_fi
...
+     */
+    bar() {
+        this.simple();
+        this.publicAbstract();
+        this.params([]);
+        this.noReturnType();
+        this.hasReturnType();
+    }
+}
+function Base_tsickle_Closure_declarations() {
```
mhevery: Can we emit 
```
/** @__PURE__ */
function AbstractClass_tsickle_Closure_declarations() {
```

This would greatly help uglify.

mhevery:
Related: https://github.com/angular/tsickle/issues/505

test_files/abstract/abstract.tsickle.ts
mprobst: That's another change. Does `@__PURE__` have an effect on declarations though, as opposed to expressions? I thought it was to mark expressions as no side-effects.

src/tsickle.ts
mprobst: Yeah, that's cleaner. Done.

src/tsickle.ts
mprobst: This matches previous behaviour. I think it's OK to leave it like this, as long as it doesn't break google3 I'd rather be stricter.

mprobst:
@evmar mea maxima culpa on mixing changes. I discovered the "fix all occurrences" feature of tslint in VSCode, and then it was all messed up with my commit.

mprobst:
I'll merge tomorrow once I get around to testing this in google3.

mprobst:
Green in google3.

mprobst:
@evmar @mhevery PTAL. I had to change the emit slightly (assign functions), and updated our test suite to run with stricter warnings.

test_files/abstract/abstract.tsickle.ts
mhevery: Actually, let me retract my statement. The `@__PURE__` is not needed at this location. The test I did for it was flawed. As long as there is no one referring to `Base_tsickle_CLosure_declarations` it will tree shake away.

mhevery:
@mprobst It seems like your recent change added bodies to the abstract methods and moved them from instance to prototype. I have verified that it still tree shakes properly in Uglify, so I approve this change.

evmar:
This PR now does even more unrelated things.  Can you at least put https://github.com/angular/tsickle/pull/506/commits/67e1ef4c6d28b9b5d6492338854a05ddd3904868 in a separate CL?

mprobst:
@evmar it does not. `@abstract` is actually illegal on a property (as opposed to function), but Closure Compiler doesn't warn unless we pass the verbose warnings flag (which is its own little wtf - the compiler doesn't warn about incorrect syntax directed at it in its default mode?!?). So to properly test the code I wrote here, I need that commit, which is why it is in this PR.

Given you don't seem to have comments on the code, I'll merge.

PR #512, opened 2017-06-08
feat: support computed property keys in externs.

### LINKED ISSUE #223: Tsickle generates invalid externs on [Symbols]
``` js
interface Symbol {
  [Symbol.toStringTag]: 'Symbol';
}
```

produces invalid externs.

mprobst:
Fixes #223.


.gitignore
```diff
@@ -1,5 +1,6 @@
 # Compiled JS
 /built
+/build
```
alexeagle: you should delete your /build folder - it's out of date
on Mac you would have been forced to do so, because the BUILD file is the same file - if you submit this change then Mac users will be ignoring BUILD which is wrong

src/tsickle.ts
```diff
@@ -1547,6 +1547,7 @@ class ExternsWriter extends ClosureRewriter {
         case ts.SyntaxKind.MethodSignature:
         case ts.SyntaxKind.MethodDeclaration:
           const method = member as ts.MethodDeclaration;
+          // if (method.name.kind !== ts.SyntaxKind.Identifier) break;
```
alexeagle: delete the comment?

.gitignore
mprobst: Done.

src/tsickle.ts

PR #513, opened 2017-06-08
feat: support heritance clauses in externs.

mprobst:
This is a roll forward of a previous attempt, now testing against g3.


mprobst:
(this is currently not correct and blocked on fixing)

mprobst:
Ready to review.

src/tsickle.ts
```diff
@@ -385,7 +385,8 @@ class ClosureRewriter extends Rewriter {
     for (const heritage of decl.heritageClauses!) {
       if (!heritage.types) continue;
       if (decl.kind === ts.SyntaxKind.ClassDeclaration &&
-          heritage.token !== ts.SyntaxKind.ImplementsKeyword) {
+          heritage.token !== ts.SyntaxKind.ImplementsKeyword &&
+          !hasModifierFlag(decl, ts.ModifierFlags.Ambient)) {
```
evmar: I was confused by this because I had forgotten that this loop is collecting up jsdoc.  I think the reason for this extra check here is that externs do not use ES6 classes, right?  Maybe put a comment at the top of the loop or here?

src/tsickle.ts
```diff
@@ -106,6 +106,16 @@ function hasModifierFlag(node: ts.Node, flag: ts.ModifierFlags): boolean {
   return (ts.getCombinedModifierFlags(node) & flag) !== 0;
 }
```
evmar: This patch's description describes what it does but not why.  It's not clear to me why it's needed.

mprobst: Done. In a nutshell, TypeTranslator has a map of import aliases, symbol->string. If we resolve aliases before passing them to TypeTranslator, its local alias map is defeated as it only contains the original, non-dereferenced symbol.

src/tsickle.ts

PR #514, opened 2017-06-13
Prefactorings for introducing transformers

tbosch:


src/source_map_utils.ts
```diff
@@ -96,3 +97,47 @@ export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenera
...
+
+export interface SourceMapper {
+  addMapping(
+      originalNode: ts.Node, original: SourcePosition, generated: SourcePosition,
+      length: number): void;
+}
+
+export const NOOP_SOURCE_MAPPER: SourceMapper = {
+  // tslint:disable-next-line:no-empty
+  addMapping: () => {}
```
evmar: addMapping() {}

?

src/tsickle_compiler_host.ts
```diff
@@ -30,56 +30,27 @@ export enum Pass {
...
-  typeBlackListPaths?: Set<string>;
-  /**
-   * Convert shorthand "/index" imports to full path (include the "/index").
-   * Annotation will be slower because every import must be resolved.
-   */
-  convertIndexImportShorthand?: boolean;
+export interface Options extends Es5ProcessorOptions, tsickle.AnnotatorOptions {
+  // This method is here for backwards compatibility.
+  // Use the method in TsickleHost instead.
+  logWarning?: TsickleHost['logWarning'];
```
evmar: I think quoted accesses like this can break optimization.  Why not TsickleHost.logWarning?

src/decorator-annotator.ts
```diff
@@ -159,128 +130,149 @@ class ClassRewriter extends Rewriter {
...
-          return true;
-        }
-        return false;
+        break;
       default:
-        return false;
+    }
+  }
+
+  maybeProcessDecorator(node: ts.Node, start?: number): boolean {
```
evmar: Shouldn't this take a ts.Decorator?  And then drop the "as ts.Decorator" below?

src/tsickle_compiler_host.ts
```diff
@@ -27,7 +27,8 @@ import {isDtsFileName} from './tsickle';
 export enum Pass {
   NONE,
   DECORATOR_DOWNLEVEL,
-  CLOSURIZE
+  CLOSURIZE,
+  DECORATOR_DOWNLEVEL_AND_CLOSURIZE,
```
evmar: Can you add a comment here about how you do both in a single pass, and also that we want to drop the other ones?

src/tsickle_compiler_host.ts
```diff
@@ -138,7 +139,10 @@ export class TsickleCompilerHost implements ts.CompilerHost {
             sourceFile, this.runConfiguration.oldProgram, fileName, languageVersion);
       case Pass.CLOSURIZE:
         return this.closurize(
-            sourceFile, this.runConfiguration.oldProgram, fileName, languageVersion);
+            sourceFile, this.runConfiguration.oldProgram, fileName, languageVersion, false);
+      case Pass.DECORATOR_DOWNLEVEL_AND_CLOSURIZE:
+        return this.closurize(
+            sourceFile, this.runConfiguration.oldProgram, fileName, languageVersion, true);
```
evmar: Can you stick in some comment like
```
..., languageVersion, /* also downlevel */ true)
``` 

and also above in the false case?

src/tsickle_compiler_host.ts
tbosch: This is for reusing the type, not the value.

src/source_map_utils.ts
tbosch: Sounds good.

src/tsickle_compiler_host.ts
tbosch: Sounds good.

src/tsickle_compiler_host.ts
tbosch: Sounds good.

src/decorator-annotator.ts
tbosch: Sounds good.

tbosch:
Attention: This breaks G3, probably a mistake in the enum changes. Need to investigate...

tbosch:
Deleted the commit that changed enums as the following reports the error "Property 'p' of exported interface has or is using private name 'MyEnum'

```
export interface I {
  p: MyEnum;
}


const MyEnum = {};
type MyEnum = number;
export {MyEnum}
```

Note that if the variable / type `MyEnum` is declared _before_ the interface this works!

PR #515, opened 2017-06-14
ts2.4 prework

evmar:
This series of commits makes adjustments to the code that are required for TS2.4, but that still work in TS2.3.  So they are safe to land on master as-is, without changing our TS version.

You can review each commit individually, might be easier.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -223,6 +223,17 @@ class ClosureRewriter extends Rewriter {
   }
 
   /**
+   * Get the ts.Symbol at a location or throw.
+   * The TypeScript API can return undefined when fetching a symbol, but
+   * in many contexts we know it won't (e.g. our input is already type-checked).
+   */
+  mustGetSymbolAtLocation(node: ts.Node): ts.Symbol {
+    const sym = this.typeChecker.getSymbolAtLocation(node);
+    if (!sym) throw new Error('no symbol');
+    return sym;
+  }
+
+  /**
    * Handles emittng the jsdoc for methods, including overloads.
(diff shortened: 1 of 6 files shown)


evmar:
@rkirov for review, @tbosch just FYI

src/tsickle.ts
```diff
@@ -188,6 +188,17 @@ class ClosureRewriter extends Rewriter {
   }
 
   /**
+   * Get the ts.Symbol at a location or throw.
+   * The TypeScript API can return undefined when fetching a symbol, but
+   * in many contexts we know it won't (e.g. our input is already type-checked).
+   */
+  mustGetSymbolAtLocation(node: ts.Node): ts.Symbol {
+    const sym = this.typeChecker.getSymbolAtLocation(node);
+    if (!sym) throw new Error('no symbol');
```
rkirov: can you throw in some `node` info in the Error.

src/tsickle.ts
```diff
@@ -256,6 +258,7 @@ class ClosureRewriter extends Rewriter {
       }
       // Merge the parameters into a single list of merged names and list of types
       const sig = typeChecker.getSignatureFromDeclaration(fnDecl);
+      if (!sig) throw new Error('invalid signature');
```
rkirov: add 'fnDecl' name to the error.

src/tsickle.ts
evmar: If we throw while traversing the AST, we show the file, line:col.

src/tsickle.ts
evmar: Done

PR #516, opened 2017-06-14
Introduce transformer support for tsickle

tbosch:
This PR introduces transformer support for tsickle.
The first 6 commits are prefactorings, which are used by the last commit which introduces the main API.

Global TAP presubmit is looking good using the new transformer API in G3.

Proposed way of landing this PR:
1. Review this PR and land it (hopefully mostly as is) on master
2. Continue to use the old entry point in G3 which is still fully functional
3. Use the new entry point in G3.

This way, we can have a bit of time between 2. and 3., without blocking other fixes for Tsickle. We are also able to rollback to using the old entry point without creating branches of Tsickle.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -22,13 +22,14 @@ export class DecoratorClassVisitor {
   /** Decorators on the class itself. */
   decorators: ts.Decorator[];
   /** The constructor parameter list and decorators on each param. */
-  ctorParameters: Array<[string | undefined, ts.Decorator[]|undefined]|null>;
+  ctorParameters: Array<[[ts.Symbol, string] | undefined, ts.Decorator[]|undefined]|null>;
   /** Per-method decorators. */
   propDecorators: Map<string, ts.Decorator[]>;
 
   constructor(
       private typeChecker: ts.TypeChecker, private rewriter: Rewriter,
-      private classDecl: ts.ClassDeclaration) {
+      private classDecl: ts.ClassDeclaration,
+      private importedNames: Array<{name: ts.Identifier, declarationNames: ts.Identifier[]}>) {
     if (classDecl.decorators) {
       const toLower = this.decoratorsToLower(classDecl);
       if (toLower.length > 0) this.decorators = toLower;
@@ -80,10 +81,11 @@ export class DecoratorClassVisitor {
    * constructor, and emits nothing.
    */
   private gatherConstructor(ctor: ts.ConstructorDeclaration) {
-    const ctorParameters: Array<[string | undefined, ts.Decorator[] | undefined]|null> = [];
+    const ctorParameters:
+        Array<[[ts.Symbol, string] | undefined, ts.Decorator[] | undefined]|null> = [];
     let hasDecoratedParam = false;
     for (const param of ctor.parameters) {
-      let paramCtor: string|undefined;
+      let paramCtor: [ts.Symbol, string]|undefined;
       let decorators: ts.Decorator[]|undefined;
       if (param.decorators) {
         decorators = this.decoratorsToLower(param);
@@ -94,8 +96,9 @@ export class DecoratorClassVisitor {
         // Verify that "Bar" is a value (e.g. a constructor) and not just a type.
         const sym = this.typeChecker.getTypeAtLocation(param.type).getSymbol();
         if (sym && (sym.flags & ts.SymbolFlags.Value)) {
-          paramCtor = new TypeTranslator(this.typeChecker, param.type)
-                          .symbolToString(sym, /* useFqn */ true);
+          const typeStr = new TypeTranslator(this.typeChecker, param.type)
+                              .symbolToString(sym, /* useFqn */ true);
+          paramCtor = [sym, typeStr];
         }
       }
       if (paramCtor || decorators) {
@@ -128,6 +131,26 @@ export class DecoratorClassVisitor {
     if (decorators.length === 0) return;
     if (!this.propDecorators) this.propDecorators = new Map<string, ts.Decorator[]>();
     this.propDecorators.set(name, decorators);
+  }
+
+  private getValueIdentifierForType(typeSymbol: ts.Symbol): ts.Identifier|null {
+    if (!typeSymbol.valueDeclaration) {
+      return null;
+    }
+    const valueName = typeSymbol.valueDeclaration.name;
+    if (!valueName || valueName.kind !== ts.SyntaxKind.Identifier) {
+      return null;
+    }
+    if (valueName.getSourceFile() === this.rewriter.file) {
+      return valueName;
+    }
+    for (let i = 0; i < this.importedNames.length; i++) {
+      const {name, declarationNames} = this.importedNames[i];
+      if (declarationNames.some(d => d === valueName)) {
+        return name;
+      }
+    }
+    return null;
   }
 
   beforeProcessNode(node: ts.Node) {
@@ -211,7 +234,22 @@ export class DecoratorClassVisitor {
           continue;
         }
         const [ctor, decorators] = param;
-        this.rewriter.emit(`{type: ${ctor}, `);
+        this.rewriter.emit(`{type: `);
+        if (!ctor) {
+          this.rewriter.emit(`undefined`);
+        } else {
+          const [typeSymbol, typeStr] = ctor;
+          let emitNode: ts.Identifier|null|undefined;
+          if (typeSymbol) {
+            emitNode = this.getValueIdentifierForType(typeSymbol);
+          }
+          if (emitNode) {
+            this.rewriter.writeRange(emitNode, emitNode.getStart(), emitNode.getEnd());
+          } else {
+            this.rewriter.emit(typeStr);
+          }
+        }
+        this.rewriter.emit(`, `);
         if (decorators) {
           this.rewriter.emit('decorators: [');
           for (const decorator of decorators) {
@@ -256,7 +294,7 @@ export class DecoratorClassVisitor {
         if (call.arguments.length) {
           this.rewriter.emit(', args: [');
           for (const arg of call.arguments) {
-            this.rewriter.emit(arg.getText());
+            this.rewriter.writeNodeFrom(arg, arg.getStart());
             this.rewriter.emit(', ');
           }
           this.rewriter.emit(']');
@@ -294,7 +332,7 @@ class DecoratorRewriter extends Rewriter {
       case ts.SyntaxKind.ClassDeclaration:
         const oldDecoratorConverter = this.currentDecoratorConverter;
         this.currentDecoratorConverter =
-            new DecoratorClassVisitor(this.typeChecker, this, node as ts.ClassDeclaration);
+            new DecoratorClassVisitor(this.typeChecker, this, node as ts.ClassDeclaration, []);
         this.writeRange(node, node.getFullStart(), node.getStart());
         visitClassContentIncludingDecorators(
             node as ts.ClassDeclaration, this, this.currentDecoratorConverter);
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -133,6 +133,18 @@ export class DecoratorClassVisitor {
     this.propDecorators.set(name, decorators);
   }
 
+  /**
+   * For lowering decorators, we need to refer to constructor types.
+   * So we start with the identifiers that represent these types.
+   * However, TypeScript does not allow use to emit them in a value position
+   * as it associated different symbol information with it.
+   *
+   * This method looks for the place where the value that is associated to
+   * the type is defined and returns that identifier instead.
+   *
+   * @param typeSymbol
+   * @return The identifier
+   */
   private getValueIdentifierForType(typeSymbol: ts.Symbol): ts.Identifier|null {
     if (!typeSymbol.valueDeclaration) {
       return null;
@@ -348,6 +360,7 @@ export function visitClassContentIncludingDecorators(
     classDecl: ts.ClassDeclaration, rewriter: Rewriter, decoratorVisitor?: DecoratorClassVisitor) {
   if (rewriter.file.text[classDecl.getEnd() - 1] !== '}') {
     rewriter.error(classDecl, 'unexpected class terminator');
+    return;
   }
   rewriter.writeNodeFrom(classDecl, classDecl.getStart(), classDecl.getEnd() - 1);
   // At this point, we've emitted up through the final child of the class, so all that
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -317,40 +317,15 @@ class DecoratorRewriter extends Rewriter {
 
 export function visitClassContentIncludingDecorators(
     classDecl: ts.ClassDeclaration, rewriter: Rewriter, decoratorVisitor?: DecoratorClassVisitor) {
-  let pos = classDecl.getStart();
-  if (decoratorVisitor) {
-    // visit decorators if needed
-    ts.forEachChild(classDecl, child => {
-      if (child.kind !== ts.SyntaxKind.Decorator) {
-        return;
-      }
-      // Note: The getFullStart() of the first decorator is the same
-      // as the getFullStart() of the class declaration.
-      // Therefore, we need to use Math.max to not print the whitespace
-      // of the class again.
-      const childStart = Math.max(pos, child.getFullStart());
-      rewriter.writeRange(classDecl, pos, childStart);
-      if (decoratorVisitor.maybeProcessDecorator(child as ts.Decorator, childStart)) {
-        pos = child.getEnd();
-      }
-    });
+  if (rewriter.file.text[classDecl.getEnd() - 1] !== '}') {
+    rewriter.error(classDecl, 'unexpected class terminator');
   }
-  if (classDecl.members.length > 0) {
-    rewriter.writeRange(classDecl, pos, classDecl.members[0].getFullStart());
-    for (const member of classDecl.members) {
-      rewriter.visit(member);
-    }
-    pos = classDecl.getLastToken().getFullStart();
-  }
+  rewriter.writeNodeFrom(classDecl, classDecl.getStart(), classDecl.getEnd() - 1);
   // At this point, we've emitted up through the final child of the class, so all that
   // remains is the trailing whitespace and closing curly brace.
   // The final character owned by the class node should always be a '}',
   // or we somehow got the AST wrong and should report an error.
   // (Any whitespace or semicolon following the '}' will be part of the next Node.)
-  if (rewriter.file.text[classDecl.getEnd() - 1] !== '}') {
-    rewriter.error(classDecl, 'unexpected class terminator');
-  }
-  rewriter.writeRange(classDecl, pos, classDecl.getEnd() - 1);
   if (decoratorVisitor) {
     decoratorVisitor.emitMetadataAsStaticProperties();
   }
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -205,19 +205,10 @@ export class DecoratorClassVisitor {
       this.rewriter.emit(
           `static ctorParameters: () => ({type: any, decorators?: ` + decoratorInvocations +
           `}|null)[] = () => [\n`);
-      let emittedInline = false;
       for (const param of this.ctorParameters || []) {
         if (!param) {
-          if (emittedInline) {
-            this.rewriter.emit(' ');
-          }
-          this.rewriter.emit('null,');
-          emittedInline = true;
+          this.rewriter.emit('null,\n');
           continue;
-        }
-        if (emittedInline) {
-          this.rewriter.emit('\n');
-          emittedInline = false;
         }
         const [ctor, decorators] = param;
         this.rewriter.emit(`{type: ${ctor}, `);
--- a/src/rewriter.ts
+++ b/src/rewriter.ts
@@ -27,10 +27,10 @@ export abstract class Rewriter {
    */
   private indent = 0;
   /**
-   * Skip emitting any code before the given offset. Used to avoid emitting @fileoverview comments
-   * twice.
+   * Skip emitting any code before the given offset. E.g. used to avoid emitting @fileoverview
+   * comments twice.
    */
-  public skipUpToOffset = 0;
+  private skipUpToOffset = -1;
 
   constructor(public file: ts.SourceFile, private sourceMapper: SourceMapper = NOOP_SOURCE_MAPPER) {
   }
@@ -84,13 +84,19 @@ export abstract class Rewriter {
     this.writeNodeFrom(node, pos);
   }
 
-  writeNodeFrom(node: ts.Node, pos: number) {
+  writeNodeFrom(node: ts.Node, pos: number, end = Number.MAX_SAFE_INTEGER) {
+    if (node.getEnd() <= this.skipUpToOffset) {
+      return;
+    }
+    const oldSkipUpToOffset = this.skipUpToOffset;
+    this.skipUpToOffset = Math.max(this.skipUpToOffset, pos);
     ts.forEachChild(node, child => {
       this.writeRange(node, pos, child.getFullStart());
       this.visit(child);
       pos = child.getEnd();
     });
-    this.writeRange(node, pos, node.getEnd());
+    this.writeRange(node, pos, Math.min(end, node.getEnd()));
+    this.skipUpToOffset = oldSkipUpToOffset;
   }
 
   /**
@@ -100,7 +106,7 @@ export abstract class Rewriter {
    */
   writeRange(node: ts.Node, from: number, to: number) {
     from = Math.max(from, this.skipUpToOffset);
-    if (this.skipUpToOffset > 0 && to <= this.skipUpToOffset) {
+    if (from !== to && to <= this.skipUpToOffset) {
       return;
     }
     // Add a source mapping. writeRange(from, to) always corresponds to
--- a/src/rewriter.ts
+++ b/src/rewriter.ts
@@ -84,8 +84,8 @@ export abstract class Rewriter {
     this.writeNodeFrom(node, pos);
   }
 
-  writeNodeFrom(node: ts.Node, pos: number, end = Number.MAX_SAFE_INTEGER) {
-    if (node.getEnd() <= this.skipUpToOffset) {
+  writeNodeFrom(node: ts.Node, pos: number, end = node.getEnd()) {
+    if (end <= this.skipUpToOffset) {
       return;
     }
     const oldSkipUpToOffset = this.skipUpToOffset;
@@ -95,7 +95,7 @@ export abstract class Rewriter {
       this.visit(child);
       pos = child.getEnd();
     });
-    this.writeRange(node, pos, Math.min(end, node.getEnd()));
+    this.writeRange(node, pos, end);
     this.skipUpToOffset = oldSkipUpToOffset;
   }
 
--- a/src/transformer.ts
+++ b/src/transformer.ts
@@ -1,0 +1,222 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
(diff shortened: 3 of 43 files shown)


tbosch:
Note: This still has some G3 failures, I am working on them. So it's probably better to wait with the review of my PR until I am really sure G3 is green.

src/decorator-annotator.ts
```diff
@@ -317,40 +317,15 @@ class DecoratorRewriter extends Rewriter {
...
-    });
-  }
-  if (classDecl.members.length > 0) {
-    rewriter.writeRange(classDecl, pos, classDecl.members[0].getFullStart());
-    for (const member of classDecl.members) {
-      rewriter.visit(member);
-    }
-    pos = classDecl.getLastToken().getFullStart();
+  if (rewriter.file.text[classDecl.getEnd() - 1] !== '}') {
+    rewriter.error(classDecl, 'unexpected class terminator');
```
evmar: Should we "return" here to avoid the rest of the function?

src/rewriter.ts
```diff
@@ -84,13 +84,19 @@ export abstract class Rewriter {
     this.writeNodeFrom(node, pos);
   }
 
-  writeNodeFrom(node: ts.Node, pos: number) {
+  writeNodeFrom(node: ts.Node, pos: number, end = Number.MAX_SAFE_INTEGER) {
```
evmar: Can write:

writeNodeFrom(node: ts.Node, pos: number, end = node.getEnd()) {

and then drop the Math.min() below as well as calling node.getEnd().

src/tsickle.ts
```diff
@@ -659,7 +660,11 @@ class Annotator extends ClosureRewriter {
         // Only emit a type annotation when it's a plain variable and
         // not a binding pattern, as Closure doesn't(?) have a syntax
         // for annotating binding patterns.  See issue #128.
-        if (varDecl.name.kind === ts.SyntaxKind.Identifier) {
+        // Don't emit type annotation when the variable statement is a @polymerBehavior,
+        // as otherweise the polymer closure checker will fail.
```
evmar: typo: otherwise

src/tsickle.ts
```diff
@@ -659,7 +660,11 @@ class Annotator extends ClosureRewriter {
...
+        // TODO(tbosch): file an issue for this!
```
evmar: Fix this with a link to the issue?

src/tsickle.ts
```diff
@@ -745,11 +761,18 @@ class Annotator extends ClosureRewriter {
           typeCopy.types = nonNullUnion;
           type = typeCopy;
         }
+        // See comment above.
+        if (this.templateSpanStackCount > 0) {
+          this.emit('(');
+        }
         this.emitJSDocType(nnexpr, undefined, type);
         // See comment above.
         this.emit('((');
```
evmar: I vaguely suspect that this double-paren is trying to work around the same(?) issue you are.  It's pretty weird that we emit *three* layers of parens here.  It's maybe not worth trying to undo in your change, but maybe you could add a TODO?

test_files/decorator/decorator.decorated.ts
```diff
@@ -17,7 +17,7 @@ function classAnnotation(t: any) { return t; }
 
 
 class DecoratorTest {
-  constructor(a: any[], n: number, b: boolean, promise: Promise<string>, arr: Array<string>, aClass: AClass, aClassWithGenerics: AClassWithGenerics<string>, aType: AType) {}
+  constructor(a: any[], n: number, b: boolean, promise: Promise<string>, arr: Array<string>, aClass: AClass, AClass: AClass, aRanmedClass: ARenamedClass, aClassWithGenerics: AClassWithGenerics<string>, aType: AType) {}
```
evmar: typo: "renamed"

(doesn't matter too much here, but if it's not hard to fix...)

src/decorator-annotator.ts
```diff
@@ -130,6 +133,26 @@ export class DecoratorClassVisitor {
     this.propDecorators.set(name, decorators);
   }
 
+  private getValueIdentifierForType(typeSymbol: ts.Symbol): ts.Identifier|null {
```
evmar: Can you add docs on this?  What is a value identifier (or value declaration)? Something about literal types?

src/tsickle.ts
```diff
@@ -50,8 +51,43 @@ export interface AnnotatorOptions {
 
 export enum AnnotatorFeatures {
   LowerDecorators = 1 << 0,
+  /**
+   * filter out types when expanding `export * from ...`
+   */
+  FilterTypesInExportStart = 1 << 1,
+  /**
+   * emit enums as
```
evmar: Can you comment here why you'd want to pass this flag or not?  I recall one of these is a temporary thing that we intend to remove.  I see below you have the "transformers need: ..." block, but is the intent that we remove all of these flags?

src/tsickle.ts
```diff
@@ -1088,13 +1158,47 @@ class Annotator extends ClosureRewriter {
...
       this.emit(` from '${importPath}';`);
+      this.writeRange(decl, decl.getEnd(), decl.getEnd());
       return true;
     } else {
       this.errorUnimplementedKind(decl, 'unexpected kind of import');
       return false;  // Use default processing.
     }
   }
 
+  private collectImportNames(decl: ts.ImportDeclaration) {
```
evmar: Can you document this?

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as ts from 'typescript';
+import * as tsickle from './tsickle';
+
+/**
+ * Adjusts the given CustomTransformers with additional transformers
+ * to fix bugs in TypeScript.
+ * @param given A
```
evmar: XXX finish comment here

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  }
+  if (context.fileContext.file.fileName !== sourceFile.fileName) {
+    throw new Error(
+        `Illegal State: File of the FileContext does not match. File: ${sourceFile.fileName}`);
+  }
+  return context.fileContext;
+}
+
+/**
+ * And extended version of the TransformationContext that stores the FileContext as well.
```
evmar: Typo: "an"

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  }
+  const trailingComments = ts.getTrailingCommentRanges(sourceFile.text, node.getEnd());
+  if (trailingComments && trailingComments.length) {
+    ts.setSyntheticTrailingComments(node, synthesizeCommentRanges(sourceFile, trailingComments));
+    return trailingComments[trailingComments.length - 1].end;
+  }
+  return -1;
+}
+
+function visitNodeStatementsWithSynthesizedComments<T extends ts.Node>(
```
evmar: docs

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+    node = visitor(node, statements);
+    if (trailing.lastCommentEnd !== -1) {
+      fileContext.lastCommentEnd = trailing.lastCommentEnd;
+    }
+    return node;
+  }
+  return visitor(node, statements);
+}
+
+function synthesizeDetachedLeadingComments(
```
evmar: docs (here and in the rest of the file)

src/transformer.ts
```diff
@@ -0,0 +1,222 @@
...
+      }
+      if (diagnostics) {
+        tsickleDiagnostics.push(...diagnostics);
+      }
+    });
+  }
+  // All diagnostics (including warnings) are treated as errors.
+  // If we've decided to ignore them, just discard them.
+  // Warnings include stuff like "don't use @type in your jsdoc"; tsickle
+  // warns and then fixes up the code to be Closure-compatible anyway.
```
evmar: The rule here is that we error on google3 code, and don't error on third-party code.  Is that still preserved here?

src/transformer_sourcemap.ts
```diff
@@ -0,0 +1,144 @@
...
+          }
+        }
+      }
+
+      return node;
+    }
+  };
+}
+
+class NodeSourceMapper implements SourceMapper {
```
evmar: docs on here and the rest of this file

I think maybe this one could use a @fileOverview, because this is the one that does the clever "map node offsets to nodes" trick

test/tsickle_transformer_test.ts
```diff
@@ -0,0 +1,225 @@
...
+  return golden;
+}
+
+function patchGolden(golden: string|null, path: string): string|null {
+  if (golden == null) {
+    return golden;
+  }
+  const patchPath = calcPatchPath(path);
+  if (fs.existsSync(patchPath)) {
+    // Note: the typings for `diff.applyPath` are wrong in that the function
```
evmar: typo: patch

src/transformer.ts
tbosch: Yes, via the `host.shouldIgnoreWarningsForPath`.

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  let allowTextRange = node.kind !== ts.SyntaxKind.ClassDeclaration &&
+      node.kind !== ts.SyntaxKind.ExportDeclaration &&
+      node.kind !== ts.SyntaxKind.ImportDeclaration &&
+      node.kind !== ts.SyntaxKind.VariableDeclaration &&
+      !(node.kind === ts.SyntaxKind.VariableStatement &&
+        tsickle.hasModifierFlag(node, ts.ModifierFlags.Export));
+  if (node.kind === ts.SyntaxKind.PropertyDeclaration) {
+    allowTextRange = false;
+    const pd = node as ts.Node as ts.PropertyDeclaration;
+    // TODO(tbosch): Using pd.initializer! as the typescript typings for intializer are incorrect.
```
tbosch: I think I will leave `TODO`s everywhere for now where I need to file a bug, as that takes a bit (with small reproduction, ...)

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+    if (predicate(node)) {
+      return node;
+    }
+  }
+  return null;
+}
+
+/**
+ * Synthesizes the comments before and after a node
+ * befor calling a callback.
```
tbosch: By default, comments are just text ranges (i.e. from - to). To synthesize means to create a node that is independent of text ranges and contains the information in itself.

For comments, we go from `ts.CommentRange` to `ts.SynthesizedComment`. The latter contains a field `text` to store the actual comment.

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+      nodePath.pop();
+    }
+  };
+}
+
+/**
+ * Transform that needs to be executed after TypeScript's transform.
+ *
+ * This fixes places where the TypeScript transformer does not
+ * emit synthetic comments.
```
tbosch: Yes. I think I will leave TODOs everywhere for now where I need to file a bug, as that takes a bit (with small reproduction, ...)

src/tsickle.ts
tbosch: We still need the double paren thing for same cases sadly (I saw it via the golden tests) :-(

src/tsickle.ts
tbosch: This is not temporary.

TSC works as this:
1. parse and type check / create symbol table
2. run transformers
3. emit

If we are expanding `*` into concrete identifiers, TypeScript has no information about them, and therefore cannot elide types from them. So we have to do this.
In the version of tsickle where we create a new pogram, TypeScript will recreate the symbol table and therefore know which types to elide.

So for transformers, we always want to do this.
Without transformers, we must not do this, as we rely on the fact that we produce functions for interfaces, which should not be elided.

src/tsickle.ts
tbosch: I think for now I will keep the comment as filing an issue takes some time...

tbosch: Actually, I think I should file an issue against tsickle to create issues for typescript. This way we don't forget to do this.

src/decorator-annotator.ts
tbosch: For lowering decorators, we need to refer to constructor types. So we start off with the identifiers that represent these types. However, we want to emit them in a value position. TypeScript does not allow to use the same ts.Identifier here. Instead, this method looks for the place where the value that is associated to the type is defined and returns that identifier instead.

src/decorator-annotator.ts
tbosch: Done

src/decorator-annotator.ts
tbosch: Added

src/rewriter.ts
tbosch: Done

src/tsickle.ts
tbosch: Done

src/tsickle.ts
tbosch: Created https://github.com/angular/tsickle/issues/529 to track this and linked in the code.

src/tsickle.ts
tbosch: Added all information about the flags directly to the enum members, and documented that these flags will be removed and on by default once the non transformer version is dropped.

src/tsickle.ts
tbosch: Updated the docs there.

src/tsickle.ts
tbosch: Done

test_files/decorator/decorator.decorated.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+          const expressionStmt =
+              lastNodeWith(nodePath, (node) => node.kind === ts.SyntaxKind.ExpressionStatement);
+          if (expressionStmt) {
+            ts.setSyntheticLeadingComments(
+                expressionStmt, ts.getSyntheticLeadingComments(parent1) || []);
+          }
+        } else if (
+            parent3 && parent3.kind === ts.SyntaxKind.VariableStatement &&
+            tsickle.hasModifierFlag(parent3, ts.ModifierFlags.Export)) {
+          // TypeScript ignore synthetic comments on exported variables.
```
tbosch: Done

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+          // find the parent ExpressionStatement like exports.foo = ...
+          const expressionStmt =
+              lastNodeWith(nodePath, (node) => node.kind === ts.SyntaxKind.ExpressionStatement);
+          if (expressionStmt) {
+            ts.setSyntheticLeadingComments(
+                expressionStmt, ts.getSyntheticLeadingComments(parent3) || []);
+          }
+        }
+      }
+      // TypeScript ignore synthetic comments on reexport / import statements.
```
tbosch: Done

src/transformer_util.ts
tbosch: Added a comment.

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+      node = getMutableClone(node);
+    }
+    const textRange = {pos: node.pos, end: node.end};
+    ts.setSourceMapRange(node, textRange);
+    ts.setTextRange(node, {pos: -1, end: -1});
+    return node;
+  }
+}
+
+function synthesizeLeadingComments(
```
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer.ts
tbosch: Updated the comment.

test/tsickle_transformer_test.ts
tbosch: Done

src/transformer_sourcemap.ts
tbosch: Done

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  syntheticNodeParents = new Map<ts.Node, ts.Node|undefined>();
+  importOrReexportDeclarations: Array<ts.Node&{moduleSpecifier: ts.StringLiteral}> = [];
+  lastCommentEnd = -1;
+  constructor(public file: ts.SourceFile) {}
+}
+
+/**
+ * Transform that needs to be executed right before TypeScript's transform.
+ *
+ * This prepares the node tree to workaround some bugs in the TypeScript emitter.
```
tbosch: Created https://github.com/angular/tsickle/issues/528 to track filing these bugs.

src/transformer_util.ts
tbosch: Created https://github.com/angular/tsickle/issues/528 to track filing these bugs.

src/tsickle.ts
```diff
@@ -1088,13 +1158,47 @@ class Annotator extends ClosureRewriter {
...
+      names.push(...namedImports.elements.map(e => e.name));
+    }
+    names.forEach(name => {
+      let symbol = this.typeChecker.getSymbolAtLocation(name);
+      if (symbol.flags & ts.SymbolFlags.Alias) {
+        symbol = this.typeChecker.getAliasedSymbol(symbol);
+      }
+      const declarationNames: ts.Identifier[] = [];
+      if (symbol.declarations) {
+        symbol.declarations.forEach(d => {
```
mprobst: as above

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  before.push(preTypeScriptTransform);
+  const after = given.after || [];
+  after.unshift(postTypescriptTransform);
+  return {before, after};
+}
+
+/**
+ * Transform to be used as first transform to reset some state.
+ */
+function firstTransform(context: ts.TransformationContext) {
```
mprobst: can you give this semantic names, e.g. `addFileContexts`?

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  return (sourceFile: ts.SourceFile) => {
+    (context as TransformationContext).fileContext = new FileContext(sourceFile);
```
mprobst: Document that all transformers run in sequence for each file before visiting the next file, which makes this operation (caching the source file here) valid.

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+ * And extended version of the TransformationContext that stores the FileContext as well.
+ */
+interface TransformationContext extends ts.TransformationContext {
+  fileContext?: FileContext;
+}
+
+/**
+ * A context that stores information per file to e.g. allow communication
+ * between transformers.
+ * Note that the ts.TransformationContext stores context about all files
```
mprobst: maybe "There is one ts.TransformationContext per emit, but because files are handled sequentially by all transformers, we can store blah here..."

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+ * in the emit.
+ */
+class FileContext {
+  syntheticNodeParents = new Map<ts.Node, ts.Node|undefined>();
```
mprobst: document why we need to use a lookaside buffer for the new parents?

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  importOrReexportDeclarations: Array<ts.Node&{moduleSpecifier: ts.StringLiteral}> = [];
```
mprobst: Is this a `ts.ExportDeclaration|ts.ImportDeclaration`?

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  lastCommentEnd = -1;
+  constructor(public file: ts.SourceFile) {}
+}
+
+/**
+ * Transform that needs to be executed right before TypeScript's transform.
+ *
+ * This prepares the node tree to workaround some bugs in the TypeScript emitter.
+ */
+function preTypeScriptTransform(context: ts.TransformationContext) {
```
mprobst: semantic name

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+        // 1) we don't want to change them at all
+        // 2) TS emit becomes errorneous in some cases if we add a synthetic parent.
+        node.parent = parent;
+      }
+      fileCtx.syntheticNodeParents.set(node, parent);
+
+      const originalNode = ts.getOriginalNode(node);
+      // Needed so that e.g. `module { ... }` prints the variable statement
+      // before the closure.
+      // tslint:disable-next-line:no-any
```
mprobst: can you add a comment why you need the any?

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+  };
+}
+
+/**
+ * Transform that needs to be executed after TypeScript's transform.
+ *
+ * This fixes places where the TypeScript transformer does not
+ * emit synthetic comments.
+ */
+function postTypescriptTransform(context: ts.TransformationContext) {
```
mprobst: semantic name

src/transformer_util.ts
```diff
@@ -0,0 +1,545 @@
...
+    }
+    const textRange = {pos: node.pos, end: node.end};
+    ts.setSourceMapRange(node, textRange);
+    ts.setTextRange(node, {pos: -1, end: -1});
+    return node;
+  }
+}
+
+function synthesizeLeadingComments(
+    sourceFile: ts.SourceFile, node: ts.Node, lastCommentEnd: number): number {
```
mprobst: this seems to always `return -1` - possible a bug? Missing a test?

src/transformer.ts
```diff
@@ -0,0 +1,222 @@
...
+  shouldIgnoreWarningsForPath(filePath: string): boolean;
+}
+
+export function mergeEmitResults(emitResults: EmitResult[]): EmitResult {
+  const diagnostics: ts.Diagnostic[] = [];
+  let emitSkipped = true;
+  const emittedFiles: string[] = [];
+  const externs: {[fileName: string]: string} = {};
+  const modulesManifest = new ModulesManifest();
+  emitResults.forEach(er => {
```
mprobst: for/of

src/transformer_sourcemap.ts
```diff
@@ -0,0 +1,144 @@
...
+        return originalNode;
+      }
+      node = visitEachChild(node, visitNode, context);
+
+      node.flags |= ts.NodeFlags.Synthesized;
+      node.parent = undefined;
+      ts.setTextRange(node, originalNode ? originalNode : {pos: -1, end: -1});
+      ts.setOriginalNode(node, originalNode);
+
+      // tslint:disable-next-line:no-any
```
mprobst: why any

src/tsickle.ts
```diff
@@ -1088,13 +1158,47 @@ class Annotator extends ClosureRewriter {
...
+    const names: ts.Identifier[] = [];
+    if (importClause.name) {
+      names.push(importClause.name);
+    }
+    if (importClause.namedBindings &&
+        importClause.namedBindings.kind === ts.SyntaxKind.NamedImports) {
+      const namedImports = importClause.namedBindings as ts.NamedImports;
+      names.push(...namedImports.elements.map(e => e.name));
+    }
+    names.forEach(name => {
```
tbosch: Done

src/tsickle.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer.ts
tbosch: Done

src/transformer_sourcemap.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

src/transformer_util.ts
tbosch: Done

b-strauss:
FYI: this introduced 71 test failures on Windows. 😞

mprobst:
What kind? At 71 failures, it's most likely something trivial to fix?

b-strauss:
Pretty much every comment is missing. Maybe I can find a simple solution.

Example:
```
./test_files/type/type.js with ./test_files/type/type.js.transform.patch
      + expected - actual

      -goog.module('test_files.type.type');var module = module || {id: 'test_files/type/type.js'};let typeAny;
      -let typeArr;
      -let typeArr2;
      -let typeNestedArr;
      -let typeObject = { a: 3, b: 'b' };
      -let typeObjectIndexable;
      -let typeObjectMixedIndexProperty;
      -let typeObjectEmpty;
      -let typeTuple = [1, 2];
      -let typeComplexTuple = ['', true];
      -let typeTupleTuple = [[1, 2]];
      -let typeTupleTuple2 = [[1, 2], ''];
      -let typeUnion = Math.random() > 0.5 ? false : '';
      -let typeUnion2 = Math.random() > 0.5 ? false : '';
      -let typeOptionalField = {};
      -let typeOptionalUnion = {};
      -let typeFunc = function () { };
      -let typeFunc2 = function (a, b) { return ''; };
      -let typeFunc3 = function (x, cb) { return ''; };
      -// TODO: let typeFunc4: (a: number, ...args: number[]) => void;
      +goog.module('test_files.type.type');var module = module || {id: 'test_files/type/type.js'};/**
      + * @suppress {checkTypes} checked by tsc
      + */
      +let /** @type {?} */ typeAny;
      +let /** @type {!Array<?>} */ typeArr;
      +let /** @type {!Array<?>} */ typeArr2;
      +let /** @type {!Array<!Array<{a: ?}>>} */ typeNestedArr;
      +let /** @type {{a: number, b: string}} */ typeObject = { a: 3, b: 'b' };
      +let /** @type {!Object<string,number>} */ typeObjectIndexable;
      +let /** @type {?} */ typeObjectMixedIndexProperty;
      +let /** @type {!Object} */ typeObjectEmpty;
      +let /** @type {!Array<?>} */ typeTuple = [1, 2];
      +let /** @type {!Array<?>} */ typeComplexTuple = ['', true];
      +let /** @type {!Array<?>} */ typeTupleTuple = [[1, 2]];
      +let /** @type {!Array<?>} */ typeTupleTuple2 = [[1, 2], ''];
      +let /** @type {(string|boolean)} */ typeUnion = Math.random() > 0.5 ? false : '';
      +let /** @type {(string|boolean)} */ typeUnion2 = Math.random() > 0.5 ? false : '';
      +let /** @type {{optional: (undefined|boolean)}} */ typeOptionalField = {};
      +let /** @type {{optional: (undefined|string|boolean)}} */ typeOptionalUnion = {};
      +let /** @type {function(): void} */ typeFunc = function () { };
      +let /** @type {function(number, ?): string} */ typeFunc2 = function (a, b) { return ''; };
      +let /** @type {function(number, function(number): string): string} */ typeFunc3 = function (x, cb) { return ''; };
      +/**
      + * @param {function(number): number} callback
      + * @return {void}
      + */
       function typeCallback(callback) { }
       typeCallback(val => val + 1);
      +/**
      + * @template T
      + * @param {function(T): T} callback
      + * @return {void}
      + */
       function typeGenericCallback(callback) { }
       typeGenericCallback(val => val);
```

PR #520, opened 2017-06-21
Handle export * conflicting with renamed exports.

### LINKED ISSUE #519: tsickle fails to override re-exports when the symbol is imported
This exports `bar` as `foo` and doesn't export `foo` from `a.ts`
```TypeScript
export * from './a'; // a.ts exports only foo
const bar = 1;
export {bar as foo};
```

This exports `bar` as `foo` and also exports `foo` from `a.ts`
```TypeScript
import {foo} from './a';
export * from './a'; // a.ts exports only foo
const bar = 1;
export {bar as foo};
```

mprobst:
The previous code did not correctly ignore a symbol from export * if it
was conflicting with a renamed export, i.e. "export {x as y}".

Also simplifies the entire operation a bit, as the exports of a module
are readily available on its symbol.

Fixes #519.


test_files/index_import/user.tsickle.ts
```diff
@@ -8,7 +8,7 @@ import {a} from './has_index/index';
 const tsickle_forward_declare_1 = goog.forwardDeclare("test_files.index_import.has_index.index");
 export {a} from './has_index/index';
 const tsickle_forward_declare_2 = goog.forwardDeclare("test_files.index_import.has_index.index");
-export {a} from './has_index/index';
+export {} from './has_index/index';
```
mprobst: @evmar I think this change is harmless and/or intended, we just avoid assigning the same export twice. I think this probably was an oversight of the previous code, which was confused about the `export {a}` also having an `import {a}` here.

src/tsickle.ts
```diff
@@ -923,33 +923,20 @@ class Annotator extends ClosureRewriter {
...
-    for (const local of locals) {
-      if (local.declarations &&
-          local.declarations.some(d => d.kind === ts.SyntaxKind.ImportSpecifier)) {
-        continue;
-      }
-      localSet.add(local.name);
-    }
-
+    const moduleSymbol = this.typeChecker.getSymbolAtLocation(this.file);
+    const moduleExports = moduleSymbol && moduleSymbol.exports || new Map<string, ts.Symbol>();
```
evmar: Is this a ternary?  Write it with ?: if so?

mprobst: It is not - it selects the new map if `exports` is undefined, in addition to `moduleSymbol` being falsy.

PR #522, opened 2017-06-22
Support namespaces (aka modules) in ambient exports.

mprobst:
This allows users to write code like:

    export declare namespace MyNamespace { ... }


mprobst:
Turns out this is an incomplete fix – we can now emit the declaration, but all the types in it are incorrectly scoped. I fixed that in another change, but then it turns out users re-export declared namespaces, and I don't know how to express that in Closure Compiler. I've reached out to the mailing list. This will need some more thinking.

evmar:
As soon as imports/exports get involved it quickly becomes very unclear what tsickle ought to do.  :~(

mprobst:
So I reasoned about this a bit, it's indeed tricky. Essentially we're mostly mapping the TypeScript namespace to the Closure namespace, and they mostly match (at least now that we have the aliasing functionality). Local symbols in modules are local, global symbols are global.

However with the ambients, there are two cases: ambients in a script (effectively global symbols) and ambients in a module (local symbols). Global symbols directly map to Closure's global symbols.

Local ambient symbols are tricky because we have to do two things: we need to move them to an externs file, where Closure does not support modules, and we have to import them into the module from the externs file, and re-export them from the module. That's hard, because it involves a bunch of mapping and mangling, and because Closure does not support certain patterns, like re-exporting a type from externs without referencing an actual symbol in code.

Thinking about it that way gives some clarity. Let's see if we can find a way to express what we need in Closure. There might be alternatives to work around the problem, like special casing exported ambient namespaces.

mprobst:
OK, PTAL.

The basic approach is still there, but the code now handles a bunch of corner cases, such as referring to a non-ambient type from an ambient context, `declare global` augmentations, and other fun things. Also adds a bunch of tests.

test_files/namespaced/export_namespace.js.transform.patch
```diff
@@ -0,0 +1,11 @@
+Index: ./test_files/namespaced/export_namespace.js
+===================================================================
+--- ./test_files/namespaced/export_namespace.js	golden
++++ ./test_files/namespaced/export_namespace.js	tsickle with transformer
+@@ -17,5 +17,5 @@
+     function Interface() { }
+     typeNamespace.Interface = Interface;
+     function Interface_tsickle_Closure_declarations() {
+     }
+-})(typeNamespace = exports.typeNamespace || (exports.typeNamespace = {}));
```
mprobst: This appears to be a bug in the emit with transformers and namespaces. I'll file an issue.

src/type-translator.ts
```diff
@@ -137,6 +137,9 @@ export class TypeTranslator {
    */
   private readonly seenTypes: ts.Type[] = [];
 
+  /** Whether the write types suitable for an \@externs file. */
```
evmar: typo: whether "to"

evmar: Is it worth elaborating on why externs are special here?

src/type-translator.ts
```diff
@@ -145,6 +148,7 @@ export class TypeTranslator {
    *     as {?}.
    * @param symbolsToPrefix a mapping from symbols (`Foo`) to a prefix they should be emitted with
    *     (`tsickle_import.Foo`).
+   * @param isAmbient whether emitting happens in an ambient context.
```
evmar: Can you clarify this?  I think what you're saying is whether the symbols are in a module or not, but is the "ambient context" the input or the output?

src/type-translator.ts
```diff
@@ -174,17 +180,38 @@ export class TypeTranslator {
...
-        // for ambient declarations in externs.
+    if (useFqn && this.isForExterns) {
+      // For regular type emit, we can use TypeScript's naming rules, as they match Closure's name
+      // scoping rules. However when emitting externs files for ambients, naming rules change. As
+      // Closure doesn't support externs modules, all names must be global and use global fully
+      // qualified names. The code below uses TypeScript to convert a symbol to a full qualified
+      // name and then emits that.
+      let fqn = this.typeChecker.getFullyQualifiedName(sym);
+      if (fqn.startsWith(`"`) || fqn.startsWith(`'`)) {
+        // Quoted FQNs mean the name is from a module, e.g. `'path/to/module'.some.qualified.Name`.
```
evmar: I feel like sometimes I see a quoted name in these and sometimes another thing, but I can't quite remember the circumstances.

src/type-translator.ts
```diff
@@ -174,17 +180,38 @@ export class TypeTranslator {
...
+        // local. However value declarations that are local to modules but not defined do not make
+        // sense if not global, e.g. "declare class X {}; new X();" cannot work unless `X` is
+        // actually a global.
+        // So this code strips the module path from the type and uses the FQN as a global.
+        fqn = fqn.replace(/^["'][^"']+['"]\./, '');
+      }
+      // Declarations in module can re-open global types using "declare global { ... }". The fqn
+      // then contains the prefix "global." here. As we're mapping to global types, just strip the
+      // prefix.
+      const isInGlobal = (sym.declarations || []).some((d: ts.Node | undefined) => {
```
evmar: Why the |undefined here?  If it's only needed for the loop, it'd be clearer to make a new local var than to rely on the function type coercing.

```
(node: ts.Node) => {
  let d: ts.Node|undefined = node;
...
````

otherwise if my guess is wrong I don't get this code.

File: src/type-translator.ts [hunk 6]
```diff
@@ -238,6 +265,41 @@ export class TypeTranslator {
     // NonPrimitive occurs on its own on the lower case "object" type. Special case to "!Object".
     if (type.flags === ts.TypeFlags.NonPrimitive) return '!Object';
 
+    function isExternalAndNotNamespace(d: ts.Node|undefined) {
+      if (!ts.isExternalModule(d!.getSourceFile())) return false;
```
evmar: Should just remove |undefined from the type.

src/type-translator.ts
```diff
@@ -238,6 +265,41 @@ export class TypeTranslator {
...
+      while (d) {
+        if (d.kind === ts.SyntaxKind.ModuleDeclaration) return false;
+        if (ts.getCombinedModifierFlags(d) & ts.ModifierFlags.Ambient) return false;
+        d = d.parent;
+      }
+      return true;
+    }
+
+
+
```
evmar: Weird whitespace here

src/type-translator.ts
```diff
@@ -238,6 +265,41 @@ export class TypeTranslator {
...
+      }
+    }
+
+    // tsickle cannot generate types for non-ambient namespaces.
+    if (isNamespace && !isAmbient) return '?';
+
+    // Types in externs cannot reference types from external modules.
+    // However ambient types in modules get moved to externs, too, so type references work and we
+    // can emit a precise type.
+    if (this.isForExterns && isModule && !(isAmbient)) return '?';
```
evmar: Drop extra parens

test_files/declare/declare-module.d.ts
```diff
@@ -3,8 +3,10 @@
 // "declare global" should show up in the global namespace.
 declare global {
   var globalX: string;
-  namespace subnamespace {
+  export class GlobalClass {}
+  export namespace subnamespace {
```
evmar: This is just an ordinary namespace, not a sub namespace right?
Maybe just global in its name so it's more obvious in the js?

src/type-translator.ts
mprobst: Done, both.

src/type-translator.ts
mprobst: Uh, that param was removed in a later fixup but I forgot to drop the `@param`. "ambient context" == "any ancestor node has 'declare' on it"

src/type-translator.ts
mprobst: I looked at the TS code, unless I'm mistaken this always refers to the file the name is defined in.

src/type-translator.ts
mprobst: Done.

src/type-translator.ts
mprobst: Done (removed)

src/type-translator.ts
mprobst: Done.

src/type-translator.ts
mprobst: Done.

test_files/declare/declare-module.d.ts
mprobst: Done.

mprobst:
@evmar thanks, worked out :-)

PR #527, opened 2017-06-27
Collect all diagnostics since the last reconfigure.

### LINKED ISSUE #523: tsickle only reports errors from the last parsed file
Tsickle replaces the value of the `diagnostics` property on every call to `getSourceFile` (see [the line](https://github.com/angular/tsickle/blob/master/src/tsickle_compiler_host.ts#L323) `this.diagnostics = diagnostics` in `TsickleCompilerHost`). This is problematic as a call to `ts.createProgram` with a `TsickleCompilerHost` will call `getSourceFile` multiple times, and therefore only keep the diagnostics of the last parsed file.

tbosch:
The previous logic would override the diagnostics
on every call to `getSourceFile`.

Fixes #523


mprobst:
@googlebot it's OK. I merged changes from master to make it easier to test the branch, will rebase before merging.

PR #530, opened 2017-07-04
fix some of the failing transformer tests on Windows

b-strauss:
This also moves the line endings normalization into 'util'.


src/util.ts
```diff
@@ -88,3 +89,19 @@ export function createOutputRetainingCompilerHost(
...
+ * True if we are running on Windows (32 or 64 bit).
+ */
+export const isWindows: boolean = os.platform() === 'win32';
+
+/**
+ * Returns the input with line endings normalized to '\n', when running on Windows.
+ * Returns the string as is, for any other OS.
+ */
+export function normalizeLineEndings(input: string): string {
+  if (isWindows)
```
mprobst: please use curly braces in multi line conditional statements

mprobst: Is it important to only do so on Windows? Because if not, I'd just unconditionally replace `\r\n` with `\n`.

b-strauss: Not really, it's unnecessary on all other platforms, so I added it to save processing time. But I can remove it if you say it's not relevant.

mprobst: Just drop it, it won't save substantial time. It is adding some more complexity/branchiness to test code, which we generally want to avoid (bugs in test setup code are the worst...).

b-strauss:
@mprobst this can be reviewed again. By the way, do I have to leave a comment, or do you get notified automatically after I push changes?

mprobst:
@b-strauss thanks for the changes, LGTM.

I get notified when you push commits, but it's not clear if you intend the pushed commits to be the final state, so it's a good idea to reply "Please take another look" (PTAL) once you intend for the code to be reviewed again.

PR #531, opened 2017-07-07
add a test that captures our current augmentation behavior

evmar:
I want to change the emitted externs for module augmentation.
To start, this test shows what our current externs look like.


test_files/augment/angular-augment.d.ts
```diff
@@ -0,0 +1,23 @@
...
+  namespace sub {
+    type AugmentSubType = string;
+  }
+
+  /**
+   * sub is a new namespace introduced by this file.
+   */
+  namespace local {
+    type LocalType = string;
+  }
```
rkirov: I wonder if an simple `export let x: string;` might also be useful here. Augmentation is not just about namespaces.

Also is type only namespaces intentional, test will have greater coverage if it has some values in those too.

evmar: You can't have values in this namespace because the root is a value already.

PR #536, opened 2017-07-12
rationalize externs tests; improve modules

### LINKED ISSUE #423: module .d.ts emit into global externs
see d3_selection.d.ts, it

evmar:
- Improve our "declare"/externs test:
  - Rename some externs files and variables to better reflect
    their purpose.  For example, if we intend a variable to be
    emitted globally in the externs tag it with "global".
  - Add an additional test file there that is a plain .ts file
    that contains a "declare"
- Add more comments about how we intend for externs to work.
- Emit module-scoped externs into a module-scoped namespace.

Fixes #423.


src/tsickle.ts
```diff
@@ -1596,6 +1634,22 @@ class ExternsWriter extends ClosureRewriter {
     switch (node.kind) {
       case ts.SyntaxKind.SourceFile:
         const sourceFile = node as ts.SourceFile;
+        if (ts.isExternalModule(sourceFile)) {
+          // This file is a d.ts that is a module, so any of its declarations
+          // should be emitted into a namespace.
+          // The actual chosen namespace doesn't matter too much because we're
```
mprobst: What does this do to re-opening types/namespaces? Is that possible at all with external modules?

evmar:
I submitted the test changes from this in 9536d14d9c702b9e32faec5b61dda529bb0167ec, and I think we'll revisit module-scoped externs later.

PR #538, opened 2017-07-15
Wrap interface members in a function _tsickle_Closure_declarations().

mprobst:
This makes it easier for tools such as rollup or uglify to detect that these
declarations are unused, which will allow users to compile the Closure emit
with such tools without getting a code size hit.

### DIFF
--- a/test_files/decorator/external.js
+++ b/test_files/decorator/external.js
@@ -17,3 +17,5 @@ exports.AClassWithGenerics = AClassWithGenerics;
  */
 function AType() { }
 exports.AType = AType;
+function AType_tsickle_Closure_declarations() {
+}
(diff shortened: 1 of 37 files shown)


mprobst:
Still needs a g3 run to confirm working.

test_files/decorator/external.js
```diff
@@ -17,3 +17,5 @@ exports.AClassWithGenerics = AClassWithGenerics;
  */
 function AType() { }
 exports.AType = AType;
+function AType_tsickle_Closure_declarations() {
+}
```
evmar: Do we care about these empty functions?  Could be solved with an extra 'if' in the code.

mprobst: I saw that, but I figured that this shouldn't hurt assuming rollup always figures out correctly that this unused (right?), and I'd rather have the code less branchy. Also doesn't add too much code in development mode.

mprobst: Re frequency, I don't think it's generally common to define interfaces that contain no values, so this should be rare.

PR #539, opened 2017-07-15
Emit enums as proper Closure types.

### LINKED ISSUE #535: Improve enum emit for Closure compatibility
1) should emit some @enum tag or at least a typedef so it is a Closure-visible type

2) should generate valid code on 'export enum ...'

I think enums were fixed up recently in @tbosch's recent patch series, need to double check what he did.

mprobst:
This was previously avoided to keep the output valid TypeScript wise, but that
is not necessary, as we do not type check the intermediate output.

This has a remaining issue where it will emit Closure types for const enums that
disappear in runtime emit. We will need to investigate whether we should emit
the enum type for those.

Fixes #535.


mprobst:
Needs a g3 run to confirm working.

src/tsickle.ts
```diff
@@ -1541,31 +1535,34 @@ class Annotator extends ClosureRewriter {
...
+    this.emit(`/** @enum {number} */\n`);
+    this.emit(`const ${name} = {\n`);
+    // Emit enum values ('BAR: 0,').
     for (const member of toArray(members.keys())) {
-      if (!this.options.untyped) this.emit(`/** @type {number} */\n`);
-      this.emit(`${name}.${member} = `);
       const value = members.get(member)!;
+      this.emit(`${member}: `);
       if (typeof value === 'number') {
         this.emit(value.toString());
```
evmar: I wonder why this branch is necessary, it seems you could just always visit value

mprobst: The number is actually a literal number, not a ts Node.

mprobst:
PTAL. I haven't cleaned up the CL descriptions, but essentially I changed this to always split the enum declaration into enum and export (because otherwise local type references don't work), and changed it to always emit a type, even in untyped mode, as that's harmless and essentially what we do already.

This requires tsickle to not run with `.d.ts` emit.

PR #540, opened 2017-07-17
Ensure TypeScript can emit a .d.ts for the given code.

mprobst:
If the effective type shape of a program refers to private names,
TypeScript cannot generate a .d.ts for it, and fails during emit.
Previously tsickle did not test for this.

This requires updating fields.ts. Because it is not a module, it
effectively exports every global variable it defines, which means its
private class type is affected by this change. Simply changing it to a
module resolves the problem.

### DIFF
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -33,6 +33,7 @@ export const compilerOptions: ts.CompilerOptions = {
   outDir: '.',
   strictNullChecks: true,
   noImplicitUseStrict: true,
+  declaration: true,
 };
 
 const {cachedLibPath, cachedLib} = (() => {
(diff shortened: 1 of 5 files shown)


test/tsickle_test.ts
```diff
@@ -200,11 +200,24 @@ function runGoldenTests(useSeparatePasses: boolean) {
...
+        const jsPath = './' + diag.file.fileName.replace(/\.ts$/, '.js');
+        const msg = tsickle.formatDiagnostics([diag]);
+        if (!transformed[jsPath]) {
+          throw new Error(`unexpected diagnostic, ${jsPath} not in ${
+                                                                     Object.keys(transformed)
+                                                                   }, diagnostic is: ${msg}`);
+        }
+        const current = transformed[jsPath] || '';
+        // Add messages in line comments, so they don't break Closure compilation.
+        transformed[jsPath] = `// ${msg}\n${current}`;
```
evmar: Can msg be multiline?  I would expect formatDiagnostics to sometimes produce some context for example.  I think even using a `/*... */` comment might be unsafe; safest thing is to
```
split(\n).map(`// ${line}`).join(\n)
```
I think.

test/test_support.ts
```diff
@@ -77,20 +78,21 @@ export function createProgramAndHost(
 }
 
 /** Emits transpiled output with tsickle postprocessing.  Throws an exception on errors. */
```
evmar: This comment is no longer correct.

test/tsickle_test.ts
```diff
@@ -200,11 +200,24 @@ function runGoldenTests(useSeparatePasses: boolean) {
       compareAgainstGolden(allExterns, test.externsPath);
 
       // Run tsickled TypeScript through TypeScript compiler
-      // and compare against goldens.
+      // and compare emit or diagnostics against goldens.
```
evmar: Why include diagnostics here?  It seems it affects no tests?  If we want this, we should have a test that exhibits it.

mprobst:
Decided not to do this, but rather disable .d.ts emit.

PR #545, opened 2017-07-21
Polish enum emit a bit.

mprobst:
- simplify value emitting. Collecting the values up front is actually
  not needed.
- make sure to retain comments on enum values, they might carry semantic
  importance (e.g. `/** @export */`).

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -753,7 +753,7 @@ class Annotator extends ClosureRewriter {
         this.visitTypeAlias(node as ts.TypeAliasDeclaration);
         return true;
       case ts.SyntaxKind.EnumDeclaration:
-        this.maybeProcessEnum(node as ts.EnumDeclaration);
+        this.processEnum(node as ts.EnumDeclaration);
         return true;
       case ts.SyntaxKind.TemplateSpan:
         this.templateSpanStackCount++;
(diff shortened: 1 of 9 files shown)


src/tsickle.ts
```diff
@@ -1461,18 +1461,33 @@ class Annotator extends ClosureRewriter {
...
+      this.writeRange(member, member.getFullStart(), member.getStart());
+      this.emit(`${memberName}: `);
+
       if (member.initializer) {
         const enumConstValue = this.typeChecker.getConstantValue(member);
         if (enumConstValue !== undefined) {
-          members.set(memberName, enumConstValue);
-          i = enumConstValue + 1;
+          this.emit(enumConstValue.toString());
+          // members.set(memberName, enumConstValue);
```
evmar: xxx

src/tsickle.ts
```diff
@@ -1461,18 +1461,33 @@ class Annotator extends ClosureRewriter {
...
+    //   const Foo = {BAR: 0, BAZ: 1, ...};
+    //   export {Foo};  // even if originally exported on one line.
+    // This declares an enum type for Closure Compiler (and Closure JS users of this TS code).
+    // Splitting the enum into declaration and export is required so that local references to the
+    // type resolve ("@type {Foo}").
+    this.emit('\n');
+    const name = node.name.getText();
+
+    this.emit(`/** @enum {number} */\n`);
+    this.emit(`const ${name}: any = {\n`);
```
evmar: We don't type check the output, right?  So you can drop the ":any".  Elsewhere in tsickle I remember using some ": DontTypeCheckMe" annotation to make this explicit.

src/tsickle.ts
```diff
@@ -1461,18 +1461,33 @@ class Annotator extends ClosureRewriter {
...
+      this.emit(`${memberName}: `);
+
       if (member.initializer) {
         const enumConstValue = this.typeChecker.getConstantValue(member);
         if (enumConstValue !== undefined) {
-          members.set(memberName, enumConstValue);
-          i = enumConstValue + 1;
+          this.emit(enumConstValue.toString());
+          // members.set(memberName, enumConstValue);
+          enumIndex = enumConstValue + 1;
```
evmar: Since you're in here, I'll note that in TS 2.4 the enum const value can be a string.  I think there's nothing for you to to here yet but just FYI in case that guides your design any.

src/tsickle.ts
```diff
@@ -1489,37 +1504,15 @@ class Annotator extends ClosureRewriter {
...
           // value in the above branch of this "if" statement.
-          members.set(memberName, member.initializer);
+          // members.set(memberName, member.initializer);
+          this.visit(member.initializer);
         }
       } else {
-        members.set(memberName, i);
-        i++;
+        this.emit(enumIndex.toString());
+        // members.set(memberName, enumIndex);
```
evmar: xxx

src/tsickle.ts
```diff
@@ -1489,37 +1504,15 @@ class Annotator extends ClosureRewriter {
...
+        enumIndex++;
```
evmar: What happens in

```
enum A {
  x = 1.3,
  y
}
```

?

(And do we even care)

src/tsickle.ts

src/tsickle.ts

src/tsickle.ts
mprobst: I've left a TODO. My understanding is that string enums require each case to have an assignment, so we'd just need to check whether `enumConstValue` is a string, and then not increment here.

src/tsickle.ts

src/tsickle.ts
mprobst: Hilariously, this all works. TS emits something like `A[A["x"] = 1.3] = "x"; A[A["y"] = 2.3] = "y";`, and we follow suit. We'll need an `enumMatches` that takes into account floating point imprecision...

mprobst:
Green in g3.

PR #546, opened 2017-07-21
Handle importing modules with no symbol (no exports)

mprobst:


src/rewriter.ts
```diff
@@ -51,8 +51,15 @@ export abstract class Rewriter {
...
     // this.logWithIndent('node: ' + ts.SyntaxKind[node.kind]);
     this.indent++;
-    if (!this.maybeProcess(node)) {
-      this.writeNode(node);
+    try {
+      if (!this.maybeProcess(node)) {
+        this.writeNode(node);
+      }
+    } catch (e) {
+      e.message += `\n at ${ts.SyntaxKind[node.kind]} in ${this.file.fileName}:`;
```
evmar: Is it guaranteed that e will have a .message, even for stuff like divide by zero errors or whatever?  (I don't understand why exceptions in TS are type any)

src/tsickle.ts
```diff
@@ -1222,6 +1222,11 @@ class Annotator extends ClosureRewriter {
     const forwardDeclarePrefix = `tsickle_forward_declare_${++this.forwardDeclareCounter}`;
     const moduleNamespace =
         nsImport !== null ? nsImport : this.host.pathToModuleName(this.file.fileName, importPath);
+    const moduleSymbol = this.typeChecker.getSymbolAtLocation(specifier);
+    // Modules that do not have exports do not have a symbol. These can still be imported, either as
```
evmar: This is a surprise to me!  But I guess it sorta makes sense.

In your test it looks like it's importing a script, not a module.  Is it worth adding an "export {}" to make it a module?  I think the answer is no, because people are likely to forget that (or intentionally leave it out) on files imported for their side effects.

So I think my suggestion here is to just adjust this comment to also mention scripts.

src/rewriter.ts
mprobst: I've added an assignment into message in case it doesn't exist. There's a chance that somebody would just `throw 'foo';`, which is probably also why they are `any` in TS.

src/tsickle.ts
mprobst: Adjusted the comment. Indeed this only happens for scripts - if the imported is a module, it'll have a symbol. What happens is that importing an empty set of symbols from a script causes its global declarations to be around, but does not emit a load statement, which is what people appear to use it for.

I'm not even sure if that's a bug in TS or intended behaviour, but we should definitely not crash on it, and it currently works in TS.

PR #550, opened 2017-07-22
Add support for NonPrimitive type "object".

mprobst:


mprobst:
Green in g3.

src/type-translator.ts
```diff
@@ -234,6 +234,10 @@ export class TypeTranslator {
     // type.flags. This mask limits the flag checks to the ones in the public API. "lastFlag" here
     // is the last flag handled in this switch statement, and should be kept in sync with
     // typescript.d.ts.
+
+    // NonPrimitive occurs on its own on the lower case "object" type. Special case to "!Object".
+    if (type.flags === ts.TypeFlags.NonPrimitive) return '!Object';
```
evmar: Can this be handled in the below switch statement?  Or is it beyond that mask?  Should you adjust the mask?

mprobst: It cannot, it's beyond the mask, and there's intervening values between it and the highest value. Also, string, number, and boolean values share the flag, and we correctly handle them due to the mask.

PR #552, opened 2017-07-24
Dummy to Import to G3

LucasSloan:


mprobst:
Can you update the PR description? I don't get what the dummy is about.

src/tsickle.ts
```diff
@@ -634,6 +634,11 @@ class Annotator extends ClosureRewriter {
    *     emit it as is and visit its children.
    */
   maybeProcess(node: ts.Node): boolean {
+    if (this.file.statements.indexOf(node as ts.Statement) !== -1) {
```
mprobst: Can you explain *why* we're doing this in the commit message? See go/cl-descriptions.

In particular, what does this fix? How does it relate to the idea of emitting mappings for each emitted string?

Conceptually, shouldn't we rather make sure we emit a mapping for each line of code that can actually be executed, assuming we have proper coverage for all tokens we emit? This seems like it'll still have issues for e.g. multiple statements in one line, or code nested deeper than the top level.

mprobst: this is O(n) on each statement, so O(n^2) in the number of statements in the source file.

src/tsickle.ts
```diff
@@ -634,6 +634,11 @@ class Annotator extends ClosureRewriter {
...
+      const pos = this.file.getLineAndCharacterOfPosition(node.getFullStart() + 1);
+      this.sourceMapper.addMapping(
+            node, {line: pos.line, column: pos.character, position: node.getFullStart() + 1}, {line: this.position.line + 1, column: 0, position: this.position.position + 1}, 1);
```
mprobst: this repeats a bunch of logic regarding indices that seems to belong into the parent class where we handle the source mapping stuff. You could introduce a method there, then you also wouldn't need to open up the visibility.

src/source_map_utils.ts
```diff
@@ -135,8 +135,8 @@ export class DefaultSourceMapper implements SourceMapper {
 
   constructor(private fileName: string) {
     this.sourceMap.addMapping({
-      original: {line: 1, column: 1},
-      generated: {line: 1, column: 1},
+      original: {line: 1, column: 0},
```
mprobst: I don't understand this change, but maybe I'm just mislead by the commit message.

In TypeScript (as in: `tsc` and the `typescript` node module & APIs), line and column are zero based. I cannot find an authoritative source on source maps (it'd be nice if you could link to one here), but at least `sourcemap.ts` in the TS code base seems to assume 1-based indexing, whereas the Closure Compiler assumes 0-based for the most recent format. https://github.com/google/closure-compiler/blob/master/src/com/google/debugging/sourcemap/FilePosition.java#L38

Maybe that's the problem you're fixing here?

In any case, please add a comment explaining what's going on, with a pointer to exactly what bits are breaking, what you are fixing, and a test case.

LucasSloan:
You shouldn't be reviewing this - I just opened a pr so Copybara had a PR to sync into G3 to test if this approach made any sense.

PR #553, opened 2017-07-25
Fix annotation emit for nested classes

### LINKED ISSUE #547: Tsickle incorrectly downlevels constructor parameters when class is declared inside a closure
See: https://github.com/angular/tsickle/pull/548

```
var describe: Function;
var it: Function;
/** @Annotation */
var Component: Function;

export function main() {
  describe(() => {
    class SomeService {}
    it(() => {
      @Component({template: '<div someDir>{{1 | somePipe}}</div>'})
      class TestComp3 {
        constructor(service: SomeService) {}
      }
    });

  });
}
```

produces syntactically incorrect code:
```
static ctorParameters = () => [ {type:, }, ];
```

mprobst:
Fixes #547.


src/rewriter.ts
```diff
@@ -93,18 +93,26 @@ export abstract class Rewriter {
...
       this.writeRange(node, pos, child.getFullStart());
       this.visit(child);
       pos = child.getEnd();
     });
     this.writeRange(node, pos, end);
-    this.skipUpToOffset = oldSkipUpToOffset;
+    this.skipCommentsUpToOffset = oldSkipCommentsUpToOffset;
+  }
+
+  writeComment(node: ts.Node) {
```
evmar: Note this is also important for whitespace.

src/decorator-annotator.ts
```diff
@@ -15,14 +15,23 @@ import {SourceMapper} from './source_map_utils';
 import {assertTypeChecked, TypeTranslator} from './type-translator';
 import {toArray} from './util';
 
+/**
+ * ConstructorParameters are gathered from constructors, so that their type information and
+ * decorators can later be emitted as an annotation.
+ */
+interface ConstructorParameter {
+  type: ts.TypeNode|null;
```
evmar: This could probably use a bit of docs, because it's pretty weird that it's a TypeNode (reference to ast) and also that we only set it if the AST refers to a value.

mprobst:
Green in g3.

src/rewriter.ts
mprobst: Renamed to writeTrivia.

src/decorator-annotator.ts
mprobst: Added docs. This is indeed still a somewhat odd code structure, but at least the fields have names now, and can be documented :-)

PR #554, opened 2017-07-25
Clean up tsickle testing (a bit)

mprobst:
This PR removes the patching system from tests in favour of separate golden files for the transformer run variant and reduces code duplication between the transformer and regular golden tests.

It also reinstates running golden tests in `watch` mode, which was removed in b26b2cbfd3873a52f8eedd0ee5a8be866adee83f for unclear reasons.


test_files/coerce/coerce.js.transformer
```diff
@@ -0,0 +1,12 @@
+goog.module('test_files.coerce.coerce');var module = module || {id: 'test_files/coerce/coerce.js'};/**
```
mprobst: @evmar LMK what you think of this - while the `.patch` files lack context to review them, these here lack clarity in what changed. Not sure what I prefer, but it'll be easy to reinstate the diffing method for these.

test_files/jsdoc/diagnostics.txt.transformer
```diff
@@ -0,0 +1,20 @@
+Warning at test_files/jsdoc/jsdoc.ts:44:3: unhandled anonymous type
```
mprobst: @evmar I don't know what's going on in here, but AFAICT this was the previous behaviour, the previous test runner just swallowed these. It looks a bit as if we re-check the comments after manipulating them to include types ourselves.

evmar: Can you file a bug about it?

test_files/coerce/coerce.js.transformer
evmar: Now that I see the bare files, I see why the patch files were nice.  I hope we are only temporarily in the state where we have transformer/non-transformer output so either seems acceptable to me.

PR #555, opened 2017-07-26
Generate source maps with 0-indexed columns.

LucasSloan:
Typescript produces source maps with 1-indexed lines, and 0-indexed columns.  We previously produced source maps with 1-indexed columns. This produced errors most noticable at the beginnings of lines - the composed source maps would have a one character wide mapping into the tsickle generated intermediate file.  The rest of the line would have correct source mappings, so all our source map tests which look for tokens that aren't at the beginnings of lines didn't catch this.

### DIFF
--- a/src/source_map_utils.ts
+++ b/src/source_map_utils.ts
@@ -135,8 +135,9 @@ export class DefaultSourceMapper implements SourceMapper {
 
   constructor(private fileName: string) {
     this.sourceMap.addMapping({
-      original: {line: 1, column: 1},
-      generated: {line: 1, column: 1},
+      // tsc's source maps use 1-indexed lines, 0-indexed columns
+      original: {line: 1, column: 0},
+      generated: {line: 1, column: 0},
       source: this.fileName,
     });
   }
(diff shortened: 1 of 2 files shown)


src/source_map_utils.ts
```diff
@@ -135,8 +135,9 @@ export class DefaultSourceMapper implements SourceMapper {
 
   constructor(private fileName: string) {
     this.sourceMap.addMapping({
-      original: {line: 1, column: 1},
-      generated: {line: 1, column: 1},
+      // tsc's source maps 1-index lines, 0-index columns
```
mprobst: nit: "source maps _use_ 1-indexed lines..." or so? The grammar seems a bit off.

test/e2e_source_map_test.ts
```diff
@@ -310,6 +310,29 @@ function createTests(useTransformer: boolean) {
           .to.equal('input2.ts', 'input file name');
     }
   });
+
+  it('maps at the start of lines correctly', () => {
+    const sources = new Map<string, string>();
```
mprobst: FYI you can do: `new Map([['input.ts', 'code goes here']])` (even with type inference!)

src/source_map_utils.ts
mprobst: Best reference I can find is here:
https://groups.google.com/forum/#!topic/mozilla.dev.js-sourcemap/kdOU6cGsplA

@concavelenz any update or insight on that? It seems like we indeed need to use 1-indexed lines, 0-indexed columns?

LucasSloan: Done.

test/e2e_source_map_test.ts
LucasSloan: Good to know!

PR #557, opened 2017-07-26
Fix compilation when generating .d.ts.

mprobst:
Without the explicit conversion to boolean, this actually returns the unexported type ConstructorParameter, which makes compilation fail when generating .d.ts files.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -196,8 +196,8 @@ export class DecoratorClassVisitor {
     return false;
   }
 
-  foundDecorators() {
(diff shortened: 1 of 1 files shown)


src/decorator-annotator.ts
```diff
@@ -197,7 +197,7 @@ export class DecoratorClassVisitor {
   }
 
   foundDecorators() {
```
evmar: Maybe add a return type here?  From the function name it's not clear if it's returning a list of found decorators or if it's returning whether it found any.  (And it'd also have revealed this bug.)

mprobst: Done.

PR #559, opened 2017-07-31
Emit just `?` for templatized unhandled types.

mprobst:
tsickle needs to emit `?` for types in several situations. If such a
type is templatized, e.g. `T<U>`, tsickle would previously translate `T`
to `?`, and then emit `?<U>`, which is illegal syntax in Closure
Compiler.

This change avoids the problem by just returning `?` directly in such
situations.

This also adds a test for blacklisted parameterized types which I
originally suspected to be the trigger. It turns out the problem was
caused by type/value conflicts, but the test coverage still seems
useful, so I'm leaving these in.

### DIFF
--- a/test_files/jsdoc_types/jsdoc_types.ts
+++ b/test_files/jsdoc_types/jsdoc_types.ts
@@ -6,7 +6,7 @@
 import * as module1 from './module1';
 import {ClassOne, value, ClassOne as RenamedClassOne, ClassTwo as RenamedClassTwo, Interface, ClassWithParams} from './module2';
 import DefaultClass from './default';
-import {NeverTyped} from './nevertyped';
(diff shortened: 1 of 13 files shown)


test_files/jsdoc_types/jsdoc_types.ts
```diff
@@ -6,7 +6,7 @@
 import * as module1 from './module1';
 import {ClassOne, value, ClassOne as RenamedClassOne, ClassTwo as RenamedClassTwo, Interface, ClassWithParams} from './module2';
 import DefaultClass from './default';
-import {NeverTyped} from './nevertyped';
+import {NeverTyped, NeverTypedTemplated} from './nevertyped';
```
mprobst: This adds some test coverage, but is a red herring - the problem turned out to be the type-vs-value dichotomy in the test below.

PR #561, opened 2017-07-31
Zoom! Enhance!

mprobst:
Allow @enhance and @enhanceable in TypeScript JSDoc tags.

Used by some internal tooling.


mprobst:
it used to be that you could ignore whitespace in diffs by appending ?w=1 to the URL...

PR #562, opened 2017-07-31
use a glob in tsconfig for file matching

evmar:
A recent rename meant that tsconfig wasn't covering some test files.
Use a glob to prevent this from happening again.


evmar:
Yeah :(

But this list keeps getting out of sync so meh.

PR #563, opened 2017-08-01
Smaller fixes for transformer mode

tbosch:


src/transformer.ts
```diff
@@ -205,9 +205,10 @@ function combineInlineSourceMaps(
 function combineSourceMaps(
     program: ts.Program, filePath: string, tscSourceMapText: string): string {
   const tscSourceMap = parseSourceMap(tscSourceMapText);
+  const fileDir = path.dirname(filePath);
   let tscSourceMapGenerator: SourceMapGenerator|undefined;
   for (const sourceFileName of tscSourceMap.sources) {
-    const sourceFile = program.getSourceFile(sourceFileName);
+    const sourceFile = program.getSourceFile(path.resolve(fileDir, sourceFileName));
```
evmar: @LucasSloan  FYI

src/tsickle.ts
```diff
@@ -1870,6 +1890,20 @@ class ExternsWriter extends ClosureRewriter {
...
 }
 
+function isPolymerBehaviorPropertyInCallExpression(pa: ts.PropertyAssignment): boolean {
+  if (pa.name.kind !== ts.SyntaxKind.Identifier ||
+      (pa.name as ts.Identifier).text !== 'behaviors') {
+    return false;
+  }
+
+  const parentParent = pa.parent && pa.parent.parent;
+  if (!pa.parent || !pa.parent.parent || pa.parent.parent.kind !== ts.SyntaxKind.CallExpression) {
```
evmar: Move this above the var decl above it?

src/tsickle.ts
```diff
@@ -785,6 +785,17 @@ class Annotator extends ClosureRewriter {
       case ts.SyntaxKind.AsExpression:
         // Both of these cases are AssertionExpressions.
         const typeAssertion = node as ts.AssertionExpression;
+        if (this.polymerBehaviorStackCount > 0) {
+          // Don't emit type casts for Polymer behaviors that are declared
+          // by calling the Polymer function
+          // as the Polymer closure plugin does not work when emitting them.
```
evmar: I don't understand what goes wrong here.  What are we doing that's illegal?

src/decorator-annotator.ts
```diff
@@ -371,6 +378,49 @@ class DecoratorRewriter extends Rewriter {
   }
 }
 
+/**
+ * Collect the Identifiers used as named bindings in the given import declaration
+ * with their Symbol.
+ * This is needed lateron to find an identifier that represents the value
```
evmar: Typo: "later on"

src/decorator-annotator.ts
```diff
@@ -148,7 +148,8 @@ export class DecoratorClassVisitor {
    * This method looks for the place where the value that is associated to
```
evmar: I don't get this patch's intent.  Why not just require the user to refer to the right import?

tbosch: Oh, the user did. But the old code did not. I.e. the old code just looked for the first import that had the right path, which might have been different than the one the user specified.

src/tsickle.ts
tbosch: We are doing the right thing. However, the polymer closure plugin does not work when we (correctly) emit type comments for casts for values in the `behaviors` property of an object literal that is passed to `Polymer`. 

The non transformer version never emitted the type cast (or rather typescript removed it due to the bug in Typescript with comments).

This needs to be fixed in Polymer, I am just adding this so we don't depend on them for landing our transformer support.

src/transformer.ts
tbosch: The original tsickle_compiler_host already does this, I just reimplemented it incorrectly.

src/decorator-annotator.ts
evmar: Ah, I see, thanks for explaining.  This is well-documented in the file, I just didn't see it in the context, so please disregard this comment.

src/decorator-annotator.ts
```diff
@@ -159,9 +160,10 @@ export class DecoratorClassVisitor {
     if (valueName.getSourceFile() === this.rewriter.file) {
       return valueName;
     }
+    const typeStr = typeNode.getText();
     for (let i = 0; i < this.importedNames.length; i++) {
       const {name, declarationNames} = this.importedNames[i];
-      if (declarationNames.some(d => d === valueName)) {
+      if (typeStr.startsWith(name.text) && declarationNames.some(d => d === valueName)) {
```
evmar: Why is this startsWith() and not ===?  Something about generics?  It feels pretty fragile.

src/tsickle.ts
evmar: OK, I think the comment makes sense, no need to change anything.

src/tsickle.ts
tbosch: Done

src/decorator-annotator.ts
tbosch: Done

src/decorator-annotator.ts
tbosch: yes, this is about generics. changed to compare the first identifier in the type instead (using `===`)

src/decorator-annotator.ts
```diff
@@ -159,9 +160,10 @@ export class DecoratorClassVisitor {
     if (valueName.getSourceFile() === this.rewriter.file) {
       return valueName;
     }
+    const typeStr = typeNode.getText();
     for (let i = 0; i < this.importedNames.length; i++) {
       const {name, declarationNames} = this.importedNames[i];
```
evmar: (By the way, I think this could be

```
for (const {name, declarationNames} of this.importedNames) {
```

Not sure though.)

tbosch: Yes, that works.

src/decorator-annotator.ts
```diff
@@ -159,10 +160,15 @@ export class DecoratorClassVisitor {
     if (valueName.getSourceFile() === this.rewriter.file) {
       return valueName;
     }
-    for (let i = 0; i < this.importedNames.length; i++) {
-      const {name, declarationNames} = this.importedNames[i];
-      if (declarationNames.some(d => d === valueName)) {
-        return name;
+    // Need to look at the first identifier only
```
tbosch: I changed this based on your review.

src/tsickle.ts
```diff
@@ -1870,6 +1890,18 @@ class ExternsWriter extends ClosureRewriter {
   }
 }
 
+function isPolymerBehaviorPropertyInCallExpression(pa: ts.PropertyAssignment): boolean {
+  const parentParent = pa.parent && pa.parent.parent;
+  if (pa.name.kind !== ts.SyntaxKind.Identifier ||
```
tbosch: I included the check for `parentParent` in this `if` clause (based on your comment).

src/decorator-annotator.ts
```diff
@@ -371,6 +382,56 @@ class DecoratorRewriter extends Rewriter {
   }
 }
 
+function firstIdentifier(node: ts.Node): ts.Identifier|null {
```
evmar: Can you add some docs on this?

PR #564, opened 2017-08-01
update to typescript 2.4

evmar:
- ts.Declaration was split to not have a name field, and there's
  now ts.NamedDeclaration for those with names.  Adjust all users.
- Drop the unused 'filterTypesForExport' option.
- Handle some more maybe-null values (e.g. 'file' in ts.Diagnostics).
- Change enum emit to handle string enums.  Do one pass over the enum
  to figure out whether it's a number/string/other enum.
- Adjust type translation of enums values because their representation
  changed.  This has the consequence of "inlining" all const enums
  (users of them now look like plain 'number' or 'string') but that
  is the semantics of a const enum so I think it's harmless.

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -123,7 +123,7 @@ export class DecoratorClassVisitor {
   /**
    * gatherMethod grabs the decorators off a class method and emits nothing.
    */
-  private gatherMethodOrProperty(method: ts.Declaration) {
+  private gatherMethodOrProperty(method: ts.NamedDeclaration) {
     if (!method.decorators) return;
     if (!method.name || method.name.kind !== ts.SyntaxKind.Identifier) {
       // Method has a weird name, e.g.
@@ -152,10 +152,9 @@ export class DecoratorClassVisitor {
    */
   private getValueIdentifierForType(typeSymbol: ts.Symbol, typeNode: ts.TypeNode): ts.Identifier
       |null {
-    if (!typeSymbol.valueDeclaration) {
-      return null;
-    }
-    const valueName = typeSymbol.valueDeclaration.name;
+    const valueDeclaration = typeSymbol.valueDeclaration as ts.NamedDeclaration;
+    if (!valueDeclaration) return null;
+    const valueName = valueDeclaration.name;
     if (!valueName || valueName.kind !== ts.SyntaxKind.Identifier) {
       return null;
     }
@@ -274,7 +273,7 @@ export class DecoratorClassVisitor {
           // rewriting. Note that we cannot use param.type as the emit node directly (not even just
           // for mapping), because that is marked as a type use of the node, not a value use, so it
           // doesn't get updated as an export.
-          const sym = this.typeChecker.getTypeAtLocation(param.type).getSymbol();
+          const sym = this.typeChecker.getTypeAtLocation(param.type).getSymbol()!;
           const emitNode = this.getValueIdentifierForType(sym, param.type);
           if (emitNode) {
             this.rewriter.writeRange(emitNode, emitNode.getStart(), emitNode.getEnd());
@@ -421,15 +420,16 @@ export function collectImportedNames(typeChecker: ts.TypeChecker, decl: ts.Impor
     names.push(...namedImports.elements.map(e => e.name));
   }
   for (const name of names) {
-    let symbol = typeChecker.getSymbolAtLocation(name);
+    let symbol = typeChecker.getSymbolAtLocation(name)!;
     if (symbol.flags & ts.SymbolFlags.Alias) {
       symbol = typeChecker.getAliasedSymbol(symbol);
     }
     const declarationNames: ts.Identifier[] = [];
     if (symbol.declarations) {
       for (const d of symbol.declarations) {
-        if (d.name && d.name.kind === ts.SyntaxKind.Identifier) {
-          declarationNames.push(d.name as ts.Identifier);
+        const decl = d as ts.NamedDeclaration;
+        if (decl.name && decl.name.kind === ts.SyntaxKind.Identifier) {
+          declarationNames.push(decl.name as ts.Identifier);
         }
       }
     }
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -583,7 +583,7 @@ class Annotator extends ClosureRewriter {
       case ts.SyntaxKind.InterfaceDeclaration:
       case ts.SyntaxKind.ClassDeclaration:
       case ts.SyntaxKind.ModuleDeclaration:
-        const decl = node as ts.Declaration;
+        const decl = node as ts.NamedDeclaration;
         if (!decl.name || decl.name.kind !== ts.SyntaxKind.Identifier) {
           break;
         }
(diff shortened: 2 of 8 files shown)


evmar:
Hold off on reviewing this, I moved part of it to https://github.com/angular/tsickle/pull/565

src/decorator-annotator.ts
```diff
@@ -266,7 +267,7 @@ export class DecoratorClassVisitor {
           // rewriting. Note that we cannot use param.type as the emit node directly (not even just
           // for mapping), because that is marked as a type use of the node, not a value use, so it
           // doesn't get updated as an export.
-          const sym = this.typeChecker.getTypeAtLocation(param.type).getSymbol();
+          const sym = this.typeChecker.getTypeAtLocation(param.type).getSymbol()!;
```
mprobst: I guess this is cool because `param.type` will always have a symbol on it?

src/tsickle.ts
```diff
@@ -867,7 +867,8 @@ class Annotator extends ClosureRewriter {
         // If it has a symbol, it's actually a regular declared property.
         if (!quotedPropSym) return false;
         const declarationHasQuotes =
-            !quotedPropSym.declarations || quotedPropSym.declarations.some(decl => {
+            !quotedPropSym.declarations || quotedPropSym.declarations.some(d => {
+              const decl = d as ts.NamedDeclaration;
               if (!decl.name) return false;
```
mprobst: the !decl.name looks a bit as if this was handling `NamedDeclaration`s and `Declaration`s. Do we have a better way to runtime test if this is named or unnamed? `if (typeof decl.name !== undefined)` or so?

src/tsickle.ts
```diff
@@ -1200,15 +1201,16 @@ class Annotator extends ClosureRewriter {
...
       if (symbol.flags & ts.SymbolFlags.Alias) {
         symbol = this.typeChecker.getAliasedSymbol(symbol);
       }
       const declarationNames: ts.Identifier[] = [];
       if (symbol.declarations) {
         for (const d of symbol.declarations) {
-          if (d.name && d.name.kind === ts.SyntaxKind.Identifier) {
-            declarationNames.push(d.name as ts.Identifier);
+          const decl = d as ts.NamedDeclaration;
+          if (decl.name && decl.name.kind === ts.SyntaxKind.Identifier) {
```
mprobst: same comment wrt `decl.name` testing

src/tsickle.ts
```diff
@@ -1500,7 +1502,27 @@ class Annotator extends ClosureRewriter {
     this.emit('\n');
     const name = node.name.getText();
 
-    this.emit(`/** @enum {number} */\n`);
+    // Figure out whether this is a string or number enum, by iterating through the enum
```
mprobst: and so the code comes back ;-)

src/tsickle.ts
```diff
@@ -1500,7 +1502,27 @@ class Annotator extends ClosureRewriter {
...
+    // members.  Note that TS allows mixed string/number enums also; just translate as {?}.
+    let enumType: string|undefined;
```
mprobst: could we make this a union?

src/tsickle.ts
```diff
@@ -1500,7 +1502,27 @@ class Annotator extends ClosureRewriter {
     this.emit('\n');
     const name = node.name.getText();
 
-    this.emit(`/** @enum {number} */\n`);
+    // Figure out whether this is a string or number enum, by iterating through the enum
+    // members.  Note that TS allows mixed string/number enums also; just translate as {?}.
```
mprobst: what happens if you have a mixed string/non-string enum value?

```ts
enum X {
  A = 1,
  B = 'x',
  C,
}
```

What's `C`? What is it in tsickle?

src/type-translator.ts
```diff
@@ -360,6 +352,34 @@ export class TypeTranslator {
           return this.translateUnion(type as ts.UnionType);
         }
 
+        // Enum literals are represented as
+        //   ts.TypeFlags.NumberLiteral | ts.TypeFlags.EnumLiteral
+        //   ts.TypeFlags.StringLiteral | ts.TypeFlags.EnumLiteral
+        if (type.flags & ts.TypeFlags.EnumLiteral) {
+          const enumLiteralBaseType = this.typeChecker.getBaseTypeOfLiteralType(type);
```
mprobst: extract a method for this? It seems a bit too long for this case.

src/type-translator.ts
```diff
@@ -360,6 +352,34 @@ export class TypeTranslator {
...
+            this.warn(`EnumLiteralType without a symbol`);
+            return '?';
+          }
+
+          if (enumLiteralBaseType === type) {
+            // For const enums, the type and base type are the same.
+            if (type.flags & ts.TypeFlags.NumberLiteral) {
+              return 'number';
+            } else if (type.flags & ts.TypeFlags.StringLiteral) {
+              return 'string';
```
mprobst: doesn't this means we emit `@type {string}` instead of `@type {EnumMemberType}`? Those are incompatible in Closure. Why is this the `const` case?

test_files/enum/enum.ts
```diff
@@ -54,3 +54,8 @@ enum EnumWithNonConstValues {
   Scheme = (x => x + 1)(3),
   UserInfoRenamed = ComponentIndex.UserInfo,
 }
+
+enum StringEnum {
```
mprobst: Add a test for a mixed enum?

test_files/enum_value_literal_type/enum_value_literal_type.tsickle.ts
```diff
@@ -10,4 +10,4 @@ const ExportedEnum: DontTypeCheckMe = {
 export {ExportedEnum};
 ExportedEnum[ExportedEnum.VALUE] = "VALUE";
 
-export const /** @type {ExportedEnum} */ x: ExportedEnum.VALUE = ExportedEnum.VALUE;
+export const /** @type {number} */ x: ExportedEnum.VALUE = ExportedEnum.VALUE;
```
mprobst: Is ExportedEnum a const enum?

src/decorator-annotator.ts
evmar: Correct.  We'll only have saved this param as a thing to investigate if it had a symbol.

src/tsickle.ts
evmar: The definition is:
```
    interface NamedDeclaration extends Declaration {
        name?: DeclarationName;
    }
```
so it's always safe to coerce downward and test the attribute.

It looks like upstream they're trying to iteratively migrate code to using NamedDeclaration where needed: https://github.com/Microsoft/TypeScript/commit/9392342f7807c28738199ce82c73e52ac3bc73ef

src/tsickle.ts
evmar: On that code:
```
ERROR(4,3): : Enum member must have initializer.
```
so tsickle will never see it.

But this is legal:
```
enum X {
  B = 'x',
  A = 1,
  C,
}
```
and in that, C is 2.

src/tsickle.ts
evmar: Fixed.

test_files/enum/enum.ts
evmar: Done.

evmar:
Rebased to master, PTAL.

src/type-translator.ts
```diff
@@ -374,6 +370,24 @@ export class TypeTranslator {
     return parts.length === 1 ? parts[0] : `(${parts.join('|')})`;
   }
 
+  private translateEnumLiteral(type: ts.Type): string {
+    // Suppose you had:
```
mprobst: Maybe this should be JSDoc?

PR #565, opened 2017-08-01
improve whitespace of enum emit

evmar:
We already emit the leading trivia of enum members, so we don't
need the extra newline after the opening curly.  Removing this
newline improves the look of the output and also reduces the
delta from the transformer output.


mprobst:
Much nicer!

PR #566, opened 2017-08-02
Strip incoming sourcemaps from the sources in the source map

tbosch:
to reduce the size of the final source map.


tbosch:
TBD: also strip the extra mapping for the line that contains the incoming sourcemap comment.

tbosch:
Yes, we still use them, but just don't want to have the input sourcemaps in the sourceContent of the output sourcemaps anymore.

tbosch:
There is no mapping for the incoming sourcemap comment in the produced sourcemap, so this PR is fine as is.

PR #568, opened 2017-08-02
Don’t get confused when import statements are elided when emitting co…

tbosch:
…mments

in transformer mode.

### DIFF
--- a/src/transformer_util.ts
+++ b/src/transformer_util.ts
@@ -72,8 +72,7 @@ class FileContext {
    * have a new parent now.
    */
   syntheticNodeParents = new Map<ts.Node, ts.Node|undefined>();
-  importOrReexportDeclarations:
-      Array<(ts.ExportDeclaration | ts.ImportDeclaration)&{moduleSpecifier: ts.StringLiteral}> = [];
+  importOrReexportDeclarations: Array<ts.ExportDeclaration|ts.ImportDeclaration> = [];
   lastCommentEnd = -1;
   constructor(public file: ts.SourceFile) {}
 }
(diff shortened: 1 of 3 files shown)


src/transformer_util.ts
```diff
@@ -124,28 +124,19 @@ function prepareNodesBeforeTypeScriptTransform(context: ts.TransformationContext
...
         // this should always be a StringLiteral, see the typescript docs.
-        if (id.moduleSpecifier.kind !== ts.SyntaxKind.StringLiteral) {
-          throw new Error(`Unexpected moduleSpecifier kind: ${id.moduleSpecifier.kind}`);
-        }
-        importOrReexportNode = node as ts.ImportDeclaration & {moduleSpecifier: ts.StringLiteral};
-      } else if (node.kind === ts.SyntaxKind.ExportDeclaration) {
-        const ed = node as ts.ExportDeclaration;
-        if (ed.moduleSpecifier) {
+        if (ied.moduleSpecifier) {
           // this should always be a StringLiteral, see the typescript docs.
```
evmar: This comment is duplicated from two lines up, I think maybe delete the upper one?

src/transformer_util.ts
```diff
@@ -202,8 +202,10 @@ function emitMissingSyntheticCommentsAfterTypescriptTransform(context: ts.Transf
       // TypeScript ignores synthetic comments on reexport / import statements.
       const moduleName = extractModuleNameFromRequireVariableStatement(node);
       if (moduleName && fileContext.importOrReexportDeclarations) {
-        const index = importOrReexportsCount++;
-        const originalImportOrReexportNode = fileContext.importOrReexportDeclarations[index];
+        const index = (importOrReexportsCount[moduleName] || 0);
+        importOrReexportsCount[moduleName] = index + 1;
+        const originalImportOrReexportNode =
+            (fileContext.importOrReexportDeclarations[moduleName] || [])[index];
```
evmar: Won't this always index out of bounds (in the case where it's `[]`)?  I guess it's terser than writing the ternary?

tbosch: I changed the implementation to merge the comments for all import/export statements of the same moduleName and added details why.

tbosch:
The current PR merges the comments in this case (see the comment in `transformer_util` why I think this is ok for now). However, Travis is red as I don't distinguish between `import` and `export`.

test/test_support.ts
```diff
@@ -71,6 +71,14 @@ export function createProgramAndHost(
     }
     throw new Error('unexpected file read of ' + fileName + ' not in ' + toArray(sources.keys()));
   };
+  const originalFileExists = host.fileExists;
+  host.fileExists = (fileName: string): boolean => {
+    if (path.isAbsolute(fileName)) fileName = path.relative(process.cwd(), fileName);
+    if (sources.has(fileName)) {
+      return true;
+    }
```
evmar: What is this for?

tbosch: So that the `import` / `export` statements in the transformer_util_test.ts actually work. Otherwise I can't test that exports are elided because TypeScript cannot see the files.

We don't need this for the regular golden tests as TS can find them via the default implementation of the compiler host.

PR #573, opened 2017-08-04
Emit @export tag for properties as part of _tsickle_Closure_declarations.

### LINKED ISSUE #558: @ExportDecoratedItems stopped working for properties
`@ExportDecoratedItems` used to work on properties, but not anymore. In fact, the existing unit test is no longer testing that properties work. I believe this behavior changed in #450. At least, that PR changed the test to no longer verify that the `exportMe` property is indeed exported.

@drjanitor, @evmar, was that intended? If so, perhaps we should change the unit test to not create the impression that you can use it on properties like `exportMe`.

drjanitor:
Fixes #558

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1436,6 +1436,9 @@ class Annotator extends ClosureRewriter {
 
     const tags = this.getJSDoc(prop) || [];
     tags.push({tagName: 'type', type});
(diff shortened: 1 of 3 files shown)


evmar:
Looks like this needs to be rebased.

drjanitor:
guess so. can't do anything till tomorrow.

drjanitor:
Looks all green at the moment.

PR #575, opened 2017-08-04
Provide a source mapping for every emit.

LucasSloan:
When tsickle rewrites a source file, it produces text 2 ways.  First it copies text from an AST node to the output.  Second, it produces its own text and outputs it.  Previously, we only provided source mappings for the first case, since we had the node to map back to.  This PR changes rewriter.ts to keep track of the current node on a stack as it traverses the AST and uses that node to provide source mappings for the second kind of emit.  In certain cases the classes that extend rewriter.ts don't use its node traversal mechanisms to visit the children of certain node types it transforms - in those cases the node stack is manually updated.

This PR also adds a bunch of source map tests, roughly one for each kind of AST node that we transform.  In writing those tests 2 issues arrose:

1.  Typescript doesn't produce good source maps for multiline/jsdoc comments, so we can't currently provide source maps for function signature annotations.
2.  The transformer system doesn't like having custom emits mapped back to original nodes - I got a bunch of errors initially and I was forced to preserve the existing source mapping behavior for the transformer system.  I'll have to work with Tobias on this.

### DIFF
--- a/src/rewriter.ts
+++ b/src/rewriter.ts
@@ -32,7 +32,15 @@ export abstract class Rewriter {
    */
   private skipCommentsUpToOffset = -1;
 
+  /**
+   * Stack of the nodes we've visited.  Protected so that extenders who perform
+   * logic outside the visit()/maybeProcess() pattern can keep track of which node
+   * they're processing.
+   */
+  protected nodeStack: ts.Node[];
+
   constructor(public file: ts.SourceFile, private sourceMapper: SourceMapper = NOOP_SOURCE_MAPPER) {
+    this.nodeStack = [file];
   }
 
   getOutput(): {output: string, diagnostics: ts.Diagnostic[]} {
@@ -50,6 +58,7 @@ export abstract class Rewriter {
    */
   visit(node: ts.Node) {
     // this.logWithIndent('node: ' + ts.SyntaxKind[node.kind]);
+    this.nodeStack.push(node);
     this.indent++;
     try {
       if (!this.maybeProcess(node)) {
@@ -63,6 +72,7 @@ export abstract class Rewriter {
       throw e;
     }
     this.indent--;
+    this.nodeStack.pop();
   }
 
   /**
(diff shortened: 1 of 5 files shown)


src/rewriter.ts
```diff
@@ -126,26 +136,44 @@ export abstract class Rewriter {
...
+   *
+   * Every emit call results in a source mapping being made, if a node and
+   * original SourcePosition aren't supplied, str is source mapped to the
+   * beginning of the current node.
+   */
+  emit(str: string): void;
+  emit(str: string, node: ts.Node, original: SourcePosition): void;
+  emit(str: string, node?: ts.Node, original?: SourcePosition): void {
+    if (!node || !original) {
+      // Only the transformer source mapper uses the node, and providing
```
tbosch: The transformer source mapper is the only source mapper that is used in G3, as the transformer version is now active.

tbosch:
Meta note: I think this PR becomes simpler after we removed the old `TsickleCompilerHost`, as we can then simplify the `SourceMapper` interface to not use `SourcePosition`s but rather just take number as position (as that is all that we need with transformers).

@evmar Also signaled that we should get rid of the old code path as soon as possible.

src/rewriter.ts
tbosch: Maybe change the signature to take an `originalPos?: number` instead of a `SourcePosition`. This way, all of the calls to `findOffsetOfFirstNonWhitespaceCharacter` are in the `emit` method, both for the original and the generated position.

src/rewriter.ts
```diff
@@ -126,26 +136,44 @@ export abstract class Rewriter {
...
+      // a source mapping for every emit messes up the transformer
+      // code path, so unless we're emitting copied text from a node, don't
+      // set it
+      original = this.sourcePositionOfNodeStart(this.getCurrentNode());
```
tbosch: To make sourcemaps work well with transformers, we should still pass the `this.getCurrentNode()` to `addMapping`, however with an additional flag that controls whether the transformer does a `ts.setOriginalNode` or just a `ts.setSourceMapRange`...
I think this difference is the problem that you ran into, as right now we are always doing both.

src/rewriter.ts
```diff
@@ -157,6 +185,36 @@ export abstract class Rewriter {
...
+  }
+
+  /**
+   * If you pass a pos, it returns a Source Position offset from that pos.
+   *  Otherwise, it just returns the absolute offsets
+   */
+  findOffsetOfFirstNonWhitespaceCharacter(str: string, pos = {line: 0, column: 0, position: 0}):
+      SourcePosition {
+    let {line, column, position} = pos;
+    for (const c of str) {
```
tbosch: Once we removed the old codepath / `SourceMapper` (as I mentioned in another comment), we don't need to care about row / column anymore, so this code can be simplified to just adjust the position, maybe via single regex.

PR #578, opened 2017-08-10
Remove import rewriting.

alexeagle:
In February, Closure Compiler added --moduleResolution=node so we no longer need this workaround.

BREAKING CHANGE: option convertIndexImportShorthand was removed.


alexeagle:
The option is only referenced by Angular. I think we need to remove it there first so that Angular can compile in google3.

mprobst:
@alexeagle I don't understand how this would work with `goog.module`. If you have a `foo/bar/index.ts` file, it'll become `goog.module('foo.bar.index')` AFAIU. That means importers will need to reference it in the same way, right? That is, the tests that we have here must still pass, removing them is not correct as long as we use `goog.module` anywhere.

alexeagle:
Update: I have not forgotten about this, haven't had time yet.

PR #579, opened 2017-08-10
Add source map tests from the tag-every-emit branch

LucasSloan:
Some of them don't actually pass on the pseudo-transformer codepath, so I've commented them out with TODOs to enable them when the relevant transforms have a proper transformer for them.

### DIFF
--- a/test/e2e_node_kind_source_map_test.ts
+++ b/test/e2e_node_kind_source_map_test.ts
@@ -1,0 +1,147 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// tslint:disable:no-unused-expression mocha .to.be.empty getters.
+
+import {expect} from 'chai';
+import * as path from 'path';
+import * as ts from 'typescript';
+
+import * as tsickle from '../src/tsickle';
+
+import {compile, getLineAndColumn} from './test_support';
+
+describe.only('source maps each node with transformer', () => {
+  it('maps import declarations correctly', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `import * as ts from 'typescript';
+      import {join, format as fmt} from 'path';
+      ts.ModuleKind.CommonJS;
+      join;
+      fmt;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+
+    {
+      const {line, column} = getLineAndColumn(compiledJS, `var ts`);
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(1, 'typescript require');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, `goog.require('typescript')`);
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(1, 'typescript require');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, `var path_1`);
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(2, 'path require');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, `goog.require('path')`);
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(2, 'path require');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps export declarations correctly', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `export const x = 4;
+      const y = 'stringy';
+      export {y};
+      export {format as fmt} from "path";`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'exports.x = 4;');
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(1, 'x export');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'exports.y = y;');
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(3, 'y export');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'exports.fmt');
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(4, 'fmt export');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'path_1.format');
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(4, 'fmt export');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps element access', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `class X {
+        [propName: string]: any;
+      }
+
+      const x = new X();
+      x.foo;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+
+    {
+      const {line, column} = getLineAndColumn(compiledJS, '["foo"];');
+      expect(sourceMap.originalPositionFor({line, column}).line)
+          .to.equal(6, 'rewritten element access');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps decorators', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `
+        /** @Annotation */
+        function classAnnotation(t: any) {
+            return t;
+        }
+
+        @classAnnotation({
+            x: 'thingy',
+        })
+        class DecoratorTest1 {
+            y: string;
+        }`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'classAnnotation, args');
+      expect(sourceMap.originalPositionFor({line, column}).line)
+          .to.equal(7, 'individual decorator type');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, `x: 'thingy'`);
+      expect(sourceMap.originalPositionFor({line, column}).line)
+          .to.equal(8, 'individual decorator args');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+});
--- a/test/e2e_source_map_test.ts
+++ b/test/e2e_source_map_test.ts
@@ -8,19 +8,17 @@
 
 // tslint:disable:no-unused-expression mocha .to.be.empty getters.
 
-import {assert, expect} from 'chai';
+import {expect} from 'chai';
 import * as path from 'path';
-import {RawSourceMap, SourceMapConsumer} from 'source-map';
+import {RawSourceMap} from 'source-map';
 import * as ts from 'typescript';
 
-import * as cliSupport from '../src/cli_support';
 import {convertDecorators} from '../src/decorator-annotator';
-import {toClosureJS} from '../src/main';
-import {containsInlineSourceMap, DefaultSourceMapper, getInlineSourceMapCount, parseSourceMap, setInlineSourceMap, sourceMapTextToConsumer} from '../src/source_map_utils';
+import {containsInlineSourceMap, DefaultSourceMapper, getInlineSourceMapCount, parseSourceMap, setInlineSourceMap} from '../src/source_map_utils';
 import * as tsickle from '../src/tsickle';
-import {createOutputRetainingCompilerHost, createSourceReplacingCompilerHost, toArray} from '../src/util';
+import {toArray} from '../src/util';
 
-import * as testSupport from './test_support';
+import {compile, createProgram, getLineAndColumn} from './test_support';
 
 describe('source maps with TsickleCompilerHost', () => {
   createTests(false);
@@ -72,7 +70,7 @@ function createTests(useTransformer: boolean) {
         let c : string = a + b;`);
 
     // Run tsickle+TSC to convert inputs to Closure JS files.
-    const {compiledJS, sourceMap} = compile(sources, {useTransformer});
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer, outFile: 'output.js'});
 
     {
       const {line, column} = getLineAndColumn(compiledJS, 'a string');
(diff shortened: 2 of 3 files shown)


test/e2e_source_map_test.ts
```diff
@@ -22,9 +22,6 @@ import {createOutputRetainingCompilerHost, createSourceReplacingCompilerHost, to
 
 import * as testSupport from './test_support';
 
-describe('source maps with TsickleCompilerHost', () => {
```
tbosch: Maybe keep this for now, and remove the tests for `TsickleCompilerHost` when the class itself is removed?

LucasSloan: Unfortunately, the tests I added don't pass with the baseline TsickleCompilerHost code path.  This seemed like the least confusing low effort option.  Do you think it's worth eg splitting them into their own file?  Can you think of an easy way to make them only run for the transformer code path without adding another set of flags?

tbosch: Ah, ok. Yes, maybe it is worth splitting them into an own file as there are not many (before your change).

test/e2e_source_map_test.ts
```diff
@@ -460,7 +1041,16 @@ function compile(
     return {compiledJS: '', dts: '', sourceMap: sourceMapTextToConsumer(''), sourceMapText: ''};
   }
 
-  const compiledJS = getFileWithName(options.outFile, jsFiles);
+  let outFile = options.outFile;
+  if (!outFile) {
```
mprobst: prefer just fixing up all callsites or a default value over optional args.

test/e2e_source_map_test.ts
```diff
@@ -22,9 +22,6 @@ import {createOutputRetainingCompilerHost, createSourceReplacingCompilerHost, to
 
 import * as testSupport from './test_support';
```
mprobst: While you're touching this - these tests run very slowly because they parse `lib.d.ts` over and over again. Our other tests reuse the same script file, which speeds them up a lot (from 1s per test to 10ms or so). Could you take a look at the test harness in the other files, and maybe harmonize it and improve on this? Not necessarily in this PR.

test/e2e_source_map_test.ts
```diff
@@ -367,6 +368,587 @@ function createTests(useTransformer: boolean) {
...
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'null');
+      expect(sourceMap.originalPositionFor({line, column}).line)
+          .to.equal(3, 'type cast with parentheses');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps type assertions', () => {
```
mprobst: How many of these tests actually test something that follows a different code path in the underlying implementation? Tests are good, but if they don't really test anything different, we'll just slow down our test suite and will have to maintain these in the future.

test/e2e_source_map_test.ts
LucasSloan: Unfortunately, tsc complains if you have imports or exports and specify an outFile.  That makes the sensible default undefined, but we still need the name of the outfile so we can get if from the compilation output, thus this code.

test/e2e_source_map_test.ts
LucasSloan: Almost all of these test new code paths.  The existing tests test the same single pass source map code paths over and over, checking out different ways of combining source maps from different passes.  These test that each of the node types we transform is mapped correctly.  There's some duplication because we have to test *some* type of source mapping in the existing tests, but not a huge amount.

test/test_support.ts
```diff
@@ -143,3 +147,178 @@ export function goldenTests(): GoldenFileTest[] {
...
+      outFile = inputFileName.substring(0, inputFileName.length - 3) + '.js';
+    } else {
+      outFile = 'input.js';
+    }
+  }
+  const compiledJS = getFileWithName(outFile, jsFiles);
+
+  if (!compiledJS) {
+    assert.fail();
+    return {compiledJS: '', dts: '', sourceMap: sourceMapTextToConsumer(''), sourceMapText: ''};
```
mprobst: `assert.fail()` doesn't return, so this line is misleading. If the type system requires you to prove compiledJS is non-null, then `throw new Error('unreachable')`, otherwise just remove the line.

test/test_support.ts
```diff
@@ -143,3 +147,178 @@ export function goldenTests(): GoldenFileTest[] {
...
+      const inputFileName = toArray(sources.keys())[0];
+      outFile = inputFileName.substring(0, inputFileName.length - 3) + '.js';
+    } else {
+      outFile = 'input.js';
+    }
+  }
+  const compiledJS = getFileWithName(outFile, jsFiles);
+
+  if (!compiledJS) {
+    assert.fail();
```
mprobst: should this have a better error message?

test/test_support.ts
```diff
@@ -143,3 +147,178 @@ export function goldenTests(): GoldenFileTest[] {
...
+      compileWithTransfromer(resolvedSources, fileNames, tsickleHost, compilerOptions) :
+      compileWithTsickleCompilerHost(resolvedSources, fileNames, tsickleHost, compilerOptions);
+
+  if (diagnostics.length) {
+    console.error(tsickle.formatDiagnostics(diagnostics));
+    assert.fail();
+    return {compiledJS: '', dts: '', sourceMap: sourceMapTextToConsumer(''), sourceMapText: ''};
+  }
+
+  let outFile = options.outFile;
```
mprobst: (New comment thread because you overwrote the commits of the old thread, so I cannot continue there)

If your code uses ES6 modules, `tsc` complains because it cannot (easily) combine modules into a single output file, so that's reasonable on that side.

Regarding `if (!outFile) {`, things to fix:
* [ ] please change the code to use a different variable name that better indicates what's going on. You're currently overloading the meaning of `outFile` to be both the file produced by tsc and the file you want to extract the output for, so this variable here should have a different name. 
* [ ] the implicit behaviour to combine inputs is non-obvious, this should have implementation documentation here. You're using `outFile` as a convenient way to get only one `compiledJS` output. That's a bit misleading, as in normal operation tsickle will generate individual files. Tests would arguably be more understandable if you returned all generated outputs and asserted on that. I understand the difference in convenience (though it might be worth considering if the alternative is so much more verbose), but this at least needs docs.
* [ ] you seem to have a bunch of tests that set outFile despite only passing a single input - that doesn't make sense then, does it? IIUC `outFile` should only be set when you pass multiple inputs and want to test composition of source maps in that situation.

test/test_support.ts
```diff
@@ -143,3 +147,178 @@ export function goldenTests(): GoldenFileTest[] {
...
+  if (!outFile) {
+    if (sources.size === 1) {
+      const inputFileName = toArray(sources.keys())[0];
```
mprobst: consider `sources.keys().next().value`.

LucasSloan:
I'm not sure I understand your objection to adding these tests.  Does this sound like a fair rephrase of your position?

Given that we're relying on the same code path to call `addSourceMapping` for all of these constructs, we don't need to check that it works for functions and classes, since it's the same code in both cases.

I think our disagreement comes down to your statement about the full transformers future:

> Given that this code will be obsolete once we move to full transformers (as TS will handle the source maps then, and we don't want to double-test their implementation), I don't see how these tests would be useful in the future either, so I don't think adding them for some future value holds - does it?

As far as I can tell, source maps don't just work when you're using full transformers - if you want a source map that provides the level of detail we want you have to call `addSourceMapping` on basically every node you produce.  I want these source map tests to ensure we get that part right in the future.  Does that make sense?

test/test_support.ts
LucasSloan: Done.

test/test_support.ts
LucasSloan: That sounds like a good idea - done.

mprobst:
> Given that we're relying on the same code path to call addSourceMapping for all of these constructs, we don't need to check that it works for functions and classes, since it's the same code in both cases.

Not quite: what I'm saying is: if in the code path we call `addSourceMapping`, it's fine to have a test asserting the result of that. AFAICT you have a bunch of tests for code paths not calling that, those are questionable.

> As far as I can tell, source maps don't just work when you're using full transformers - if you want a source map that provides the level of detail [...]

This might deserve investigation into what detail of source maps you get out of the box, and what detail we desire to have. E.g. for the fake `tsickle_declare_type` functions, I'm not sure if we even need source maps. For the `/** jsdoc */` sections we're adding, it might be fine to just have the entire section point at the function, constructor, etc. that they are caused by, without further resolution.

I'd expect transformers to give us that rough level of mapping, which would be enough by my book. Do you think there are use cases that need more fine grained detail? Do you think you understand them sufficiently to make the right calls in implementing them now?

test/test_support.ts
LucasSloan: Yep, that was a hack to get the type system not to complain - thanks for the suggestion!

test/test_support.ts
LucasSloan: 1.  Done - renamed to `compiledJSFileName`
2.  Unfortunately, I'm also testing that if we apply 2 different code paths in the source map combining code for a single file, that everything still works, so we can't just make assertions for each generated file.  I added a comment to the CompilerOptions interface.
3.  There was one test that was setting outFile unnecessarily when there was only one input file, in the other case, we're specifying that the compiled js is in another folder.

LucasSloan:
>Not quite: what I'm saying is: if in the code path we call addSourceMapping, it's fine to have a test asserting the result of that. AFAICT you have a bunch of tests for code paths not calling that, those are questionable.

I think it pays to take a more expansive view of what needs to be tested.  I'd like to know that our source maps are good, and that we're not going to have a user report  in 3 weeks finding some corner case we didn't think of.  One counterfactual way to do that would be to compile a large code base and sanity check all the resulting source maps, but I haven't been able to come up with a satisfactory sanity check.  In the absence of that, the next best thing I can think of is to make some rigorous assertions about each type of transform that we apply.  Some of those assertions turn out to be trivially true, but I didn't know that I wasn't missing something until I wrote the tests.

>This might deserve investigation into what detail of source maps you get out of the box, and what detail we desire to have. E.g. for the fake tsickle_declare_type functions, I'm not sure if we even need source maps. For the /** jsdoc */ sections we're adding, it might be fine to just have the entire section point at the function, constructor, etc. that they are caused by, without further resolution.  

We do need to map the tsickle_declare_type functions - you opened a bug when that wasn't working.

>I'd expect transformers to give us that rough level of mapping, which would be enough by my book. Do you think there are use cases that need more fine grained detail? Do you think you understand them sufficiently to make the right calls in implementing them now?

Unfortunately, they don't necessarily automatically give you that level of mapping.  If you're replacing a node tree with a different node tree (eg a class), you don't get reduced resolution mappings, you just get nothing.

mprobst:
> I think it pays to take a more expansive view of what needs to be tested. I'd like to know that our source maps are good, and that we're not going to have a user report in 3 weeks finding some corner case we didn't think of. One counterfactual way to do that would be to compile a large code base and sanity check all the resulting source maps, but I haven't been able to come up with a satisfactory sanity check. In the absence of that, the next best thing I can think of is to make some rigorous assertions about each type of transform that we apply. Some of those assertions turn out to be trivially true, but I didn't know that I wasn't missing something until I wrote the tests.

By your testing method, we'll essentially end up with a test for every syntactical construct there is. That's over-testing and sets us up for a lot of maintenance cost, in particular given that we already have a test suite for these constructs to maintain. What we should test is that our source mapping method works, and that the specific constructs that we emit using strings do get mapped correctly. Please make the appropriate changes to this PR.

> We do need to map the tsickle_declare_type functions - you opened a bug when that wasn't working.

I'm not arguing that we should not map that method, I'm calling it out as an example where it's not 100% clear what detail we really need. That is, we need the mapping not to have holes, but do we need to map every token in them, or is it sufficient to map them to the class that declares the fields? It seems like that'll be fine in most situations, given that the only situation in which users should see these in error messages is arguably a tool failure on our side (unlike actually executed code that could show up in stack traces).

> Unfortunately, they don't necessarily automatically give you that level of mapping. If you're replacing a node tree with a different node tree (eg a class), you don't get reduced resolution mappings, you just get nothing.

If we don't set the appropriate mapping in that situation, we end up with incorrect output that should be captured by the regular test case, don't we?

LucasSloan:
>By your testing method, we'll essentially end up with a test for every syntactical construct there is. That's over-testing and sets us up for a lot of maintenance cost, in particular given that we already have a test suite for these constructs to maintain. What we should test is that our source mapping method works, and that the specific constructs that we emit using strings do get mapped correctly. Please make the appropriate changes to this PR.

Ok, I've gone through and removed test cases for anything that we're not custome emitting.  Most notable changes for classes, where I've removed assertions about the class body, retaining only the assertions about type annotations, and template spans where I rewrote the assertions to focus on type casts, since that's the main thing that depends on our depth in template span trees.

>I'm not arguing that we should not map that method, I'm calling it out as an example where it's not 100% clear what detail we really need. That is, we need the mapping not to have holes, but do we need to map every token in them, or is it sufficient to map them to the class that declares the fields? It seems like that'll be fine in most situations, given that the only situation in which users should see these in error messages is arguably a tool failure on our side (unlike actually executed code that could show up in stack traces).

I think it's worth mapping to the individual field - it works out to be the same amount of work.

>If we don't set the appropriate mapping in that situation, we end up with incorrect output that should be captured by the regular test case, don't we?
Nope, the emitted code can be good even if the source map is bad.

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+    }
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'y.z');
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(4, 'object access line');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  // TODO(lucassloan): enable when we have a custom transformer for classes
```
mprobst: We typically do not check commented out code in - experience is that it has bad chances to stay correct even in the short term. Can you move these bits to a separate branch in your fork, and just send them out once we're ready?

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+      expect(sourceMap.originalPositionFor({line, column}).line).to.equal(1, 'string literal');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps constructors', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `class X {
+      constructor(private x: number, public y: string) {}
```
mprobst: ping?

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+  //     expect(sourceMap.originalPositionFor({line, column}).line)
+  //         .to.equal(2, 'enum option name declaration');
+  //     expect(sourceMap.originalPositionFor({line, column}).source)
+  //         .to.equal('input.ts', 'input file name');
+  //   }
+  // });
+
+  it('maps template spans', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', '`text text text\n${1 + 1}\n${null as string}`');
```
mprobst: ping

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+      const {line, column} = getLineAndColumn(compiledJS, 'null');
+      expect(sourceMap.originalPositionFor({line, column}).line)
+          .to.equal(3, 'type cast with parentheses');
+      expect(sourceMap.originalPositionFor({line, column}).source)
+          .to.equal('input.ts', 'input file name');
+    }
+  });
+
+  it('maps type assertions', () => {
+    const sources = new Map<string, string>();
```
mprobst: I don't see anything source map specific about type assertions and source maps, can you point me to it?

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+      'input.ts', `let x : number = 2;
+      x + 1;
+      let y = {z: 2};
+      y.z;`
+    ]]);
+
+    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+
+    {
+      const {line, column} = getLineAndColumn(compiledJS, 'var /** @type {number} */ x');
```
mprobst: These test are overall a bit boilerplatey. Expressing "here's a line of source code, and the output should map 'blah' back to line X in the file" takes 13 lines of code, with an extra 6 per assertion.

It's always a balance to avoid obscuring the test too much, but maybe this would be a good place to write something like:

```ts
const TESTS = [
  {description: '...', source: '...', assertions: [{text: '...', line: ...}, ...],
]
```

LucasSloan: Maybe, not this PR, I think.

test/e2e_node_kind_source_map_test.ts
LucasSloan: Sure.

test/e2e_node_kind_source_map_test.ts
LucasSloan: Removed.

test/e2e_node_kind_source_map_test.ts
```diff
@@ -0,0 +1,625 @@
...
+    //       .to.equal(2, 'field declaration');
+    //   expect(sourceMap.originalPositionFor({line, column}).source)
+    //       .to.equal('input.ts', 'input file name');
+    // }
+  });
+
+  // TODO(lucassloan): enable when we have a custom transformer for functions
+  // it('maps function definitions', () => {
+  //   const sources = new Map<string, string>();
+  //   sources.set('input.ts', `function incr(n: number): number {
```
LucasSloan: Removed.

test/e2e_node_kind_source_map_test.ts
LucasSloan: Removed.

test/e2e_node_kind_source_map_test.ts
LucasSloan: Removed.

test/test_support.ts
```diff
@@ -6,15 +6,19 @@
  * found in the LICENSE file at https://angular.io/license
```
mprobst: Thanks for unifying this test support, that's much better!

PR #582, opened 2017-08-15
Update to the latest typings for source-map.

LucasSloan:
This enables removing several casts to any, since the latest typings include the constructors and fields we need to use.

### DIFF
--- a/src/source_map_utils.ts
+++ b/src/source_map_utils.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {RawSourceMap, SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import {BasicSourceMapConsumer, RawSourceMap, SourceMapConsumer, SourceMapGenerator} from 'source-map';
 import * as ts from 'typescript';
 
 /**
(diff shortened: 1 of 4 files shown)


src/source_map_utils.ts
```diff
@@ -97,16 +97,12 @@ export function sourceMapGeneratorToConsumer(
   return new SourceMapConsumer(rawSourceMap);
 }
 
-export function sourceMapTextToConsumer(sourceMapText: string): SourceMapConsumer {
-  // tslint:disable-next-line:no-any constructor actually supports text.
-  const sourceMapJson: any = sourceMapText;
-  return new SourceMapConsumer(sourceMapJson);
+export function sourceMapTextToConsumer(sourceMapText: string): BasicSourceMapConsumer {
+  return new SourceMapConsumer(sourceMapText) as BasicSourceMapConsumer;
```
evmar: Why is the 'as' needed here?

LucasSloan: It turns out it isn't - does me changing the signature to BasicSourceMapConsumer | IndexedSourceMapConsumer (matching what's returned by the constructor) alleviate your concern?

PR #584, opened 2017-08-17
update bazel rules to latest tag

alexeagle:
required so that projects with node_modules/tsickle don't conflict on the workspace name


alexeagle:
The bazel build isn't finished yet, and I haven't had time. I've been hoping the next person who works actively on tsickle will do some build system maintenance. @LucasSloan maybe?

PR #585, opened 2017-08-17
	Don't publish BUILD/WORKSPACE files in the distribution

alexeagle:
If any users want to use Bazel to build from head, they need to check out our sources, not use the published distro.


alexeagle:
Sure but the commits are good, no one but you will see the content of the PR

PR #587, opened 2017-08-18
Speed up the e2e test suite.

LucasSloan:
Improves the performance of the e2e test suite by 10x (40s to 5s on my laptop) by caching typescript's lib.d.ts and setting the types option to the empty list to ensure that typings files in node_modules aren't read.  Improves readability/understandability by removing the old compile() method used for source map tests and replacing it with a set of composable functions to generate the desired compiler options and retrieve the desired output files.  Also reduces boilerplate by introducing a function to make assertions about source maps, instead of copying and pasting 7 lines everywhere.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -124,28 +124,19 @@ function loadTscConfig(args: string[], allDiagnostics: ts.Diagnostic[]):
   return {options, fileNames};
 }
 
-export interface ClosureJSOptions {
-  tsickleCompilerHostOptions: tsickle.Options;
-  tsickleHost: tsickle.TsickleHost;
-  files: Map<string, string>;
-  tsicklePasses: tsickle.Pass[];
-}
-
-function getDefaultClosureJSOptions(fileNames: string[], settings: Settings): ClosureJSOptions {
+/**
+ * Provides a tsickle host that performs tsickle processing for the given
+ * file names.
+ *
+ * If you need to change module resolution, or only tsickle process some of the
+ * files you want to pass to tsc, you'll need to provide your own host.
+ */
+function getDefaultTsickleHost(fileNames: string[]): tsickle.TsickleHost {
   return {
-    tsickleCompilerHostOptions: {
-      googmodule: true,
-      es5Mode: false,
-      untyped: !settings.isTyped,
-    },
-    tsickleHost: {
-      shouldSkipTsickleProcessing: (fileName) => fileNames.indexOf(fileName) === -1,
-      pathToModuleName: cliSupport.pathToModuleName,
-      shouldIgnoreWarningsForPath: (filePath) => false,
-      fileNameToModuleId: (fileName) => fileName,
-    },
-    files: new Map<string, string>(),
-    tsicklePasses: [tsickle.Pass.CLOSURIZE],
+    shouldSkipTsickleProcessing: (fileName) => fileNames.indexOf(fileName) === -1,
+    pathToModuleName: cliSupport.pathToModuleName,
+    shouldIgnoreWarningsForPath: (filePath) => false,
+    fileNameToModuleId: (fileName) => fileName,
   };
 }
 
@@ -155,27 +146,25 @@ function getDefaultClosureJSOptions(fileNames: string[], settings: Settings): Cl
  */
 export function toClosureJS(
     options: ts.CompilerOptions, fileNames: string[], settings: Settings,
-    allDiagnostics: ts.Diagnostic[], partialClosureJSOptions = {} as Partial<ClosureJSOptions>):
-    {jsFiles: Map<string, string>, externs: string}|null {
-  const closureJSOptions: ClosureJSOptions = {
-    ...getDefaultClosureJSOptions(fileNames, settings),
-    ...partialClosureJSOptions
+    allDiagnostics: ts.Diagnostic[], compilerHost: ts.CompilerHost = ts.createCompilerHost(options),
+    tsickleHost: tsickle.TsickleHost = getDefaultTsickleHost(fileNames),
+    tsicklePasses: tsickle.Pass[] = [tsickle.Pass.CLOSURIZE]) {
+  const tsickleCompilerHostOptions = {
+    googmodule: true,
+    es5Mode: false,
+    untyped: !settings.isTyped,
   };
+
+  const jsFiles = new Map<string, string>();
+  const outputRetainingHost = createOutputRetainingCompilerHost(jsFiles, compilerHost);
+
+  const tch = new tsickle.TsickleCompilerHost(
+      outputRetainingHost, options, tsickleCompilerHostOptions, tsickleHost);
+
   // Parse and load the program without tsickle processing.
   // This is so:
   // - error messages point at the original source text
   // - tsickle can use the result of typechecking for annotation
-  const jsFiles = new Map<string, string>();
-  const outputRetainingHost =
-      createOutputRetainingCompilerHost(jsFiles, ts.createCompilerHost(options));
-
-  const sourceReplacingHost =
-      createSourceReplacingCompilerHost(closureJSOptions.files, outputRetainingHost);
-
-  const tch = new tsickle.TsickleCompilerHost(
-      sourceReplacingHost, options, closureJSOptions.tsickleCompilerHostOptions,
-      closureJSOptions.tsickleHost);
-
   let program = ts.createProgram(fileNames, options, tch);
   {  // Scope for the "diagnostics" variable so we can use the name again later.
     const diagnostics = ts.getPreEmitDiagnostics(program);
@@ -187,12 +176,12 @@ export function toClosureJS(
 
   // Reparse and reload the program, inserting the tsickle output in
   // place of the original source.
-  if (closureJSOptions.tsicklePasses.indexOf(tsickle.Pass.DECORATOR_DOWNLEVEL) !== -1) {
+  if (tsicklePasses.indexOf(tsickle.Pass.DECORATOR_DOWNLEVEL) !== -1) {
     tch.reconfigureForRun(program, tsickle.Pass.DECORATOR_DOWNLEVEL);
     program = ts.createProgram(fileNames, options, tch);
   }
 
-  if (closureJSOptions.tsicklePasses.indexOf(tsickle.Pass.CLOSURIZE) !== -1) {
+  if (tsicklePasses.indexOf(tsickle.Pass.CLOSURIZE) !== -1) {
     tch.reconfigureForRun(program, tsickle.Pass.CLOSURIZE);
     program = ts.createProgram(fileNames, options, tch);
   }
@@ -223,7 +212,6 @@ function main(args: string[]): number {
     return 1;
   }
 
-  // Run tsickle+TSC to convert inputs to Closure JS files.
   const closure = toClosureJS(config.options, config.fileNames, settings, diagnostics);
   if (closure === null) {
     console.error(tsickle.formatDiagnostics(diagnostics));
--- a/test/e2e_node_kind_source_map_test.ts
+++ b/test/e2e_node_kind_source_map_test.ts
@@ -14,80 +14,50 @@ import * as ts from 'typescript';
 
 import * as tsickle from '../src/tsickle';
 
-import {compile, getLineAndColumn} from './test_support';
+import {assertSourceMapping, compileWithTransfromer, findFileContentsByName, getSourceMapWithName, sourceMapCompilerOptions} from './test_support';
 
-describe.only('source maps each node with transformer', () => {
+describe('source maps each node with transformer', () => {
   it('maps import declarations correctly', () => {
     const sources = new Map<string, string>();
-    sources.set('input.ts', `import * as ts from 'typescript';
-      import {join, format as fmt} from 'path';
-      ts.ModuleKind.CommonJS;
-      join;
-      fmt;`);
+    sources.set('exporter1.ts', `export const foo = 1;`);
+    sources.set('exporter2.ts', `export const bar = 2;
+      export const baz = 3;`);
+    sources.set('input.ts', `import * as foo from './exporter1';
+      import {bar, baz as quux} from './exporter2';
+      foo.foo;
+      bar;
+      quux;`);
 
     // Run tsickle+TSC to convert inputs to Closure JS files.
-    const {compiledJS, sourceMap} = compile(sources, {useTransformer: true});
+    const {files} = compileWithTransfromer(sources, sourceMapCompilerOptions);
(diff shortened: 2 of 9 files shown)


test/e2e_tsickle_compiler_host_test.ts
```diff
@@ -85,7 +84,7 @@ describe('tsickle compiler host', () => {
 
   it('lowers decorators to annotations', () => {
     const [program, compilerHost, options] =
-        makeProgram('foo.ts', '/** @Annotation */ const A: Function = null; @A class B {}');
+        makeProgram('foo.ts', '/** @Annotation */ function A(t: any) {return t}; @A class B {}');
```
mprobst: For my information, is this just a cleanup you noticed, or do we type check the program now and didn't before? In either case, a reasonably workaround for "I need an annotation and cannot be bothered to write its gnarly type" is also `declare const A: any;`.

test/test_support.ts
```diff
@@ -6,22 +6,22 @@
...
 
 import * as cliSupport from '../src/cli_support';
 import * as es5processor from '../src/es5processor';
 import {toClosureJS} from '../src/main';
 import {sourceMapTextToConsumer} from '../src/source_map_utils';
 import * as tsickle from '../src/tsickle';
-import {createOutputRetainingCompilerHost, createSourceReplacingCompilerHost, toArray} from '../src/util';
+import {createOutputRetainingCompilerHost, toArray} from '../src/util';
 
 /** The TypeScript compiler options used by the test suite. */
```
mprobst: maybe update this comment to reflect the new scope?

test/test_support.ts
```diff
@@ -148,8 +154,15 @@ export function goldenTests(): GoldenFileTest[] {
   return tests;
 }
 
+export function getSourcesFromPaths(filePaths: string[]): Map<string, string> {
```
mprobst: maybe `readSources(filePaths)`? This could use some docs.

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
-  useTransformer: boolean;
-  /**
-   * Controls the name of the file produced by the compiler.  If there's more
-   * than one input file, they'll all be concatenated together in the outFile
-   */
-  outFile?: string;
-  filesNotToProcess: Set<string>;
-  inlineSourceMap: boolean;
-  generateDTS: boolean;
+export function assertSourceMapping(
```
mprobst: nit: the canonical order is `assert...(msg, ambientData/helpers, expected, actual)`, which in this case would (I think?) be `assertSourceMapping(sourceMap, expectedPosition, compiledJs, element)`.

also: `element` is unclear to me, maybe `sourceSnippet`?

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
+    compiledJS: string, sourceMap: SourceMapConsumer, element: string,
```
mprobst: `compiledJs`, lowerCamelCase even in abbreviations, to avoid run ins like XMLHTTPRequest (which would be XmlHttpRequest).

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
+    expectedPosition: Partial<{line: number, column: number, source: string}>) {
```
mprobst: if you're giving the type inline, why not just `expectedPosition: {line?: number, column?: number, source?: string}`?

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
+  if (expectedPosition.source) {
+    expect(originalPosition.source).to.equal(expectedPosition.source);
+  }
 }
 
-const DEFAULT_COMPILER_OPTIONS = {
-  filesNotToProcess: new Set<string>(),
-  inlineSourceMap: false,
-  generateDTS: false,
+export const sourceMapCompilerOptions: ts.CompilerOptions = {
```
mprobst: Add a comment on why/when these are useful?

test/test_support.ts
```diff
@@ -279,56 +243,63 @@ function extractInlineSourceMap(source: string): string {
...
   } while (result !== null);
   const base64EncodedMap = previousResult![1];
-  return Buffer.from(base64EncodedMap, 'base64').toString('utf8');
+  const sourceMapJson = Buffer.from(base64EncodedMap, 'base64').toString('utf8');
+  return sourceMapTextToConsumer(sourceMapJson);
 }
 
-function getFileWithName(filename: string, files: Map<string, string>): string|undefined {
-  for (const filepath of toArray(files.keys())) {
+export function getFileWithName(filename: string, files: Map<string, string>): string {
```
mprobst: given that we no longer `ts.sys.resolve` the paths, do we still need this? it seems like we should end up with human readable paths now, even for the outputs?

mprobst: `getFileContents...`? `findFileContentsByName`? This could also use docs, it's not obvious from the function name that `withName` means basename in this case.

test/test_support.ts
```diff
@@ -279,56 +243,63 @@ function extractInlineSourceMap(source: string): string {
...
+  const filePaths = toArray(files.keys());
+  let file = '';
+  for (const filepath of filePaths) {
```
mprobst: nit: you could iterate over `const [filepath, contents] of filePaths.entries())` to avoid the second lookup.

test/test_support.ts
```diff
@@ -279,56 +243,63 @@ function extractInlineSourceMap(source: string): string {
...
     if (path.parse(filepath).base === path.parse(filename).base) {
-      return files.get(filepath);
+      file = files.get(filepath)!;
```
mprobst: could you early return here, and just consistently fail below?

test/test_support.ts
```diff
@@ -279,56 +243,63 @@ function extractInlineSourceMap(source: string): string {
...
-          googmodule: true,
-          es5Mode: false,
-          untyped: false,
-        },
-        tsHost, compilerOptions);
-    allDiagnostics.push(...diagnostics);
-  }
-  return {jsFiles, diagnostics: allDiagnostics};
+  expect(ts.getPreEmitDiagnostics(program))
+      .lengthOf(0, tsickle.formatDiagnostics(ts.getPreEmitDiagnostics(program)));
```
mprobst: `.to.be.empty` is more idiomatic I think, or `.to.equal`. You should just avoid `.not.to.be.empty`, as the failure message is much better with a positive assertion.

test/test_support.ts
```diff
@@ -40,25 +40,31 @@ export const compilerOptions: ts.CompilerOptions = {
...
 
 /** Creates a ts.Program from a set of input files. */
 export function createProgram(
     sources: Map<string, string>,
-    tsCompilerOptions: ts.CompilerOptions = compilerOptions): ts.Program {
+    tsCompilerOptions: ts.CompilerOptions = goldenCompilerOptions): ts.Program {
   return createProgramAndHost(sources, tsCompilerOptions).program;
 }
 
+export function createHost(
```
mprobst: createHost calling createProgramAndHost and taking sources seems surprising. Maybe we should re-structure the call sequence to extract a caching `createHost`?

test/test_support.ts
```diff
@@ -40,25 +40,31 @@ export const compilerOptions: ts.CompilerOptions = {
 };
 
 const {cachedLibPath, cachedLib} = (() => {
-  const host = ts.createCompilerHost(compilerOptions);
-  const fn = host.getDefaultLibFileName(compilerOptions);
-  const p = ts.getDefaultLibFilePath(compilerOptions);
+  const host = ts.createCompilerHost(goldenCompilerOptions);
```
mprobst: add a comment why/if it's OK to share these even for calls that don't use the same goldenCompilerOptions?

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
-      outFile: options.outFile,
-      experimentalDecorators: true,
-      declaration: options.generateDTS,
-    };
-  }
+/**
+ * Controls the name of the file produced by the compiler.  If there's more
+ * than one input file, they'll all be concatenated together in the outFile
+ */
+export function generateOutfileCompilerOptions(outFile: string): ts.CompilerOptions {
```
mprobst: I cannot find a place where this is used?

test/test_support.ts
```diff
@@ -159,113 +172,64 @@ export function getLineAndColumn(source: string, token: string): {line: number,
...
+  return {
+    outFile,
+    module: ts.ModuleKind.None,
+  };
+}
 
-  const fileNames = toArray(sources.keys());
+export const inlineSourceMapCompilerOptions: ts.CompilerOptions = {
```
mprobst: same here re finding a usage site, I must be missing something?

test/test_support.ts
```diff
@@ -279,56 +243,63 @@ function extractInlineSourceMap(source: string): string {
...
+    tsickleHost: tsickle.TsickleHost = createTsickleHost(sources)) {
+  const fileNames = toArray(sources.keys());
   const diagnostics: ts.Diagnostic[] = [];
-  const closure =
-      toClosureJS(compilerOptions, fileNames, {isTyped: true}, diagnostics, closureJSOPtions);
-  return {jsFiles: closure ? closure.jsFiles : new Map<string, string>(), diagnostics};
+  const closure = toClosureJS(
+      compilerOptions, fileNames, {isTyped: true}, diagnostics,
+      createHost(sources, compilerOptions), tsickleHost,
+      [tsickle.Pass.DECORATOR_DOWNLEVEL, tsickle.Pass.CLOSURIZE]);
```
mprobst: should these specific options here be reflected in the function name or docs?

test/e2e_node_kind_source_map_test.ts
```diff
@@ -14,80 +14,50 @@ import * as ts from 'typescript';
 
 import * as tsickle from '../src/tsickle';
 
-import {compile, getLineAndColumn} from './test_support';
+import {assertSourceMapping, compileWithTransfromer, getFileWithName, getSourceMapWithName, sourceMapCompilerOptions} from './test_support';
 
-describe.only('source maps each node with transformer', () => {
```
LucasSloan: Yes we did.  Might be worth adding a linter for that.

test/e2e_tsickle_compiler_host_test.ts
LucasSloan: Switching to goldCompilerOptions turned on strict null checks.  And good point.

test/test_support.ts
LucasSloan: Got it.

test/test_support.ts
LucasSloan: I like the current order.  It's `(stuffToFindTheOrginalPostion, expected)`, which matches the `assert(actual).equal(expected)` order.

And sure, sourceSnippet is better.

test/test_support.ts
LucasSloan: Done.

test/test_support.ts
LucasSloan: It's used in e2e_source_map_tests, which was collapsed by the github UI as having too many changes to one file.  You should go review it!

test/test_support.ts
LucasSloan: It's used in e2e_source_map_tests, which was collapsed by the github UI as having too many changes to one file. You should go review it!

test/test_support.ts
LucasSloan: Done.

test/test_support.ts
LucasSloan: Just tried it and apparently we don't have access to iterators in tsickle.  Do you know why we're not on (I assume) a higher language level?

test/test_support.ts
LucasSloan: Renamed to `findFileContentsByName`.

LucasSloan: We sometimes specify paths to files for testing purposes, so it's still useful.

test/test_support.ts
LucasSloan: Added a doc comment.

test/test_support.ts
LucasSloan: Done.

LucasSloan:
Also, the github UI collapsed the file with the most changes - e2e_source_map_test.ts.  Could you open it up and review it?

src/main.ts
```diff
@@ -124,28 +124,12 @@ function loadTscConfig(args: string[], allDiagnostics: ts.Diagnostic[]):
...
 
-export interface ClosureJSOptions {
-  tsickleCompilerHostOptions: tsickle.Options;
-  tsickleHost: tsickle.TsickleHost;
-  files: Map<string, string>;
-  tsicklePasses: tsickle.Pass[];
-}
-
-function getDefaultClosureJSOptions(fileNames: string[], settings: Settings): ClosureJSOptions {
+function getDefaultTsickleHost(fileNames: string[]): tsickle.TsickleHost {
```
mprobst: Is there a way to explain briefly what "sensible" means? The way the comment currently reads, it doesn't add all that much information on top of the method name. If it's too hard for you to figure out what this does, let's rather keep the comment shorter.

test/test_support.ts
mprobst: OK.

If you wanted to fully follow that style, you could have an `assertSourcePosition(compiledJs, sourceMap, sourceSnippet).to.be(expectedPosition)`. I'm not particularly religious about either of these.

test/test_support.ts
mprobst: You'll need to enable downleveling iterators in our tsconfig.json, then things should work.

test/test_support.ts
mprobst: But aren't they all just local names, so that we could just do `outFiles['input.ts']`?

test/test_support.ts
LucasSloan: Ok, we use toArray a bunch of places - I'll handle that in a seperate PR.

test/test_support.ts
LucasSloan: Most of them are - I'll pull those ones out.

src/main.ts
LucasSloan: I've removed the second sentence.

test/test_support.ts
LucasSloan: Got it.

test/test_support.ts
LucasSloan: Done.

LucasSloan:
Went through and responded to all your remaining comments.

PR #588, opened 2017-08-22
Convert `@fileoverview` rewriting into a proper transformer.

mprobst:


mprobst:
@tbosch ping, please take a look! :-)

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
+import * as ts from 'typescript';
+import * as jsdoc from './jsdoc';
+
+const FILEOVERVIEW_COMMENTS: ReadonlySet<string> =
```
evmar: Can you doc?  I think these are the comments that mark a jsdoc block as a fileoverview comment?  Or are they the ones allowed in a block that must have fileoverview?

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+    new Set(['fileoverview', 'externs', 'modName', 'mods', 'pintomodule']);
+
+function skipLeadingCommentsRecursively(n: ts.SourceFile) {
```
evmar: Docs?  Seems like it sets as flag on a bunch of nodes?

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * A transformer that ensures the emitted JS file has an \@fileoverview comment that contains a
+ * \@suppress {checkTypes} annotation by either adding or updating an existing comment.
+ */
+export function transformFileoverviewComment(context: ts.TransformationContext) {
+  return (sf: ts.SourceFile) => {
+    let comments: ts.SynthesizedComment[] = [];
+    if (sf.statements.length) comments = ts.getSyntheticTrailingComments(sf.statements[0]) || [];
```
evmar: Why trailing comments here, rather than leading?

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+      ]);
+      let target = ts.createNotEmittedStatement(sf);
+      target = ts.addSyntheticTrailingComment(
+          target, ts.SyntaxKind.MultiLineCommentTrivia, commentText, true);
+      return ts.updateSourceFileNode(sf, [target, ...sf.statements]);
+    }
+
+    const comment = comments[fileoverviewIdx];
+    const parsed = jsdoc.parseContents(comment.text);
+    if (!parsed) throw new Error('internal error: JSDoc comment does not parse');
```
evmar: Is this an internal error?  I think we need to report is as diagnostics.

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+    const {tags} = parsed;
+
+    // Add @suppress {checkTypes}, or add to the list in an existing @suppress tag.
+    // Closure compiler barfs if there's a duplicated @suppress tag in a file, so the tag must
+    // only appear once and be merged.
+    const suppressIdx = tags.findIndex(t => t.tagName === 'suppress');
+    if (suppressIdx !== -1) {
```
evmar: Maybe Array.find so you don't need the index here?

src/jsdoc.ts
```diff
@@ -253,7 +264,17 @@ function tagToString(tag: Tag, escapeExtraTags = new Set<string>()): string {
 const SINGLETON_TAGS = new Set(['deprecated']);
 
 /** Serializes a Comment out to a string usable in source code. */
+export function getContents(tags: Tag[], escapeExtraTags = new Set<string>()): string {
+  return serialize(tags, false, escapeExtraTags);
+}
+
+/** Serializes a Comment out to a string usable in source code. */
```
evmar: This comment is identical to the one above, but it seems the functions do different things?

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+    // that has not been attached to some other tree node to be the file overview comment, and
+    // only applies @suppress tags from it.
+    // AJD considers *any* comment mentioning @fileoverview.
+    if (fileoverviewIdx === -1) {
+      // No existing comment to merge with, just emit a new one.
+      const commentText = jsdoc.getContents([
+        {tagName: 'fileoverview', text: 'added by tsickle'},
+        {tagName: 'suppress', type: 'checkTypes', text: 'checked by tsc'},
+      ]);
+      let target = ts.createNotEmittedStatement(sf);
```
LucasSloan: syntheticFirstStatement maybe?

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,80 @@
...
+      target = ts.addSyntheticTrailingComment(
+          target, ts.SyntaxKind.MultiLineCommentTrivia, commentText, true);
+      return ts.updateSourceFileNode(sf, [target, ...sf.statements]);
```
LucasSloan: Maybe pull this block into a function?  It can be difficult to read this code when it isn't supper clear which blocks return early and which are just doing some logic.

`return addInitialFileOverviewComment(sf);`

LucasSloan:
I also noticed that the goldens changed in different ways.  Some of them moved the file overview comment onto its own line, others just added some extra `*` only lines above and below.  Is there a pattern there you're aware of?

src/fileoverview_comment_transformer.ts
mprobst: done.

src/fileoverview_comment_transformer.ts
mprobst: actually this is unused by now. Whoops.

src/fileoverview_comment_transformer.ts
mprobst: It's what `transformer_util.ts` converts comments to. Added a (leading!) comment.

src/fileoverview_comment_transformer.ts
mprobst: done.

src/fileoverview_comment_transformer.ts
mprobst: done.

src/fileoverview_comment_transformer.ts
mprobst: this cannot happen because of the loop above. But actually, let me restructure this so that the type system can prove the condition to you.

src/fileoverview_comment_transformer.ts
mprobst: done.

src/jsdoc.ts
mprobst: Done, and renamed the function.

mprobst:
@LucasSloan I believe the difference in the tests is that the files that do not get a line break before their `/**` comment are scripts, as opposed to ES modules.

src/jsdoc.ts
```diff
@@ -139,16 +139,27 @@ export function parse(comment: string): ParsedJSDocComment|null {
...
+
+/**
+ * parseContents parses JSDoc out of a comment text.
+ * Returns null if comment is not JSDoc.
+ *
+ * @param commentText a comment's text content, i.e. the comment w/o /* and * /.
+ */
+export function parseContents(commentText: string): {tags: Tag[], warnings?: string[]}|null {
+  // Make sure we have proper line endings before parsing on Windows.
+  commentText = normalizeLineEndings(commentText);
```
evmar: I notice this Windows code is new (?), not sure if you meant to include it.

mprobst: I think Github's UI is fooling you, the normalizeLineEndings call has been here before.

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function addNewFileoverviewComment(sf: ts.SourceFile): ts.SourceFile {
+  const commentText = jsdoc.toStringWithoutStartEnd([
+    {tagName: 'fileoverview', text: 'added by tsickle'},
+    {tagName: 'suppress', type: 'checkTypes', text: 'checked by tsc'},
+  ]);
+  let syntheticFirstStatement = ts.createNotEmittedStatement(sf);
+  syntheticFirstStatement = ts.addSyntheticTrailingComment(
+      syntheticFirstStatement, ts.SyntaxKind.MultiLineCommentTrivia, commentText, true);
+  return ts.updateSourceFileNode(sf, [syntheticFirstStatement, ...sf.statements]);
```
tbosch: This might break in combination with other transformations especially decorators. Consider using `updateSourceFileNode` from `transformer_util` here if you see errors on TAP.

test_files/arrow_fn/arrow_fn.js
```diff
@@ -1,8 +1,8 @@
-goog.module('test_files.arrow_fn.arrow_fn');var module = module || {id: 'test_files/arrow_fn/arrow_fn.js'};/**
+goog.module('test_files.arrow_fn.arrow_fn');var module = module || {id: 'test_files/arrow_fn/arrow_fn.js'};
+/**
  * @fileoverview added by tsickle
  * @suppress {checkTypes} checked by tsc
  */
-
```
tbosch: I think this change is because of the missing `EmitFlags.CustomPrologue` that we are using in transformer_util.ts.

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A transformer that ensures the emitted JS file has an \@fileoverview comment that contains a
+ * \@suppress {checkTypes} annotation by either adding or updating an existing comment.
+ */
+export function transformFileoverviewComment(context: ts.TransformationContext) {
+  return (sf: ts.SourceFile) => {
+    let comments: ts.SynthesizedComment[] = [];
+    // Use trailing comments because that's what transformer_util.ts creates (i.e. by convention).
+    if (sf.statements.length) comments = ts.getSyntheticTrailingComments(sf.statements[0]) || [];
```
tbosch: you could also check that the statement is a `NonEmittedStatement`, i.e. enforcing that the original user did add an extra newline between the comment and the first statement.
E.g. match this:
```
/** @fileoverview **/

class X {}
```

But not this:
```
/** @fileoverview **/
class X {}
```

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * A set of JSDoc tags that mark a comment as a fileoverview comment. These are recognized by other
+ * pieces of infrastructure (Closure Compiler, module system, ...).
+ */
+const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
+    new Set(['fileoverview', 'externs', 'modName', 'mods', 'pintomodule']);
+
+/**
+ * A transformer that ensures the emitted JS file has an \@fileoverview comment that contains a
+ * \@suppress {checkTypes} annotation by either adding or updating an existing comment.
```
evmar: an @fileoverview vs
a @suppress
pick one of a/an

src/fileoverview_comment_transformer.ts

src/fileoverview_comment_transformer.ts

src/fileoverview_comment_transformer.ts
```diff
@@ -0,0 +1,81 @@
...
+    // only applies @suppress tags from it.
+    // AJD considers *any* comment mentioning @fileoverview.
+    if (!parsed) {
+      // No existing comment to merge with, just emit a new one.
+      return addNewFileoverviewComment(sf);
+    }
+
+    // Add @suppress {checkTypes}, or add to the list in an existing @suppress tag.
+    // Closure compiler barfs if there's a duplicated @suppress tag in a file, so the tag must
+    // only appear once and be merged.
```
evmar: In retrospect, I wonder if it woulda been better to be more strict here and reject input files that are full of nonsense.

mprobst:
I went over the PR with @tbosch in person, I think we've clarified everything.

Issue #577, opened 2017-08-09
Use goog.reflect.object when downleveling property decorators


robwormald:
In NgRx, we use decorators in a similar fashion to Angular - we "mark" class properties using a decorator, so that at runtime we can wire up the application logic:
```ts
import {Effect, Actions} from '@ngrx/store'

export class MyEffects {
  constructor(private actions: Actions){}
  @Effect() someProperty:Observable<any> = this.actions.switchMap(doSomeWork);
  
  @Effect() someOtherProperty:Observable<any> = this.actions.flatMap(doOtherWork);
}
```
These classes are then passed to the Effect library as a multi-provider:
```ts
@NgModule({
  imports: [ EffectsModule.forRoot([ MyEffects ])
})
```
and when the application launches, internally, we inspect those classes, and retrieve the marked properties (which are Observables), and subscribe to them:
```ts

const observables = getTheListOfSpecialPropertiesSomehow(MyEffects);
observables.forEach(obs => obs.subscribe(updateApplicationState);

```
Outside of Google, we do this with Reflect at runtime, and use decorators as decorators: https://github.com/ngrx/platform/blob/master/modules/effects/src/effects_metadata.ts#L38-L52 which all works fine (with the exception of this requiring the reflect-metadata polyfill, which is soon to be unnecessary in Angular core...)

Such usage of decorators is disallowed in google3, and so recently we implemented the `@Annotation` JSDoc tag, which "downlevels" the decorator to a static class property:
```ts
//output after tsickle/ngc

class MyEffect {}

MyEffect.propDecorators = {
  someProperty: [ { type: Effect, args: [] }],
  someOtherProperty: [ { type: Effect, args: [] } ]
}

```
This removes the dependency on Reflect (great!) but unfortunately, does not play nicely with closure compiler advanced optimizations, because post-property mangling, the property names captured in the downleveled annotation no longer match the (mangled) property names. 

It turns out that Closure Compiler has the ability to handle this scenario - using the `goog.reflect.object` API mentioned in the last paragraph here - https://github.com/google/closure-compiler/wiki/Type-Based-Property-Renaming. That seems to indicate we can have our cake (using annotation-style decorators from third parties) and eat it too (maintain full closure optimizations).

Other options discussed w/ @mprobst include having developers handle this case themselves (eg, by returning an array of the special properties themselves) but that adds a potential footgun and removes some of the declarative nature of NgRx. 

[...]

PR #590, opened 2017-08-23
Don't collapse propDecorators, needed for ngrx v4 effects to work

ukrukarg:
@robwormald

### DIFF
--- a/src/decorator-annotator.ts
+++ b/src/decorator-annotator.ts
@@ -228,6 +228,8 @@ export class DecoratorClassVisitor {
       this.rewriter.emit(`${className}.ctorParameters;\n`);
     }
     if (this.propDecorators) {
+      this.rewriter.emit(`/**\n`);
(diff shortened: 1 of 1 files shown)


src/decorator-annotator.ts
```diff
@@ -228,6 +228,8 @@ export class DecoratorClassVisitor {
       this.rewriter.emit(`${className}.ctorParameters;\n`);
     }
     if (this.propDecorators) {
+      this.rewriter.emit(`/**\n`);
+      this.rewriter.emit(` * @nocollapse\n`);
```
mprobst: I don't think this will work with the `/**` below, will it?

ukrukarg:
Actually, this might not be needed if https://github.com/angular/tsickle/issues/577 get's fixed properly. My solution still requires us to use @export on all effects, I guess proper solution described in that issue wouldn't.

I'm gonna hold of on this PR for now.

PR #591, opened 2017-08-28
Remove support for non-transformer based runs

mprobst:
This series of commits simplifies tsickle by removing support to *not* run in transformer mode. It also simplifies tsickle's API a bit and reduces the number of modes our tests run in.


test/golden_tsickle_test.ts
```diff
@@ -75,28 +75,11 @@ function compareAgainstGolden(output: string|null, path: string) {
 const testFn = TEST_FILTER ? describe.only : describe;
 
 testFn('golden tests with TsickleCompilerHost', () => {
```
tbosch: maybe `golden tests for tsickled sources` or something similar?

src/es5processor.ts
```diff
@@ -25,11 +25,13 @@ export interface Es5ProcessorHost {
    * for each file.
    */
   fileNameToModuleId(fileName: string): string;
-}
-
-export interface Es5ProcessorOptions {
```
tbosch: +1 on merging the `options` with the `hosts`.

src/tsickle.ts
```diff
@@ -1897,3 +1899,213 @@ export function getGeneratedExterns(externs: {[fileName: string]: string}): stri
   }
   return allExterns;
 }
+
+export interface TransformerHost extends es5processor.Es5ProcessorHost, AnnotatorHost {
```
tbosch: You could rename this back into `TsickleHost`..

test/golden_tsickle_test.ts
tbosch: Never mind, I saw that you changed it in a future commit.

src/tsickle.ts

src/tsickle.ts
```diff
@@ -59,7 +57,7 @@ export enum AnnotatorFeatures {
    * TypeScript symbol information associated. Because of this, Typescript does not
    * elide exports for types in this case.
    *
-   * This flag will be removed and always set once we drop the on transformer version of tsickle.
+   * This flag will be removed and always set once we drop the on transformer version of
```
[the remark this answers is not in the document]
mprobst: Missed this, thanks for catching!

src/tsickle.ts
```diff
@@ -1905,3 +1899,213 @@ export function getGeneratedExterns(externs: {[fileName: string]: string}): stri
...
+  externs: {[fileName: string]: string};
+}
+
+export interface EmitTransformers {
+  beforeTsickle?: Array<ts.TransformerFactory<ts.SourceFile>>;
+  beforeTs?: Array<ts.TransformerFactory<ts.SourceFile>>;
+  afterTs?: Array<ts.TransformerFactory<ts.SourceFile>>;
+}
+
+export function emitWithTsickle(
```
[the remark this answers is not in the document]
mprobst: see the commit description - it turns out `transformer.ts` and `tsickle.ts` had a cyclical value dependency. Given that we'll start removing features from `tsickle.ts` and moving them into their own transformer packages, I think this is the more sensible source organization: have the main transformation driver in `tsickle.ts`, and the individual transformers in their own files.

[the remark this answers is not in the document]
evmar: BTW to run tsickle in the TS playground we need to have no circular deps, so fixing this (by any mechanism, I didn't look at this one) is nice.

mprobst:
Green in g3, committed.

PR #602, opened 2017-09-05
fix(annotations): retain and do not quote propDecorators

robwormald:


src/decorator-annotator.ts
```diff
@@ -298,10 +301,11 @@ export class DecoratorClassVisitor {
     }
 
     if (this.propDecorators) {
+      this.rewriter.emit(`/** @nocollapse */\n`);
       this.rewriter.emit(
           `static propDecorators: {[key: string]: ` + decoratorInvocations + `} = {\n`);
       for (const name of toArray(this.propDecorators.keys())) {
-        this.rewriter.emit(`"${name}": [`);
+        this.rewriter.emit(`${name}: [`);
```
mprobst: @evmar do you remember why we needed the quotes here?

evmar: I doubt I thought too hard about it.  It's a map containing strings so I made the input strings.

PR #604, opened 2017-09-05
Fix bazel build on CircleCI

alexeagle:
Note, it's fishy that `@bazel/typescript` depends on tsickle, and this change also makes the reverse dependency.
You end up with tsickle 0.23 in the node_modules.

I don't think it breaks anything...

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,11 +1,19 @@
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 git_repository(
-    name = "build_bazel_rules_typescript",
-    remote = "https://github.com/bazelbuild/rules_typescript.git",
-    tag = "0.0.5",
+    name = "build_bazel_rules_nodejs",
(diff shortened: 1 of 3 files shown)


evmar:
I guess it sorta makes sense -- all ts compilations rely on tsickle internally, and tsickle itself is a ts compilation.  So, making sure I followed you, it would make sense for the tsickle build to depend on ts bazel rules which then depend on an older version of tsickle?

evmar:
But from that I don't follow what this CL is doing.  It should be fine for this to depend on rules_typescript directly, no?

alexeagle:
we don't even need tsickle in the codepath when we compile tsickle, it's just a smell that a version of it is present.
this does depend on rules_typescript, via the distributed package @bazel/typescript

evmar:
I still don't quite understand what this is doing.  Can you maybe make the commit message describe it?

alexeagle:
expanded the commit message.

PR #605, opened 2017-09-06
fix(): find and emit all JSDoc.

mprobst:
Previously, tsickle would silently drop JSDoc that happened to be in documents
where the JSDoc comment was shorter than the trivia at the front of the document
and the offset of the document following the JSDoc comment's length would
start with `\n\n`.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -367,12 +367,15 @@ class ClosureRewriter extends Rewriter {
     // We need to search backwards for the first JSDoc comment to avoid ignoring such when another
     // code-level comment is between that comment and the function declaration (see
     // testfiles/doc_params for an example).
-    let startPosition = 0;
+    let docRelativePos = 0;
     let parsed: jsdoc.ParsedJSDocComment|null = null;
     for (let i = comments.length - 1; i >= 0; i--) {
       const {pos, end} = comments[i];
-
-      if (end <= this.file.getStart() && this.file.text.substring(end).startsWith('\n\n')) {
+      // end is relative within node.getFullText(), add getFullStart to obtain coordinates that are
+      // comparable to node positions.
+      const docRelativeEnd = end + node.getFullStart();
+      if (docRelativeEnd <= this.file.getStart() &&
+          this.file.text.substring(docRelativeEnd).startsWith('\n\n')) {
         // This comment is at the very beginning of the file and there's an empty line between the
         // comment and this node. That means we should treat it as a file-level comment, not
         // attached to this code node.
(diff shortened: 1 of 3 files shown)


src/tsickle.ts
```diff
@@ -367,12 +367,14 @@ class ClosureRewriter extends Rewriter {
...
-    let startPosition = 0;
+    let fullStartPosition = 0;
     let parsed: jsdoc.ParsedJSDocComment|null = null;
     for (let i = comments.length - 1; i >= 0; i--) {
       const {pos, end} = comments[i];
-
-      if (end <= this.file.getStart() && this.file.text.substring(end).startsWith('\n\n')) {
+      // end is relative within node.getFullText(), add getFullStart to obtain coordinates that are
+      // comparable to node positions.
+      const fullEnd = end + node.getFullStart();
```
evmar: It's weird that you have both "fullEnd" and "end" in different coordinate systems, while "start" vs "fullStart" on a node are two positions in the same coordinate system.

Maybe you mean a different word than "full" here?  Like perhaps "fileEnd" or some such?  Or maybe rename "end" to "len" or something?

Same with renaming startPosition to fullStartPosition, it seems like the wrong name.  "full" means whether it includes trivia or not.

mprobst: Good point. I changed both to `docRelative*` - does that help?

test_files/functions/two_jsdoc_blocks.js
```diff
@@ -0,0 +1,22 @@
+goog.module('test_files.functions.two_jsdoc_blocks');var module = module || {id: 'test_files/functions/two_jsdoc_blocks.js'};/**
+ * @fileoverview This text here matches the     text below in length.
```
LucasSloan: This comment looks wrong.  Both of the functions seems to have Jsdoc comments that are shorter than this.

.vscode/launch.json
```diff
@@ -38,7 +38,7 @@
       ],
       "env": {
         "NODE_ENV": "development",
-        "TEST_FILTER": "^file_comment$"
+        "TEST_FILTER": "^functions$"
```
LucasSloan: What does this change do?  I'm suspicious that a dot file has something important for us to check in...

mprobst: Result of an accidental `git commit -am ...`. 

If you don't know what this file does, we should sit down and see how your workflow looks. This is a launch description for VS Code that allows you to debug tsickle from within VS Code very conveniently. This line here is the line that picks the test to debug.

test_files/functions/two_jsdoc_blocks.js
mprobst: The second function's document is exactly 77 chars, as is this one.

PR #606, opened 2017-09-06
Remove unnecessary usages of toArray().

LucasSloan:
We were using toArray to use iterators in for of loops.  Since we've set the flag that has TypeScript downlevel them for us, we don't need to do it manually, so they're redundant.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -8,7 +8,6 @@
 
 import * as path from 'path';
 import * as ts from 'typescript';
-import {toArray} from './util';
 
 /**
  * Determines if fileName refers to a builtin lib.d.ts file.
@@ -532,7 +531,10 @@ export class TypeTranslator {
       return `function(new: (${constructedType})${paramsStr}): ?`;
     }
 
-    for (const field of toArray(type.symbol.members.keys())) {
+    // members is an ES6 map, but the .d.ts defining it defined their own map
+    // type, so typescript doesn't believe that .keys() is iterable
+    // tslint:disable-next-line:no-any
+    for (const field of (type.symbol.members.keys() as any)) {
       switch (field) {
         case '__call':
           callable = true;
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -18,7 +18,6 @@ import * as es5processor from '../src/es5processor';
 import {toClosureJS} from '../src/main';
 import {sourceMapTextToConsumer} from '../src/source_map_utils';
 import * as tsickle from '../src/tsickle';
-import {toArray} from '../src/util';
 
 /** Base compiler options to be customized and exposed. */
 const baseCompilerOptions: ts.CompilerOptions = {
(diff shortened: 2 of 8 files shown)


LucasSloan:
Unfortunately, no.  We also use it when something expects an array, but we want to provide an iterator (ie we want to pass in the keys of a map as an array).

evmar:
Can we use Array.from() there?

LucasSloan:
Yes we can, thanks!

test/test_support.ts
```diff
@@ -253,13 +253,12 @@ export function extractInlineSourceMap(source: string): BasicSourceMapConsumer {
...
 
 export function findFileContentsByName(filename: string, files: Map<string, string>): string {
-  const filePaths = toArray(files.keys());
-  for (const filepath of filePaths) {
+  for (const filepath of files.keys()) {
     if (path.parse(filepath).base === path.parse(filename).base) {
       return files.get(filepath)!;
     }
   }
-  assert(undefined, `Couldn't find file ${filename} in files: ${JSON.stringify(filePaths)}`);
```
evmar: I think the reason for the JSON.stringify here is that it quotes the elements of the list.  I think we want to keep that.

files.toString()
"a,b.ts"
JSON.stringify(files)
"["a","b.ts"]"

src/type-translator.ts
```diff
@@ -532,7 +531,8 @@ export class TypeTranslator {
       return `function(new: (${constructedType})${paramsStr}): ?`;
     }
 
-    for (const field of toArray(type.symbol.members.keys())) {
+    // tslint:disable-next-line:no-any
+    for (const field of (type.symbol.members.keys() as any)) {
```
evmar: Why is this needed?  Isn't members already a Map?

LucasSloan: It's a ts.Map, which wants to be a real Map, but isn't for some reason.

test/test_support.ts
LucasSloan: Done.

src/type-translator.ts
evmar: It looks like it's just named Map, but it's not actually compatible with ES6.  I would guess this code doesn't work as written.

file: 'file:///usr/local/google/home/evanm/projects/tsickle/src/type-translator.ts'
severity: 'Error'
message: 'Type 'Iterator<string>' is not an array type or a string type or does not have a '[Symbol.iterator]()' method that returns an iterator.'
at: '535,21'
source: 'ts'

LucasSloan: The comments on the type say it's an ES6 map.  I'm pretty sure that the object is a proper map, they just produced a bad .d.ts for some reason.  All the tests pass, which wouldn't be the case if this threw an error instead of doing the right thing.

evmar: Can you add a comment explaining this?

LucasSloan: Done.

PR #607, opened 2017-09-06
Run decorator-annotator unit tests through the transformer codepath.

LucasSloan:
Reworks the decorator-annotator unit tests to run in transformer mode so that we can migrate code out of the psuedo-tranformer into a proper transformer and be assured that the behavior hasn't changed.  Incidentally removes the function `visitEachChildIgnoringTypes`, which was used to work around a TypeScript bug that was fixed in 2.4, and made it impossible to properly pretty print the AST after all our transforms ran.

### DIFF
--- a/src/transformer_util.ts
+++ b/src/transformer_util.ts
@@ -605,18 +605,6 @@ export function updateSourceFileNode(
   return sf;
 }
 
-/**
- * This is a version of `ts.visitEachChild` that does not visit children of types,
- * as this leads to errors in TypeScript < 2.4.0 and as types are not emitted anyways.
- */
-export function visitEachChildIgnoringTypes<T extends ts.Node>(
-    node: T, visitor: ts.Visitor, context: ts.TransformationContext): T {
-  if (isTypeNodeKind(node.kind) || node.kind === ts.SyntaxKind.IndexSignature) {
-    return node;
-  }
-  return ts.visitEachChild(node, visitor, context);
-}
-
 // Copied from TypeScript
 export function isTypeNodeKind(kind: ts.SyntaxKind) {
   return (kind >= ts.SyntaxKind.FirstTypeNode && kind <= ts.SyntaxKind.LastTypeNode) ||
--- a/test/ast_printing_transform.ts
+++ b/test/ast_printing_transform.ts
@@ -1,0 +1,15 @@
+import * as ts from 'typescript';
+
+/**
+ * Creates a transformer that captures the state of the AST as pretty printed
+ * TypeScript in the provided 'files' map.
+ */
+export function createAstPrintingTransform(files: Map<string, string>) {
+  return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
+    return (node: ts.SourceFile): ts.SourceFile => {
+      const sf = node as ts.SourceFile;
+      files.set(sf.fileName, ts.createPrinter().printFile(sf));
+      return sf;
+    };
+  };
+}
--- a/test/decorator-annotator_test.ts
+++ b/test/decorator-annotator_test.ts
@@ -12,196 +12,257 @@ import {expect} from 'chai';
 import {SourceMapConsumer} from 'source-map';
 import * as ts from 'typescript';
 
+import * as cliSupport from '../src/cli_support';
 import {convertDecorators} from '../src/decorator-annotator';
 import {DefaultSourceMapper} from '../src/source_map_utils';
 import * as tsickle from '../src/tsickle';
 
-import * as testSupport from './test_support';
+import {createAstPrintingTransform} from './ast_printing_transform';
+import {compilerOptions, createProgram, createProgramAndHost} from './test_support';
 
 const testCaseFileName = 'testcase.ts';
 
 function sources(sourceText: string): Map<string, string> {
   const sources = new Map<string, string>([
     [testCaseFileName, sourceText],
-    ['bar.d.ts', 'declare module "bar" { export class BarService {} }']
+    // Provides a rename of any 'FakeDecorator' that we can use as an annotator
+    // without the compiler complaining we didn't actually provide a value
+    [
+      'bar.d.ts', `declare module "bar" {
+      export class BarService {}
+      type FakeDecorator = any;
+    }`
+    ]
   ]);
   return sources;
 }
 
-function verifyCompiles(sourceText: string) {
-  // This throws an exception on error.
-  testSupport.createProgram(sources(sourceText));
+describe('decorator-annotator', () => {
+  function translate(sourceText: string, allowErrors = false) {
+    const {host, program} = createProgramAndHost(sources(sourceText), compilerOptions);
+    if (!allowErrors) {
+      const diagnostics = ts.getPreEmitDiagnostics(program);
+      expect(diagnostics, tsickle.formatDiagnostics(ts.getPreEmitDiagnostics(program))).to.be.empty;
+    }
+
+    const transformerHost: tsickle.TsickleHost = {
+      shouldSkipTsickleProcessing: (filePath) => !sources(sourceText).has(filePath),
+      pathToModuleName: cliSupport.pathToModuleName,
+      shouldIgnoreWarningsForPath: (filePath) => false,
+      fileNameToModuleId: (filePath) => filePath,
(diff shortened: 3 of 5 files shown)


test/decorator-annotator_test.ts
```diff
@@ -28,68 +30,82 @@ function sources(sourceText: string): Map<string, string> {
...
 
-function verifyCompiles(sourceText: string) {
-  // This throws an exception on error.
-  testSupport.createProgram(sources(sourceText));
-}
+describe('decorator-annotator', () => {
+  function translate(sourceText: string, allowErrors = false) {
+    const {host, program} = createProgramAndHost(sources(sourceText), compilerOptions);
+    if (!allowErrors) {
+      expect(ts.getPreEmitDiagnostics(program))
```
mprobst: can you follow the `to.be.empty` pattern we use below? I might be misremembering, but I think I explicitly removed this pattern before because it doesn't actually assert and fail. In any case, `to.be.empty` is more idiomatic.

test/decorator-annotator_test.ts
```diff
@@ -28,68 +30,82 @@ function sources(sourceText: string): Map<string, string> {
...
+    return sourceText.replace(/"/g, `'`);
+  }
+
+  function expectUnchanged(sourceText: string) {
+    expectEqual(sourceText, sourceText);
+  }
+
+  function expectEqual(sourceText: string, expected: string) {
+    expect(normalizeQuotes(translate(sourceText).output))
+        .to.equal(normalizeQuotes(prettyPrint(expected)));
```
mprobst: can't you make the expected you're passing in here match the pretty print? That'd safe some code and avoid a potential pitfall/source of error.

test/decorator-annotator_test.ts
```diff
@@ -28,68 +30,82 @@ function sources(sourceText: string): Map<string, string> {
...
-        });
+    });
 
-        it('transforms decorated classes', () => {
-          expect(translate(`
-/** @Annotation */ let Test1: Function;
-/** @Annotation */ let Test2: Function;
+    it('transforms decorated classes', () => {
+      expectEqual(`
+/** @Annotation */ let Test1: any;
```
mprobst: it might be worth documenting somewhere here that we use `any` to fake out a decorator. Maybe you could define `type FakeDecorator = any` in the .d.ts file in sources above, and leave a comment there?

test/decorator-annotator_test.ts
```diff
@@ -28,68 +30,82 @@ function sources(sourceText: string): Map<string, string> {
...
+
+  function normalizeQuotes(sourceText: string) {
+    return sourceText.replace(/"/g, `'`);
+  }
+
+  function expectUnchanged(sourceText: string) {
+    expectEqual(sourceText, sourceText);
+  }
+
+  function expectEqual(sourceText: string, expected: string) {
```
mprobst: I think it's more idiomatic in Jasmine to return an expectation from helper methods like this. E.g. `expectWobbled(...).to.equal(...);`

test/ast_printing_transform.ts
```diff
@@ -0,0 +1,22 @@
+import * as ts from 'typescript';
+
+/**
+ * Creates a transformer that captures the state of the AST as pretty printed
+ * TypeScript in the provided 'files' map.
+ */
+export function generateAstPrintingTransform(files: Map<string, string>) {
```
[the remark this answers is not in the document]
LucasSloan: Done.

test/ast_printing_transform.ts
```diff
@@ -0,0 +1,22 @@
...
+  return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
+    const visitor: ts.Visitor = (node: ts.Node): ts.Node => {
+      switch (node.kind) {
+        case ts.SyntaxKind.SourceFile:
+          const sf = node as ts.SourceFile;
+          files.set(sf.fileName, ts.createPrinter().printFile(sf));
+          return ts.visitEachChild(node, visitor, context);
```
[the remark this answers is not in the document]
LucasSloan: I was doing that when I was looking at other nodes for debugging, but yes, no longer necessary.

test/decorator-annotator_test.ts
LucasSloan: Done.

test/decorator-annotator_test.ts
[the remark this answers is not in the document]
LucasSloan: Renamed to `expectTranslatedToEqual`;

test/decorator-annotator_test.ts
LucasSloan: I'd be happy to do that in a follow up PR, but for this one, I wanted the golden diff to be as small as possible so we could actually be confident that the tests were still asserting the right stuff.

test/decorator-annotator_test.ts
LucasSloan: Understood, but see below for why it isn't done that way.

test/decorator-annotator_test.ts
LucasSloan: Done.

src/transformer_util.ts
```diff
@@ -617,18 +617,6 @@ export function getMutableClone<T extends ts.Node>(node: T): T {
   return clone;
 }
 
-/**
- * This is a version of `ts.visitEachChild` that does not visit children of types,
- * as this leads to errors in TypeScript < 2.4.0 and as types are not emitted anyways.
- */
-export function visitEachChildIgnoringTypes<T extends ts.Node>(
```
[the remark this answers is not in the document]
LucasSloan: Done.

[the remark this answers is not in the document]
tbosch: Did you already run a (global) presubmit to make sure we really don't need this?

test/ast_printing_transform.ts
```diff
@@ -0,0 +1,22 @@
...
+export function createAstPrintingTransform(files: Map<string, string>) {
+  return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
+    const visitor: ts.Visitor = (node: ts.Node): ts.Node => {
+      switch (node.kind) {
+        case ts.SyntaxKind.SourceFile:
+          const sf = node as ts.SourceFile;
+          files.set(sf.fileName, ts.createPrinter().printFile(sf));
+          return sf;
+        default:
+          return ts.visitEachChild(node, visitor, context);
```
tbosch: When will this case be hit? As you are passing in a `ts.SourceFile` to `ts.visitNode`, which will then call the `visitor` with a `ts.SourceFile` again...

tbosch:
In general, note that the unit tests for the decorator downlevel are far from capturing the important cases for transformers. E.g. lowering something that was only used in a type position but imported from another file, ...

src/transformer_util.ts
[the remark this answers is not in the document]
LucasSloan: I have not - I was relying on the comment.  Will do.

test/ast_printing_transform.ts
LucasSloan: I don't think this will ever be hit, but we need a default case to make the type checker happy.

tbosch: how about throwing?

tbosch: Or even better, don't use `ts.visitNode` at all, call the visitor directly, and type it to receive and return a `ts.SourceFile`. I.e. just return the visitor:
```
return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
  return (node: ts.SourceFile): ts.Sourcefile => {
    const sf = node as ts.SourceFile;
    files.set(sf.fileName, ts.createPrinter().printFile(sf));
    return sf;  
  };
};
```

src/transformer_util.ts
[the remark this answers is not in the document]
LucasSloan: Presubmit looks good!

test/ast_printing_transform.ts
LucasSloan: That works, thanks!

LucasSloan:
Understood, I'm not just relying on the unit tests, I'm also working on the golden tests, and I'll make sure to run a presubmit and add tests for anything that catches.

PR #608, opened 2017-09-08
Smaller fixes for Angular's new compiler

tbosch:


src/tsickle.ts
```diff
@@ -2129,3 +2125,7 @@ function combineSourceMaps(
   }
   return tscSourceMapGenerator ? tscSourceMapGenerator.toString() : tscSourceMapText;
 }
+
+function normalizeArray<T>(arr: T[]|undefined): T[] {
+  return arr ? arr : [];
+}
```
evmar: Maybe include trailing newline here?

src/tsickle.ts
```diff
@@ -2129,3 +2125,7 @@ function combineSourceMaps(
   }
   return tscSourceMapGenerator ? tscSourceMapGenerator.toString() : tscSourceMapText;
 }
+
+function normalizeArray<T>(arr: T[]|undefined): T[] {
```
evmar: Is this the same as writing
 [
    ...(foo || []),
   ...(bar || []),
]

above?  It might make more sense to inline, I dunno.

tbosch: Yes, good point.

src/tsickle.ts
tbosch: Removed this function, see above.

PR #610, opened 2017-09-11
Move processing of decorators on classes into its own transformer.

LucasSloan:
Takes the code for downleveling decorators on classes out of decorator-annotator, and moves it into its own transformer, which is run after the monolithic pseudo-transformer.

### DIFF
--- a/src/class_decorator_downlevel_transformer.ts
+++ b/src/class_decorator_downlevel_transformer.ts
@@ -1,0 +1,136 @@
+import * as ts from 'typescript';
+
+import {shouldLower} from './decorator-annotator';
+import {visitEachChild} from './transformer_util';
+
+/**
+ * Creates the AST for the decorator field type annotation, which has the form
+ * { type: Function, args?: any[] }[]
+ */
+function createClassDecoratorType() {
+  const typeElements: ts.TypeElement[] = [];
+  typeElements.push(ts.createPropertySignature(
+      undefined, 'type', undefined,
+      ts.createTypeReferenceNode(ts.createIdentifier('Function'), undefined), undefined));
+  typeElements.push(ts.createPropertySignature(
+      undefined, 'args', ts.createToken(ts.SyntaxKind.QuestionToken),
+      ts.createArrayTypeNode(ts.createKeywordTypeNode(ts.SyntaxKind.AnyKeyword)), undefined));
+  return ts.createArrayTypeNode(ts.createTypeLiteralNode(typeElements));
+}
+
+/**
+ * Extracts the type of the decorator, as well as all the arguments passed to
+ * the decorator.  Returns an AST with the form
+ * { type: decorator, args: [arg1, arg2] }
+ */
+function extractMetadataFromSingleDecorator(decorator: ts.Decorator, diagnostics: ts.Diagnostic[]) {
+  const metadataProperties: ts.ObjectLiteralElementLike[] = [];
+  const expr = decorator.expression;
+  switch (expr.kind) {
+    case ts.SyntaxKind.Identifier:
+      // The decorator was a plain @Foo.
+      metadataProperties.push(ts.createPropertyAssignment('type', expr));
+      break;
+    case ts.SyntaxKind.CallExpression:
+      // The decorator was a call, like @Foo(bar).
+      const call = expr as ts.CallExpression;
+      metadataProperties.push(ts.createPropertyAssignment('type', call.expression));
+      if (call.arguments.length) {
+        const args: ts.Expression[] = [];
+        for (const arg of call.arguments) {
+          args.push(arg);
+        }
+        const argsArrayLiteral = ts.createArrayLiteral(args);
+        argsArrayLiteral.elements.hasTrailingComma = true;
+        metadataProperties.push(ts.createPropertyAssignment('args', argsArrayLiteral));
+      }
+      break;
+    default:
+      diagnostics.push({
+        file: decorator.getSourceFile(),
+        start: decorator.getStart(),
+        length: decorator.getEnd() - decorator.getStart(),
+        messageText:
+            `${ts.SyntaxKind[decorator.kind]} not implemented in gathering decorator metadata`,
+        category: ts.DiagnosticCategory.Error,
+        code: 0,
+      });
+      break;
+  }
+  return ts.createObjectLiteral(metadataProperties);
+}
+
+/**
+ * Takes a list of decorator metadata object ASTs and produces an AST for a
+ * static class property of an array of those metadata objects.
+ */
+function createDecoratorClassProperty(decoratorList: ts.ObjectLiteralExpression[]) {
+  const modifier = ts.createToken(ts.SyntaxKind.StaticKeyword);
+  const type = createClassDecoratorType();
+  const initializer = ts.createArrayLiteral(decoratorList, true);
+  initializer.elements.hasTrailingComma = true;
+  return ts.createProperty(undefined, [modifier], 'decorators', undefined, type, initializer);
+}
+
+function isNameEqual(classMember: ts.ClassElement, name: string): boolean {
+  if (classMember.name === undefined) {
+    return false;
+  }
+  const id = classMember.name as ts.Identifier;
+  return id.text === name;
+}
+
+/**
+ * Inserts the decorator metadata property in the place that the old
+ * decorator-annotator visitor would put it, so the unit tests don't have to
+ * change.
+ * TODO(lucassloan): remove this when all 3 properties are put in via
+ * transformers
+ */
+function insertBeforeDecoratorProperties(
+    classMembers: ts.NodeArray<ts.ClassElement>, decoratorMetadata: ts.PropertyDeclaration) {
+  let insertionPoint = classMembers.findIndex(
+      m => isNameEqual(m, 'ctorParameters') || isNameEqual(m, 'propDecorators'));
+  if (insertionPoint === -1) {
+    insertionPoint = classMembers.length - 1;
+  }
+  classMembers.splice(insertionPoint, 0, decoratorMetadata);
+}
+
+export function classDecoratorDownlevelTransformer(
+    typeChecker: ts.TypeChecker, diagnostics: ts.Diagnostic[]):
+    (context: ts.TransformationContext) => ts.Transformer<ts.SourceFile> {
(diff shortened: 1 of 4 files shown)


tbosch:
I think this is missing the logic for storing constructor parameters too. Is your plan to do this in a separate run?

LucasSloan:
This is just for providing the `decorators` parameter, the `ctorParameters` and `propDecorators` are coming later.

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
+import * as ts from 'typescript';
+
+import {DecoratorClassVisitor} from './decorator-annotator';
+
+function createClassDecoratorType() {
```
mprobst: docs

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+        for (const arg of call.arguments) {
+          args.push(arg);
+        }
+        const argsArrayLiteral = ts.createArrayLiteral(args);
+        argsArrayLiteral.elements.hasTrailingComma = true;
+        metadataProperties.push(ts.createPropertyAssignment('args', argsArrayLiteral));
+      }
+      break;
+    default:
+      throw new Error(`Unimplemented kind: ${ts.SyntaxKind[expr.kind]}`);
```
mprobst: do not throw, report diagnostics, with file, line, etc.

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+  typeElements.push(ts.createPropertySignature(
+      undefined, 'type', undefined,
+      ts.createTypeReferenceNode(ts.createIdentifier('Function'), undefined), undefined));
+  typeElements.push(ts.createPropertySignature(
+      undefined, 'args', ts.createToken(ts.SyntaxKind.QuestionToken),
+      ts.createArrayTypeNode(ts.createKeywordTypeNode(ts.SyntaxKind.AnyKeyword)), undefined));
+  return ts.createArrayTypeNode(ts.createTypeLiteralNode(typeElements));
+}
+
+function extractMetadataFromSingleDecorator(decorator: ts.Decorator) {
```
mprobst: docs

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+
+function isNameEqual(classMember: ts.ClassElement, name: string): boolean {
+  if (classMember.name === undefined) {
+    return false;
+  }
+  const id = classMember.name as ts.Identifier;
+  return id.text === name;
+}
+
+function insertBeforeDecoratorProperties(
```
mprobst: why is this important - document?

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+    classMembers: ts.NodeArray<ts.ClassElement>, decoratorMetadata: ts.PropertyDeclaration) {
+  const ctorParameterIndex = classMembers.findIndex(m => isNameEqual(m, 'ctorParameters'));
```
mprobst: why not just search for isNameEqual('ctorParams') || isNameEqual('propDecorators'), then use that index or 0?

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+          }
+
+          const decoratorList = [];
+          for (const decorator of decorators) {
+            if (DecoratorClassVisitor.shouldLower(decorator, typeChecker)) {
+              decoratorList.push(extractMetadataFromSingleDecorator(decorator));
+            }
+          }
+
+          if (decoratorList.length === 0) {
```
mprobst: consider single line ifs

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+  classMembers.splice(insertionPoint, 0, decoratorMetadata);
+}
+
+export function transformer(typeChecker: ts.TypeChecker): (context: ts.TransformationContext) =>
+    ts.Transformer<ts.SourceFile> {
+  return (context: ts.TransformationContext) => {
+    const visitor: ts.Visitor = (node: ts.Node): ts.Node => {
+      switch (node.kind) {
+        case ts.SyntaxKind.ClassDeclaration:
+          const cd = ts.visitEachChild(node, visitor, context) as ts.ClassDeclaration;
```
mprobst: Why is `cd` mutable here? Wouldn't you normally need to create a mutable clone?

src/class_decorator_downlevel_transformer.ts
```diff
@@ -0,0 +1,119 @@
...
+
+          cd.decorators = undefined;
+
+          return cd;
+        default:
+          return ts.visitEachChild(node, visitor, context);
+      }
+    };
+
+    const transformer: ts.Transformer<ts.SourceFile> = (sf: ts.SourceFile) =>
```
mprobst: If you don't need the `this` pointer, just make it a nested function?

mprobst: you have too many things called `transformer`

src/decorator-annotator.ts
```diff
@@ -52,8 +52,8 @@ export class DecoratorClassVisitor {
   /**
    * Determines whether the given decorator should be re-written as an annotation.
    */
-  private shouldLower(decorator: ts.Decorator) {
-    for (const d of getDecoratorDeclarations(decorator, this.typeChecker)) {
+  public static shouldLower(decorator: ts.Decorator, typeChecker: ts.TypeChecker) {
```
mprobst: make this a top level function if it doesn't access state from the class? static class level functions don't really make a lot of sense in TS.

src/decorator-annotator.ts
```diff
@@ -189,8 +189,13 @@ export class DecoratorClassVisitor {
     }
   }
 
+  isClassDecorator(decorator: ts.Decorator): boolean {
```
mprobst: private?

mprobst: docs! why does this method exist?

src/tsickle.ts
```diff
@@ -2005,6 +2006,7 @@ export function emitWithTsickle(
     before: [
       ...(customTransformers.beforeTsickle || []),
       ...(tsickleTransformers.before || []).map(tf => skipTransformForSourceFileIfNeeded(host, tf)),
+      classDecoratorDownlevelTransformer(typeChecker),
```
mprobst: does this transformer need the `skipTransformForSourceFileIfNeeded` check?

LucasSloan:
And the presubmit looks good.

src/class_decorator_downlevel_transformer.ts
mprobst: Are you sure? Wouldn't that mean ts.visitEachChild will always copies the entire tree?

The docs say: `possibly returning a new Node of the same kind in its place` - that doesn't sounds as if you can rely on that behaviour, does it?

src/decorator-annotator.ts
```diff
@@ -28,6 +28,38 @@ interface ConstructorParameter {
   decorators: ts.Decorator[]|null;
 }
 
+export function shouldLower(decorator: ts.Decorator, typeChecker: ts.TypeChecker) {
+  for (const d of getDecoratorDeclarations(decorator, typeChecker)) {
+    // Switch to the TS JSDoc parser in the future to avoid false positives here.
```
mprobst: While you're at it, should this be a TODO() comment?

src/decorator-annotator.ts
```diff
@@ -28,6 +28,38 @@ interface ConstructorParameter {
...
+    const range = ts.getLeadingCommentRanges(commentNode.getFullText(), 0);
+    if (!range) continue;
+    for (const {pos, end} of range) {
+      const jsDocText = commentNode.getFullText().substring(pos, end);
+      if (jsDocText.includes('@Annotation')) return true;
+    }
+  }
+  return false;
+}
+
```
mprobst: nit: obscure line breaks?

src/decorator-annotator.ts
```diff
@@ -189,13 +191,19 @@ export class DecoratorClassVisitor {
     }
   }
 
-  isClassDecorator(decorator: ts.Decorator): boolean {
+  /**
+   * Checks if the decorator is on a class, as opposed to a field or an
```
mprobst: your editor seems to be wrapping at 80, is that intentional? Shouldn't it be 100?

src/class_decorator_downlevel_transformer.ts
LucasSloan: You're right, we do need to call `ts.getMutableClone` - I talked to Tobias, and it would cause issues with our blaze source-file caching.

src/decorator-annotator.ts
LucasSloan: Done.

src/class_decorator_downlevel_transformer.ts
LucasSloan: Done.

src/decorator-annotator.ts
LucasSloan: I insert the line breaks manually.

src/decorator-annotator.ts
LucasSloan: Done.

test_files/decorator/decorator.js
```diff
@@ -85,7 +85,7 @@ __decorate([
 ], DecoratorTest.prototype, "x", void 0);
 __decorate([
     decorator,
-    __metadata("design:type", external_1.AClass)
+    __metadata("design:type", Object)
```
mprobst: As discussed offline, we'll need to make sure not to regress here for non-aot mode usage. Try using the update... methods in `transformer_util.ts` instead of the builtin ones, one of them has a fix for this behaviour.

src/transformer_util.ts
```diff
@@ -582,6 +582,24 @@ function getAllLeadingCommentRanges(
 }
 
 /**
+ * This is a version of `ts.visitEachChild` that works that calls our version
+ * of `updateSourceFileNode`, so that property decorators work.
```
mprobst: can you be a bit more specific with the "property decorators work"?

PR #612, opened 2017-09-14
Add license blocks.

LucasSloan:
g3 third_party requires code we check in to have license blocks, so this adds license blocks to the two new transformer files we've added.


LucasSloan:
That sounds like a great idea - will do!

PR #616, opened 2017-09-20
Handle @fileoverview in all initial comments.

LucasSloan:
The fileoverview comment transformer handled @fileoverview comments that weren't attached to any
node, but @fileoverview comments just have to be before any js, so they can be in leading comments on the first statement in the file.  Thus, we need to look for @fileoverview in a leading NotEmittedStatement's comments as well as in the leading comments on the first emitted statement.  To avoid handling comments that were on a node that isn't emitted (ie an interface), move the fileoverview transformer to run after tsc runs its compile.

### DIFF
--- a/test_files/file_comment/comment_with_no_newline.ts
+++ b/test_files/file_comment/comment_with_no_newline.ts
@@ -1,0 +1,12 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+/**
+ * @fileoverview
+ * @suppress {missingRequire}
+ */
+console.log('hello');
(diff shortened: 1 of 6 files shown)


test_files/file_comment/comment_with_no_newline.ts
```diff
@@ -0,0 +1,12 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+/**
+ * @fileoverview
+ * @suppress {missingRequire}
+ */
```
evmar: I think we use tslint to disallow this kind of input (with the missing newline), don't we?

mprobst:
I actually think the tslint workaround is sort of a hack, and we'd be better off if we'd always find and emit these comments. That was previously near impossible due to the way comments were handled - in particular, we'd occasionally swallow the comments if they were preceding a not emitted statement, which is of course bad UX.

If you can find a way to do this consistently regardless of line breaks after the comment, even for not-emitted code (interfaces), I'd be in favour.

mprobst:
I think this is obsolete, right?

PR #618, opened 2017-09-26
Emit `@suppress {duplicate}` on the statement level, and not on the file level.

tbosch:


src/tsickle.ts
```diff
@@ -1628,8 +1628,13 @@ class ExternsWriter extends ClosureRewriter {
         this.writeExternsFunction(name, params, namespace);
         break;
       case ts.SyntaxKind.VariableStatement:
-        for (const decl of (node as ts.VariableStatement).declarationList.declarations) {
-          this.writeExternsVariableDecl(decl, namespace);
+        const varStmt = node as ts.VariableStatement;
+        // Note: This is needed as e.g. suppressing duplicates for variable declarations
+        // only wors when the suppress tag is on the variable statement.
```
evmar: typo: works

src/tsickle.ts
```diff
@@ -1760,12 +1765,13 @@ class ExternsWriter extends ClosureRewriter {
...
 
-  private writeExternsVariableDecl(decl: ts.VariableDeclaration, namespace: string[]) {
+  private writeExternsVariableDecl(
+      decl: ts.VariableDeclaration, namespace: string[], extraTags: jsdoc.Tag[] = []) {
     if (decl.name.kind === ts.SyntaxKind.Identifier) {
       const name = getIdentifierText(decl.name as ts.Identifier);
       if (closureExternsBlacklist.indexOf(name) >= 0) return;
-      this.emitJSDocType(decl);
-      this.emit('\n');
+      const type = this.typeToClosure(decl);
```
evmar: The reason emitJSDocType emits whitespace first is that it is replacing the existing trivia on the node.  If you had input like `foo\n\nbar` and we add jsdoc to `bar`, we need some whitespace in place of the `\n`.  I think in the case of externs we are emitting a line at a time (and jsdoc.toString inserts a newline) so your new layout is correct.  In retrospect we probably should've attempted to replicate the whitespace of whatever content we're deleting, but at the time we wrote this we didn't think the output needed to be very human-readable.

test_files/declare/declare.d.ts
```diff
@@ -51,6 +51,11 @@ declare module "DeclareTest-QuotedModule" {
 declare var declareGlobalVar: number;
 declare function declareGlobalFunction(x: string): number;
 
+/**
+ * @suppress {duplicate}
+ */
+declare var declareGlobalVarWithSuppress: number;
```
evmar: My recollection is that even with suppress duplicate, if the types don't match it's an unsuppressable compiler error.  So usually using suppress won't save you.

Also, I experimented on go/jscomp-full and it appears a file-level suppress does work if the first block is tagged @externs.  Maybe we should just always emit that?

src/tsickle.ts
```diff
@@ -49,18 +49,32 @@ export interface AnnotatorHost {
...
+ * @suppress {checkTypes}
  */
 // NOTE: generated by tsickle, do not edit.
 `;
 
 /**
+ * Always suppress duplicate definitions at the variable declaration,
+ * and not at the file level. This allows to prevent errors from duplicates
+ * no matter in which order closure evaluates the externs in the case
+ * that the colliding extern has no @suppress {duplicate} annotation.
```
evmar: Can you clarify this comment a bit about what we discovered?  That is, something about "if the @suppress is at the file level, then it doesn't affect other instances of the variable in other files, while apparently applying it to the variable does".

Also, I wonder if we should just put the checktypes suppression in here too, so that they're all in one place?  What do you think?

src/tsickle.ts
```diff
@@ -1691,7 +1702,12 @@ class ExternsWriter extends ClosureRewriter {
         jsdocTags.push({tagName: 'record'});
         jsdocTags.push({tagName: 'struct'});
       }
-      if (writeJsDoc) this.emit(jsdoc.toString(jsdocTags));
+      if (writeJsDoc) {
+        jsdocTags.push(SUPPRESS_DUPLICATE_TAG);
+        this.emit(jsdoc.toString(jsdocTags));
+      } else {
+        this.emit(jsdoc.toString([SUPPRESS_DUPLICATE_TAG]));
```
evmar: This is weird, why are we emitting jsdoc if the flag is false?

I don't quite follow the logic above, but I think maybe we should drop the writeJSDoc flag and instead have the code above clear the jsdocTags array (?)

src/tsickle.ts
tbosch: Updated the docs.

I would keep the `checktypes` suppression where it is for now, if that is ok with you. Emits less code.

src/tsickle.ts
tbosch: Looked into it more, and when there is a constructor, we emit the jsdoc via `emitFunctionType` in this function. So we should actually add the suppress docs to that call. Changed the code to do so.

PR #622, opened 2017-09-26
fix(decorators): do not quote propDecorators

robwormald:
fixes part of #577


evmar:
Should we remove quotes from the other decorators?

mprobst:
This has been subsumed by the recent rewrite of decorators, they are now always unquoted.

PR #623, opened 2017-09-27
Build and Test with Bazel

LucasSloan:
Use bazel as the main build system instead of gulp.  Now you can `bazel test ...`

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,2 @@
+test --test_output=errors
+#TODO(alexeagle): enable worker mode
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - restore_cache:
           key: tsickle-{{ checksum "yarn.lock" }}
-      - run: bazel run @yarn//:yarn
+      - run: yarn install # don't bazel's yarn, since yarn install calls bazel, and bazel can't reenter
       - run: bazel build ...
       - save_cache:
           key: tsickle-{{ checksum "yarn.lock" }}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -9,18 +9,19 @@
       "args": [
         "--colors",
         "--timeout=0",
-        "./built/test"
+        "./../../../test/"
       ],
-      "cwd": "${workspaceRoot}",
+      "cwd": "${workspaceRoot}/bazel-out/local-fastbuild/bin/test/golden_test.runfiles/io_angular_tsickle",
       "runtimeArgs": [
         "--nolazy"
       ],
       "env": {
-        "NODE_ENV": "development"
+        "NODE_ENV": "development",
+        "RUNFILES": ".."
       },
       "console": "internalConsole",
       "sourceMaps": true,
-      "outDir": "${workspaceRoot}/built"
+      "outDir": "${workspaceRoot}/bazel-out/local-fastbuild/bin"
     },
     {
       "name": "Run file based tests",
@@ -30,19 +31,19 @@
       "args": [
         "--colors",
         "--timeout=0",
-        "./built/test/golden_tsickle_test.js"
+        "./../../../test/golden_tsickle_test.js"
       ],
-      "cwd": "${workspaceRoot}",
+      "cwd": "${workspaceRoot}/bazel-out/local-fastbuild/bin/test/golden_test.runfiles/io_angular_tsickle",
       "runtimeArgs": [
         "--nolazy"
       ],
       "env": {
         "NODE_ENV": "development",
-        "TEST_FILTER": "^functions/two_jsdoc_blocks"
+        "RUNFILES": ".."
       },
       "console": "internalConsole",
       "sourceMaps": true,
-      "outDir": "${workspaceRoot}/built"
+      "outDir": "${workspaceRoot}/bazel-out/local-fastbuild/bin"
     }
 
   ]
--- a/BUILD
+++ b/BUILD
@@ -3,3 +3,5 @@ exports_files(["tsconfig.json"])
 
 # NOTE: this will move to node_modules/BUILD in a later release
 filegroup(name = "node_modules", srcs = glob(["node_modules/**/*"]))
+
+filegroup(name = "test_files", srcs = glob(["test_files/**/*"]))
--- a/README.md
+++ b/README.md
@@ -102,23 +102,24 @@ Example:
 
 ## Development
 
-### Gulp tasks
+### Test commands
 
-- `gulp watch` executes the unit tests in watch mode (use `gulp test.unit` for a
+- `ibazel test test:unit_test` executes the unit tests in watch mode (use `bazel test test:unit_test` for a
   single run),
-- `gulp test.e2e` executes the e2e tests,
-- `guld test.golden` executes the golden tests,
+- `bazel test test:e2e_test` executes the e2e tests,
+- `bazel test test:golden_test` executes the golden tests,
 - `gulp test.check-format` checks the source code formatting using
   `clang-format`,
-- `gulp test` runs unit tests, e2e tests and checks the source code formatting.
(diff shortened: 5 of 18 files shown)


WORKSPACE
```diff
@@ -1,9 +1,11 @@
+workspace(name="io_angular_tsickle")
+
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 git_repository(
     name = "build_bazel_rules_nodejs",
     remote = "https://github.com/bazelbuild/rules_nodejs.git",
-    tag = "0.1.0",
+    commit = "71a5613",
```
alexeagle: I just tagged 0.1.7 which has the latest commits

install.sh
```diff
@@ -0,0 +1,7 @@
+mkdir tmp
```
alexeagle: filename should indicate this is just for use on travis

also it needs some explanatory comment inside the file

package.json
```diff
@@ -2,9 +2,9 @@
   "name": "tsickle",
   "version": "0.24.1",
   "description": "Transpile TypeScript code to JavaScript with Closure annotations.",
-  "main": "built/src/tsickle.js",
-  "typings": "built/definitions/tsickle.d.ts",
-  "bin": "built/src/main.js",
+  "main": "bazel-bin/src/tsickle.js",
```
alexeagle: I'm slightly concerned that the symlinks might do the wrong thing here

you can use
npm pack
 as a dry-run for npm publish to see that it will upload a working package

test/BUILD
```diff
@@ -1,10 +1,73 @@
...
+    name = "test_support",
+    srcs = ["test_support.ts"],
     deps = ["//src"],
     tsconfig = "//:tsconfig.json",
-    testonly = 1,
 )
+
+ts_library(
+    name = "unit_test_lib",
+    srcs = [
```
alexeagle: any thoughts yet about re-using this build file in google3 after this PR is synced? hopefully this would be the only place these files are listed

package.json
LucasSloan: Done.

test/BUILD
LucasSloan: Not yet, but that should be possible.  I've recently spent some time improving the tsickle sync proces.

alexeagle:
You can't re-enter Bazel, the server only does one request at a time.
bazel run yarn -> yarn postinstall -> bazel build ...
Doesn't work...

.bazelrc
```diff
@@ -0,0 +1 @@
+test --test_output=errors
```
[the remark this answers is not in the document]
alexeagle: maybe add a TODO then?

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
       - checkout
       - restore_cache:
           key: tsickle-{{ checksum "yarn.lock" }}
-      - run: bazel run @yarn//:yarn
+      - run: yarn install
```
alexeagle: add a comment why we don't use the yarn installed by bazel
(in general, any time you have to fix/workaround something, leave a comment explaining)

README.md
```diff
@@ -102,15 +102,15 @@ Example:
...
+- `ibazel test test:unit_test` executes the unit tests in watch mode (use `bazel test test:unit_test` for a
   single run),
-- `gulp test.e2e` executes the e2e tests,
-- `guld test.golden` executes the golden tests,
+- `bazel test test:e2e_test` executes the e2e tests,
+- `bazel test test:golden_test` executes the golden tests,
 - `gulp test.check-format` checks the source code formatting using
   `clang-format`,
-- `gulp test` runs unit tests, e2e tests and checks the source code formatting.
+- `yarn test` runs unit tests, e2e tests and checks the source code formatting.
```
alexeagle: does it work with the environment variables below?

travis_install.sh
```diff
@@ -0,0 +1,8 @@
+# installs bazel so travis can build and run tests
+mkdir tmp
+cd tmp
+curl --location --compressed https://github.com/bazelbuild/bazel/releases/download/0.5.4/bazel-0.5.4-installer-linux-x86_64.sh > bazel-0.5.2-installer-linux-x86_64.sh
```
alexeagle: there's a mix of 0.5.4 and 0.5.2 strings in here
should probably be extracted as a variable

alexeagle:
Thanks for taking this on!

.vscode/launch.json
```diff
@@ -9,18 +9,19 @@
       "args": [
         "--colors",
         "--timeout=0",
-        "./built/test"
+        "./../../../test/"
       ],
-      "cwd": "${workspaceRoot}",
+      "cwd": "${workspaceRoot}/bazel-out/local-fastbuild/bin/test/golden_test.runfiles/io_angular_tsickle",
```
evmar: Isn't this the same as bazel-bin/ without some intervening dirs?

.vscode/launch.json
```diff
@@ -9,18 +9,19 @@
...
       ],
       "env": {
-        "NODE_ENV": "development"
+        "NODE_ENV": "development",
+        "RUNFILES": ".."
       },
       "console": "internalConsole",
       "sourceMaps": true,
-      "outDir": "${workspaceRoot}/built"
+      "outDir": "${workspaceRoot}/bazel-out/local-fastbuild/bin"
```
evmar: here too

.vscode/launch.json
```diff
@@ -30,19 +31,19 @@
       "args": [
         "--colors",
         "--timeout=0",
-        "./built/test/golden_tsickle_test.js"
+        "./../../../test/golden_tsickle_test.js"
       ],
-      "cwd": "${workspaceRoot}",
+      "cwd": "${workspaceRoot}/bazel-out/local-fastbuild/bin/test/golden_test.runfiles/io_angular_tsickle",
```
evmar: here too

.vscode/launch.json
```diff
@@ -30,19 +31,19 @@
...
       ],
       "env": {
         "NODE_ENV": "development",
-        "TEST_FILTER": "^functions/two_jsdoc_blocks"
+        "RUNFILES": ".."
       },
       "console": "internalConsole",
       "sourceMaps": true,
-      "outDir": "${workspaceRoot}/built"
+      "outDir": "${workspaceRoot}/bazel-out/local-fastbuild/bin"
```
evmar: and here

BUILD
```diff
@@ -3,3 +3,5 @@ exports_files(["tsconfig.json"])
 
 # NOTE: this will move to node_modules/BUILD in a later release
 filegroup(name = "node_modules", srcs = glob(["node_modules/**/*"]))
+
+filegroup(name = "test_files", srcs = glob(["test_files/**/*"]))
```
evmar: trailing newline

gulpfile.js
```diff
@@ -6,24 +6,14 @@
...
-var ts = require('gulp-typescript');
 var tslint = require('gulp-tslint');
-var typescript = require('typescript');
 
-var tsProject = ts.createProject('tsconfig.json', {
-  // Specify the TypeScript version we're using.
-  typescript: typescript,
-});
+var onError = function(err) {
+  process.exit(1);
```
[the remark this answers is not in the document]
LucasSloan: This is a slightly cleaned up version of a function that did this - I don't think it's a good idea to change behavior.

.vscode/launch.json
LucasSloan: I'm using this path because this lines up better with the source mappings we produce.

BUILD
LucasSloan: Done

README.md
LucasSloan: Now it does.

travis_install.sh
LucasSloan: OK.

.bazelrc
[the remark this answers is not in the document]
LucasSloan: done.

PR #624, opened 2017-10-03
Add clutz aliases to .d.ts files.

LucasSloan:
Add a flag "addDtsClutzAliases" to control whether or not to add aliases to .d.ts files to add all the exports into the ಠ_ಠ.clutz namespace.


src/cli_support.ts
```diff
@@ -10,7 +10,7 @@ import * as path from 'path';
 
 // Postprocess generated JS.
 export function pathToModuleName(context: string, fileName: string): string {
-  fileName = fileName.replace(/\.js$/, '');
+  fileName = fileName.replace(/\.[^/.]+$/, '');
```
rkirov: why was this change needed?

src/tsickle.ts
```diff
@@ -1978,6 +1983,9 @@ export function emitWithTsickle(
         } else {
           content = combineSourceMaps(program, fileName, content);
         }
+        if (isDtsFileName(fileName) && host.addDtsClutzAliases && sourceFiles) {
```
rkirov: Put the experimental feature flag first - `if(host.addDtsClutzAliases && ...)`

src/tsickle.ts
```diff
@@ -2017,6 +2025,41 @@ export function emitWithTsickle(
...
+function addClutzAliases(
+    dtsFileContent: string, sourceFiles: ts.SourceFile[], typeChecker: ts.TypeChecker,
+    host: TsickleHost): string {
+  let reexports = '';
+  for (const sf of sourceFiles!) {
+    const moduleSymbol = typeChecker.getSymbolAtLocation(sf);
+    const moduleExports = moduleSymbol && moduleSymbol.exports || new Map<string, ts.Symbol>();
+
+    for (const e of moduleExports.entries()) {
+      const symbolName = e[0];
```
rkirov: just `for (const {symbolName, symbol} of moduleExports.entries())` should work

src/tsickle.ts
```diff
@@ -2017,6 +2025,41 @@ export function emitWithTsickle(
   };
 }
 
+function addClutzAliases(
+    dtsFileContent: string, sourceFiles: ts.SourceFile[], typeChecker: ts.TypeChecker,
+    host: TsickleHost): string {
+  let reexports = '';
+  for (const sf of sourceFiles!) {
```
rkirov: Why do you need ! in `sourceFiles!`. It look non-nullable to me `ts.SourceFile[]`

src/tsickle.ts
```diff
@@ -2017,6 +2025,41 @@ export function emitWithTsickle(
...
+    const moduleSymbol = typeChecker.getSymbolAtLocation(sf);
+    const moduleExports = moduleSymbol && moduleSymbol.exports || new Map<string, ts.Symbol>();
```
rkirov: why not just an early exit if `!moduleSymbol || !moduleSymbol.exports`. A bit odd to create empty map just to iterate of it.

src/tsickle.ts
```diff
@@ -2017,6 +2025,41 @@ export function emitWithTsickle(
...
+
+    for (const e of moduleExports.entries()) {
+      const symbolName = e[0];
+      const symbol = e[1];
+      const clutzSymbolName = 'module$contents$' +
```
rkirov: add a comment that this renaming scheme has to match internal JS compiler. Maybe even point to clutz unit tests.

src/tsickle.ts
```diff
@@ -2017,6 +2025,41 @@ export function emitWithTsickle(
...
+          host.pathToModuleName('', sf.fileName).replace(/\./g, '$') + '_' + e[0];
+      const symbolType = typeChecker.getTypeAtLocation(symbol.valueDeclaration!);
+      if ((symbolType as ts.ObjectType).objectFlags &&
```
rkirov: can we have a safe way to handle non-object types, instead of assertions. I.e. just bail if non-object type. I think you can get those cases if one writes CommonJs style typescript, instead of ES6
 
```
export = 0;
```

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,69 @@
...
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `const x = 1;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {files} = compileWithTransfromer(sources, dtsCompilerOptions);
+    const dts = files.get('input.d.ts')!;
+
+    expect(dts).not.to.contain('namespace ಠ_ಠ.clutz');
+  });
+
```
rkirov: add test for CJS style export - 'let x = 0; export = x;`

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,69 @@
...
+    sources.set('path/to/file/input.ts', `
+      export const x = 1;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
+    const {files} = compileWithTransfromer(sources, dtsCompilerOptions);
+    const dts = files.get('path/to/file/input.d.ts')!;
+
+    expect(dts).to.contain('const module$contents$path$to$file$input_x = x;');
+  });
+});
```
rkirov: can we have a test that the output is syntactically valid .d.ts. I am afraid we will miss a closing brace somewhere.

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,69 @@
...
+    expect(dts).to.contain('type module$contents$input_foo = foo;');
+    expect(dts).to.contain('const module$contents$input_foo: typeof foo;');
+  });
+
+  it('handles multisegment paths', () => {
+    const sources = new Map<string, string>();
+    sources.set('path/to/file/input.ts', `
+      export const x = 1;`);
+
+    // Run tsickle+TSC to convert inputs to Closure JS files.
```
rkirov: not an useful comment

src/cli_support.ts
LucasSloan: I started asking for the module name given the path to a .ts file, not just the .js files that were handled before.

test/e2e_clutz_dts_test.ts
LucasSloan: Done.

test/e2e_clutz_dts_test.ts
LucasSloan: Done.

src/tsickle.ts
```diff
@@ -2017,6 +2025,51 @@ export function emitWithTsickle(
   };
 }
 
+function addClutzAliases(
+    dtsFileContent: string, sourceFiles: ts.SourceFile[], typeChecker: ts.TypeChecker,
+    host: TsickleHost): string {
+  let reexports = '';
+  for (const sf of sourceFiles) {
+    const moduleSymbol = typeChecker.getSymbolAtLocation(sf);
+    // const moduleExports = moduleSymbol && moduleSymbol.exports;
```
rkirov: remove

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+// tslint:disable:no-unused-expression mocha .to.be.empty getters.
```
rkirov: cute getter apis getting in the way of static analysis, heh :/

No action required.

src/tsickle.ts
```diff
@@ -2017,6 +2025,51 @@ export function emitWithTsickle(
...
+    if (!moduleExports) {
+      return dtsFileContent;
+    }
+
+    // moduleExports is a ts.Map<ts.Symbol> which is an es6 Map, but has a
+    // different type for no reason
+    for (const symbol of moduleExports) {
+      // Want to alias the symbol to match what clutz would produce
+      // see examples at
+      // https://github.com/angular/clutz/tree/master/src/test/java/com/google/javascript/clutz
```
mprobst: Is this load bearing, or just something we do to mimick Clutz? Can you comment on that?

src/tsickle.ts
```diff
@@ -2017,6 +2025,51 @@ export function emitWithTsickle(
...
+      const clutzSymbolName = 'module$contents$' +
+          host.pathToModuleName('', sf.fileName).replace(/\./g, '$') + '_' + symbol.name;
+
+      const symbolType = typeChecker.getTypeAtLocation(symbol.valueDeclaration!);
+      console.log(symbolType);
+      if (symbolType.flags & ts.TypeFlags.Object &&
+          (symbolType as ts.ObjectType).objectFlags & ts.ObjectFlags.Class) {
```
mprobst: iirc there's a way to ask TypeScript whether a symbol is a value or only a type. Check the code base – we do make that decision somewhere else already, I think related to dereferencing `*` exports.

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,89 @@
...
+    sources.set('input.ts', `
+      export const x = 1;
+      export function incr(n) { return x + 1; }`);
+
+    const {files} = compileWithTransfromer(sources, dtsCompilerOptions);
+    const dts = files.get('input.d.ts')!;
+
+    expect(dts).to.contain('namespace ಠ_ಠ.clutz');
+    expect(dts).to.contain('const module$contents$input_x: typeof x;');
+    expect(dts).to.contain('const module$contents$input_incr: typeof incr;');
```
mprobst: shouldn't tsickle emit something that can be directly imported in JS land, i.e. the whole external ambient module declaration with a `goog:` prefix? That seems a lot less brittle than relying on some Closure Compiler name mangling scheme to transport the type over.

test/test_support.ts
```diff
@@ -43,15 +43,20 @@ export const compilerOptions: ts.CompilerOptions = {
...
+};
+
 /**
  * Basic compiler options for source map tests. Compose with
  * generateOutfileCompilerOptions() or inlineSourceMapCompilerOptions to
  * customize the options.
  */
 export const sourceMapCompilerOptions: ts.CompilerOptions = {
-  ...baseCompilerOptions,
+  ...dtsCompilerOptions,
```
mprobst: do these options require the dts emit?

test/test_support.ts
```diff
@@ -43,15 +43,20 @@ export const compilerOptions: ts.CompilerOptions = {
   outDir: '.',
 };
 
+/** Compiler options to use if you want to produce .d.ts outputs. */
+export const dtsCompilerOptions: ts.CompilerOptions = {
```
mprobst: Could you instead keep these options local to where they are used, i.e. in the dts test file?

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  it('exports the type and value of a class', () => {
+    const sources = new Map<string, string>();
+    sources.set('input.ts', `export class foo {}`);
```
mprobst: upper case `Foo`

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,89 @@
...
+    sources.set('path/to/file/input.ts', `let x = 0; export = x;`);
+
+    const {files} = compileWithTransfromer(sources, dtsCompilerOptions);
+    const dts = files.get('path/to/file/input.d.ts')!;
+
+    const program = createProgram(new Map([['input.d.ts', dts]]));
+    const diagnostics = ts.getPreEmitDiagnostics(program);
+
+    expect(diagnostics, tsickle.formatDiagnostics(diagnostics)).to.be.empty;
+  });
```
mprobst: You only have tests exporting values. Can you add tests that export types, e.g. `export interface ...`, `export type ...`?

test/e2e_clutz_dts_test.ts
```diff
@@ -0,0 +1,89 @@
...
+// tslint:disable:no-unused-expression mocha .to.be.empty getters.
+
+import {expect} from 'chai';
+import * as ts from 'typescript';
+
+import * as tsickle from '../src/tsickle';
+
+import {compileWithTransfromer, createProgram, dtsCompilerOptions} from './test_support';
+
+describe('clutz dts', () => {
```
mprobst: can you add a test that ensures the `.d.ts` files we emit are valid TS, and ideally one that shows they can be used?

Previously we've usually created a new test folder suffix (`.untyped`) to test variants of the emit, and then use golden file based tests. Maybe that would be a better approach here, too? It makes it easier to make an e2e with something using the `.d.ts` emit, and it keeps our tests consistently in one style. You'd need to change the test runner slight to support the pattern, but it might end up being not much more work than this here.

I'd also prefer if we had fewer one off test suites, and more code would use the golden file tests.

LucasSloan: I've moved all the tests that assert that strings appear in the .d.ts to be golden tests with the`.declaration` suffix.  I've also changed e2e_clutz_dts_test.ts to act similar to the previous `produces a valid .d.ts` test case, by checking for preemit errors after making a program with all the `.declaration` goldens.

test/e2e_clutz_dts_test.ts
LucasSloan: Done.

src/tsickle.ts
LucasSloan: Yep, made this code much more elegant.

mprobst:
> JS tooling doesn't consume .d.ts's, typescript does. This is about producing a .d.ts that will fit in with the .d.ts we produce from clutz to be consumed by tsc.

Yes. My concern is that we're emitting code here in tsickle that depends on an implementation detail of JSCompiler (not even of Clutz!). Can we do better than depending on the exact name mangling pattern? Could e.g. Clutz have a documented format it expects, and emit an alias for the types it knows will be missing?

If that's too hard, we might have to deal with it, but I presume you agree this is a bit suboptimal.

rkirov:
Opened https://github.com/angular/clutz/issues/575 to discuss the goog.module name mangling in clutz.

PR #625, opened 2017-10-06
Enable writing new goldens.

LucasSloan:
Previously, updating goldens worked because the runfiles directory contains symlinks back to the original file.  However, if we needed to create a new golden file, there wasn't a symlink, so we'd create the new golden in the bazel runfiles directory, not in the test_files directory.  By parsing the symlink to the source ts files for the test, we can find the absolute path to the tsickle workspace root, which allows us to construct the path to the actual file we want to update.

This commit also changes the semantics of requesting a goldens update, by only requiring you to bazel run test:golden_test, instead of having to run the target as well as setting an environment variable.

### DIFF
--- a/test/golden_tsickle_test.ts
+++ b/test/golden_tsickle_test.ts
@@ -29,7 +29,7 @@ const TEST_FILTER = (() => {
 
 // If true, update all the golden .js files to be whatever tsickle
 // produces from the .ts source. Do not change this code but run as:
-//     UPDATE_GOLDENS=y gulp test
+//     UPDATE_GOLDENS=y bazel run test:golden_test
 const UPDATE_GOLDENS = !!process.env.UPDATE_GOLDENS;
 
 function readGolden(path: string): string|null {
@@ -56,10 +56,11 @@ function readGolden(path: string): string|null {
  *    externs files, where the majority of tests are not expected to
  *    produce one.)
  */
-function compareAgainstGolden(output: string|null, path: string) {
+function compareAgainstGolden(
+    output: string|null, goldenPath: string, test: testSupport.GoldenFileTest) {
   let golden: string|null = null;
   try {
-    golden = fs.readFileSync(path, 'utf-8');
+    golden = fs.readFileSync(goldenPath, 'utf-8');
   } catch (e) {
     if (e.code === 'ENOENT' && (UPDATE_GOLDENS || output === null)) {
       // A missing file is acceptable if we're updating goldens or
@@ -74,20 +75,24 @@ function compareAgainstGolden(output: string|null, path: string) {
   if (output != null) output = normalizeLineEndings(output);
 
   if (UPDATE_GOLDENS && output !== golden) {
-    console.log('Updating golden file for', path);
+    console.log('Updating golden file for', goldenPath);
+    // If we need to write a new file, we won't have a symlink into the real
+    // test_files directory, so we need to get an absolute path by combining
+    // the relative path with the workspaceRoot
+    const goldenSourcePath = path.join(test.getWorkspaceRoot(), goldenPath);
     if (output !== null) {
-      fs.writeFileSync(path, output, {encoding: 'utf-8'});
+      fs.writeFileSync(goldenSourcePath, output, {encoding: 'utf-8'});
     } else {
       // The desired golden state is for there to be no output file.
       // Ensure no file exists.
       try {
-        fs.unlinkSync(path);
+        fs.unlinkSync(goldenSourcePath);
       } catch (e) {
         // ignore.
       }
     }
   } else {
-    expect(output).to.equal(golden, `${path}`);
+    expect(output).to.equal(golden, `${goldenPath}`);
   }
 }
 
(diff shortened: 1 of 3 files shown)


test/golden_tsickle_test.ts
```diff
@@ -28,9 +28,12 @@ const TEST_FILTER = (() => {
...
 // If true, update all the golden .js files to be whatever tsickle
-// produces from the .ts source. Do not change this code but run as:
-//     UPDATE_GOLDENS=y gulp test
-const UPDATE_GOLDENS = !!process.env.UPDATE_GOLDENS;
+// produces from the .ts source. Do not change this code but bazel run
+// the golden target instead of bazel test'ing it.
+// Detect that we are not running as a test, by using a well-known environment
+// variable. See https://docs.bazel.build/versions/master/test-encyclopedia.html
+
+const UPDATE_GOLDENS = !process.env.TEST_TMPDIR;
```
mprobst: I find this a bit spooky. Could we just have an explicit command line argument or environment variable passed in that controls this?

It's generally better to explicitly pass configuration into software than having software try to be clever and detect run modes from its local environment or other factors. In particular for something like this, where we could end up destroying data or end up with a test that never fails, I'd much rather be explicit.

test/test_support.ts
```diff
@@ -173,6 +173,19 @@ export class GoldenFileTest {
         .map(f => path.join(this.path, GoldenFileTest.tsPathToJs(f)));
   }
 
+  /**
+   * Find the absolute path to the tsickle root directory by reading the
+   * symlink bazel puts into bazel-bin back into the test_files directory
+   * and chopping off the test_files portion.
+   */
+  get workspaceRoot(): string {
```
mprobst: given that this has substantial logic, can fail, and also some performance impact (reads the file system), can you make it a regular method, to reflect that?

test/test_support.ts
```diff
@@ -173,6 +173,19 @@ export class GoldenFileTest {
...
+    if (!this.tsFiles) {
```
mprobst: do not silently swallow errors, fail loudly.

test/test_support.ts
```diff
@@ -173,6 +173,19 @@ export class GoldenFileTest {
...
+      return '';
+    }
+    const resolvedFileSymLink = fs.readlinkSync(path.join(this.path, this.tsFiles[0]));
+    return resolvedFileSymLink.replace(/\/test_files.*$/, '');
```
mprobst: this won't work on windows, and I think we still have some people interested in that.

Can you use path.separator here and split based on it, then drop segments after the first `test_file`? Maybe there's also a more elegant formulation.

test/test_support.ts
```diff
@@ -173,6 +173,19 @@ export class GoldenFileTest {
...
+  /**
+   * Find the absolute path to the tsickle root directory by reading the
+   * symlink bazel puts into bazel-bin back into the test_files directory
+   * and chopping off the test_files portion.
+   */
+  get workspaceRoot(): string {
+    if (!this.tsFiles) {
+      return '';
+    }
+    const resolvedFileSymLink = fs.readlinkSync(path.join(this.path, this.tsFiles[0]));
```
mprobst: this will fail if tsFiles is set but empty.

test/test_support.ts
LucasSloan: Done.

test/golden_tsickle_test.ts
LucasSloan: You already have to `bazel run` the target to get the goldens to update at all, because otherwise you're sandboxed and can't update the goldens.  If this magic fails you'll just get an error that you couldn't write the file, not a silently passing test.  We use this trick in g3 for our tsconfig golden tests, so I think we're on solid ground here.

LucasSloan:
PTAL

test/golden_tsickle_test.ts
mprobst: I don't like the pattern in google3 either. `bazel run` vs `bazel test` having the side effect of updating arbitrary test inputs is way too magic. An explicit flag here doesn't cost us anything and makes sure we and the user know what we're doing.

test/test_support.ts
```diff
@@ -178,12 +178,15 @@ export class GoldenFileTest {
    * symlink bazel puts into bazel-bin back into the test_files directory
    * and chopping off the test_files portion.
    */
-  get workspaceRoot(): string {
-    if (!this.tsFiles) {
-      return '';
+  getWorkspaceRoot(): string {
+    if (!this.tsFiles || !this.tsFiles[0]) {
```
mprobst: what about `!this.tsFiles.length` for the second condition? Avoids indexing into an array out of bounds.

LucasSloan: Done.

test/golden_tsickle_test.ts
LucasSloan: Reverted, then.

LucasSloan:
PTAL

test/golden_tsickle_test.ts
```diff
@@ -28,8 +28,8 @@ const TEST_FILTER = (() => {
 })();
 
 // If true, update all the golden .js files to be whatever tsickle
-// produces from the .ts source. Do not change this code but run as:
-//     UPDATE_GOLDENS=y gulp test
+// produces from the .ts source. Do not change this code but export
+// UPDATE_GOLDENS=1 before bazel running the target
```
mprobst: can you run as `UPDATE_GOLDEN=1 bazel run ...`? Please document the full command, it's nice if you can just copy&paste it.

LucasSloan: Done.

PR #627, opened 2017-10-14
Upgrade tsickle to TS2.5

rkirov:
The major change is that NodeArray is now readonly, but we still
want to modify it in a number of places.

### DIFF
--- a/src/class_decorator_downlevel_transformer.ts
+++ b/src/class_decorator_downlevel_transformer.ts
@@ -96,13 +96,18 @@ function isNameEqual(classMember: ts.ClassElement, name: string): boolean {
  * transformers
  */
 function insertBeforeDecoratorProperties(
-    classMembers: ts.NodeArray<ts.ClassElement>, decoratorMetadata: ts.PropertyDeclaration) {
+    classMembers: ts.NodeArray<ts.ClassElement>,
+    decoratorMetadata: ts.PropertyDeclaration): ts.NodeArray<ts.ClassElement> {
   let insertionPoint = classMembers.findIndex(
       m => isNameEqual(m, 'ctorParameters') || isNameEqual(m, 'propDecorators'));
   if (insertionPoint === -1) {
     insertionPoint = classMembers.length - 1;
   }
-  classMembers.splice(insertionPoint, 0, decoratorMetadata);
+  const members = [
+    ...classMembers.slice(0, insertionPoint), decoratorMetadata,
+    ...classMembers.slice(insertionPoint)
+  ];
+  return ts.setTextRange(ts.createNodeArray(members, classMembers.hasTrailingComma), classMembers);
 }
 
 export function classDecoratorDownlevelTransformer(
@@ -128,7 +133,7 @@ export function classDecoratorDownlevelTransformer(
 
           const newClassDeclaration = ts.getMutableClone(cd);
 
-          insertBeforeDecoratorProperties(
+          newClassDeclaration.members = insertBeforeDecoratorProperties(
               newClassDeclaration.members, createDecoratorClassProperty(decoratorList));
 
           newClassDeclaration.decorators = undefined;
--- a/src/transformer_util.ts
+++ b/src/transformer_util.ts
@@ -288,7 +288,8 @@ export function visitNodeWithSynthesizedComments<T extends ts.Node>(
  * @param node
  */
 function resetNodeTextRangeToPreventDuplicateComments<T extends ts.Node>(node: T): T {
-  ts.setEmitFlags(node, (ts.getEmitFlags(node) || 0) | ts.EmitFlags.NoComments);
+  // tslint:disable-next-line:no-any TS 2.5 typescript.d.ts does not expose getEmitFlags.
+  ts.setEmitFlags(node, ((ts as any).getEmitFlags(node) || 0) | ts.EmitFlags.NoComments);
   // See also addSyntheticCommentsAfterTsTransformer.
   // Note: Don't reset the textRange for ts.ExportDeclaration / ts.ImportDeclaration
   // until after the TypeScript transformer as we need the source location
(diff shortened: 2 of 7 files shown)


rkirov:
Not planning to merge this yet, but up for review to see if someone has better ways of working through the API changes. I mostly just added `any`s due to lack of understanding of the code.

mprobst:
Regarding the NodeArray, they must have a pattern they use for their own decorators that still allows manipulating node arrays, right? Or maybe we should re-arrange the code so that we only piece the whole array together at the very end?

/CC @tbosch

tbosch:
Agree, we should just create a new `NodeArray` at the end.

tbosch:
FYI, when using tsickle with TS 2.5 (without the changes in this PR), users get broken compilations. See e.g. https://github.com/angular/angular/issues/19587.

tbosch:
Note that for Angular, as we already committed to supporting TS 2.4, we would need a version of tsickle that works in TS 2.4 and TS 2.5.

rkirov:
Updated the PR with '.slice()' to get a mutable copy of NodeArrays. I didn't see any other convenience APIs for creating new readonly arrays with added/deleted members.

Talked with Chuck about supporting 2.5 and 2.4 through shim. There will be a separate PR for that. When I land this PR, I will create a branch for 2.4 at the current state that we can use to backport patches if needed.

src/transformer_util.ts
```diff
@@ -392,18 +391,19 @@ function visitNodeStatementsWithSynthesizedComments<T extends ts.Node>(
   const leading = synthesizeDetachedLeadingComments(sourceFile, node, statements);
   const trailing = synthesizeDetachedTrailingComments(sourceFile, node, statements);
   if (leading.commentStmt || trailing.commentStmt) {
-    statements = ts.setTextRange(ts.createNodeArray(statements), {pos: -1, end: -1});
```
chuckjaz: How about something like:
```ts
    const newStatements = [...arrayOf(leading.commentStmt), ...statements, ...arrayOf(trailing.commendStmt)];
```
where
```ts
function arrayOf<T>(value: T | undefined): T[] {
  return value ? [value] : [];
}
```

rkirov: done

rkirov:
I got the tests to pass with only one questionable assignment.

src/class_decorator_downlevel_transformer.ts
```diff
@@ -96,13 +96,25 @@ function isNameEqual(classMember: ts.ClassElement, name: string): boolean {
...
+    decoratorMetadata: ts.PropertyDeclaration): ts.NodeArray<ts.ClassElement> {
   let insertionPoint = classMembers.findIndex(
       m => isNameEqual(m, 'ctorParameters') || isNameEqual(m, 'propDecorators'));
   if (insertionPoint === -1) {
     insertionPoint = classMembers.length - 1;
   }
-  classMembers.splice(insertionPoint, 0, decoratorMetadata);
+  const members: ts.ClassElement[] = [];
+
+  for (let i = 0; i < classMembers.length;) {
```
mprobst: what about `members = [classMembers.slice(0, insertionPoint)..., decoratorMetadata, classMembers.slice(insertionPoint)]`?

src/transformer_util.ts
```diff
@@ -288,7 +288,8 @@ export function visitNodeWithSynthesizedComments<T extends ts.Node>(
  * @param node
  */
 function resetNodeTextRangeToPreventDuplicateComments<T extends ts.Node>(node: T): T {
-  ts.setEmitFlags(node, (ts.getEmitFlags(node) || 0) | ts.EmitFlags.NoComments);
+  // tslint:disable-next-line:no-any TS 2.5 typescript.d.ts does not expose getEmitNodes.
```
mprobst: typo getEmitFlags

src/transformer_util.ts
```diff
@@ -301,11 +302,9 @@ function resetNodeTextRangeToPreventDuplicateComments<T extends ts.Node>(node: T
   if (node.kind === ts.SyntaxKind.PropertyDeclaration) {
     allowTextRange = false;
     const pd = node as ts.Node as ts.PropertyDeclaration;
-    // TODO(tbosch): Using pd.initializer! as the typescript typings before 2.4.0
-    // are incorrect. Remove this once we upgrade to TypeScript 2.4.0.
     node = ts.updateProperty(
-               pd, pd.decorators, pd.modifiers, resetTextRange(pd.name) as ts.PropertyName, pd.type,
-               pd.initializer!) as ts.Node as T;
+               pd, pd.decorators, pd.modifiers, resetTextRange(pd.name) as ts.PropertyName,
+               pd.questionToken, pd.type, pd.initializer) as ts.Node as T;
```
mprobst: does the 2.4 updateProperty accept the questionToken here?

src/transformer_util.ts
```diff
@@ -392,13 +395,17 @@ function visitNodeStatementsWithSynthesizedComments<T extends ts.Node>(
...
-    }
-    if (trailing.commentStmt) {
-      statements.push(trailing.commentStmt);
-    }
+    const newStatements: ts.Statement[] =
+        [...arrayOf(leading.commentStmt), ...statements, ...arrayOf(trailing.commentStmt)];
+    statements = ts.setTextRange(ts.createNodeArray(newStatements), {pos: -1, end: -1});
+
+    // The visitor creates a new node with the new statements. However, doing so
+    // reveals a TypeScript bug.
```
mprobst: reference the bug/issue, or at least explain what the bug behaviour is here?

src/transformer_util.ts
```diff
@@ -392,13 +395,17 @@ function visitNodeStatementsWithSynthesizedComments<T extends ts.Node>(
...
+    // The line below is a workaround that ensures that updateSourceFileNode and
+    // updateBlock never create new Nodes.
+    // TODO: narrow this down and file a bug with TS team.
```
mprobst: TODO format plz

src/transformer_util.ts
```diff
@@ -604,6 +611,7 @@ export function visitEachChild(
  * This is a version of `ts.updateSourceFileNode` that works
  * well with property decorators.
  * See https://github.com/Microsoft/TypeScript/issues/17384
+ * TODO: This has been fixed in TS 2.5. Investigate removal.
```
mprobst: TODO format

src/transformer_util.ts
rkirov: This change is not meant to be backwards compatible with 2.4. I am confused why are you asking about 2.4?

src/class_decorator_downlevel_transformer.ts
rkirov: done

src/transformer_util.ts
rkirov: done

src/transformer_util.ts
rkirov: As discussed, this change does not have backwards compatibility with TS2.4. I will update the 2.4 branch to the current head before submitting this. Chuck will work on making tsickle 2.4 + 2.5 compatible separately.

src/transformer_util.ts
rkirov: I paired with @tbosch on this comment, already forgot what exactly was breaking (other than tests failing). Tobias, can you help me expanding on the comment?

src/transformer_util.ts
rkirov: done

src/transformer_util.ts
rkirov: done

src/transformer_util.ts
rkirov: Paired with Tobias more today. Still not ideal and somewhat mysterious but we narrowed it down to a small repro case.

PR #628, opened 2017-10-18
Include generic type parameters in .d.ts clutz aliases.

LucasSloan:
Adds support and a test of generic type parameters in .d.ts clutz aliases.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -2028,7 +2028,7 @@ export function emitWithTsickle(
 function addClutzAliases(
     fileName: string, dtsFileContent: string, sourceFiles: ts.SourceFile[],
     typeChecker: ts.TypeChecker, host: TsickleHost): string {
-  let reexports = '';
+  const reexports: string[] = [];
   for (const sf of sourceFiles) {
     const moduleSymbol = typeChecker.getSymbolAtLocation(sf);
     const moduleExports = moduleSymbol && typeChecker.getExportsOfModule(moduleSymbol);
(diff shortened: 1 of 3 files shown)


src/tsickle.ts
```diff
@@ -2079,6 +2084,56 @@ function addClutzAliases(
...
+function getGenericTypeParameters(symbol: ts.Symbol): {params: string, constrainedParams: string} {
+  if (!symbol.declarations) {
+    return {params: '', constrainedParams: ''};
+  }
+
+  if (!symbol.declarations[0]) {
+    return {params: '', constrainedParams: ''};
+  }
+
+  const declaration = symbol.declarations[0];
```
rkirov: add a comment about the assumption we talked about here 'all declarations have to have matching generic types'

src/tsickle.ts
```diff
@@ -2079,6 +2084,56 @@ function addClutzAliases(
...
+  if (!declarationWithTypeParameters.typeParameters) {
+    return {params: '', constrainedParams: ''};
+  }
+
+  const paramList: string[] = [];
+  const constrainedParamList: string[] = [];
+  for (const param of declarationWithTypeParameters.typeParameters) {
+    if (param.constraint) {
+      constrainedParamList.push(`${param.name.text} extends ${
+          (param.constraint as ts.TypeReferenceNode).typeName.getText()}`);
```
rkirov: This cast is not safe - consider `class C<T extends {a: string}>`. Technically the bound could be any type expression (not just reference). Why can't we just do ` param.constraint.getText()`.

src/tsickle.ts
```diff
@@ -2079,6 +2084,56 @@ function addClutzAliases(
...
+    } else {
+      constrainedParamList.push(param.name.text);
```
rkirov: .text vs .getText() is there a difference?

src/tsickle.ts
```diff
@@ -2050,18 +2050,23 @@ function addClutzAliases(
       // https://github.com/angular/clutz/tree/master/src/test/java/com/google/javascript/clutz
       const clutzSymbolName = 'module$contents$' + clutzModuleName + '_' + symbol.name;
 
+      const genericParams = getGenericTypeParameters(symbol);
+
       if (symbol.flags & ts.SymbolFlags.Class) {
         // classes need special care to match clutz, which seperates class types into a
         // type for the static properties and a type for the instance properties
-        reexports += `\t\ttype ${clutzSymbolName} = ${symbol.name};\n`;
+        reexports += `\t\ttype ${clutzSymbolName}${genericParams.constrainedParams} = ${
```
rkirov: This is getting gnarly, consider helper methods like emit, indent,unindent. `https://github.com/angular/clutz/blob/master/src/main/java/com/google/javascript/clutz/DeclarationGenerator.java#L1129`.

Could be done in the next PR.

src/tsickle.ts
```diff
@@ -2079,6 +2084,56 @@ function addClutzAliases(
   return dtsFileContent;
 }
 
+/**
+ * Returns 2 strings specifying the generic type arguments for the symbol.  The constrained params
+ * include any `T extends foo` arguments, the regular params are just a list of the type symbols,
+ * since we need the constraints on the LHS of the alias declaration, but can't have them on the
+ * RHS.
+ */
+function getGenericTypeParameters(symbol: ts.Symbol): {params: string, constrainedParams: string} {
```
rkirov: I think 'paramsWithContraint' is a better name. For awhile, I thought you split the params into two buckets, ones with and without contraints.

src/tsickle.ts
```diff
@@ -2079,6 +2084,56 @@ function addClutzAliases(
...
+    if (param.constraint) {
+      constrainedParamList.push(`${param.name.text} extends ${
+          (param.constraint as ts.TypeReferenceNode).typeName.getText()}`);
+    } else {
+      constrainedParamList.push(param.name.text);
+    }
+    paramList.push(param.name.getText());
+  }
+
+  const params = `<${paramList.join(',')}>`;
```
rkirov: nit, I think the < > and ',' belong in the emit code, and this helper should emit {param: string[], paramWithConstrain: string[]}

src/tsickle.ts
```diff
@@ -2050,18 +2050,23 @@ function addClutzAliases(
       // https://github.com/angular/clutz/tree/master/src/test/java/com/google/javascript/clutz
       const clutzSymbolName = 'module$contents$' + clutzModuleName + '_' + symbol.name;
 
+      const genericParams = getGenericTypeParameters(symbol);
```
rkirov: `const {params, constrainedParams}`

src/tsickle.ts
LucasSloan: Some things have both and some just have the `.getText()`.  I've standardized on `.getText()`.

src/tsickle.ts
LucasSloan: I had it that way originally - I don't like how that requires additional variables, or a ternary expression at every usage.

src/tsickle.ts
LucasSloan: Made a partial step in that direction.

src/tsickle.ts
LucasSloan: Done.

src/tsickle.ts
```diff
@@ -2050,35 +2050,95 @@ function addClutzAliases(
...
   }
 
   return dtsFileContent;
 }
 
+/**
+ * Returns 2 strings specifying the generic type arguments for the symbol.  The constrained params
+ * include any `T extends foo` arguments, the regular params are just a list of the type symbols,
+ * since we need the constraints on the LHS of the alias declaration, but can't have them on the
+ * RHS.
```
rkirov: add that we include default here too.

PR #640, opened 2017-11-02
document yarn setup workaround

evmar:
- Move setup instructions lower (most users will just be installing
  this from npm, so they don't need to know about bazel).
- Document current workaround for setup.

See https://github.com/bazelbuild/rules_nodejs/issues/46 .


alexeagle:
Give me a few minutes, I am hoping to just fix that bug rather than need to have the workaround...

alexeagle:
how about https://github.com/angular/tsickle/pull/642

alexeagle:
this commit merged in #642

PR #641, opened 2017-11-02
    handle tslib-style export * rewriting

evmar:
If code is built using tsc's importHelpers flag, "export *" expressions
    produce code like:
      var tslib_1 = require('tslib');
      tslib_1.__exportStar(require('foo'), exports);
    
    Adjust the code that rewrites "require" into "goog.require" to also
    match this specific pattern, by looking for use of the special name
    __exportStar().

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -206,11 +206,12 @@ class ES5Processor extends Rewriter {
       // It's possibly of the form:
       // - require(...);
       // - __export(require(...));
-      // Both are CallExpressions.
+      // - tslib_1.__exportStar(require(...));
+      // All are CallExpressions.
       const exprStmt = node as ts.ExpressionStatement;
       const expr = exprStmt.expression;
       if (expr.kind !== ts.SyntaxKind.CallExpression) return false;
-      const call = expr as ts.CallExpression;
+      let call = expr as ts.CallExpression;
 
       let require = this.isRequire(call);
       let isExport = false;
@@ -220,8 +221,11 @@ class ES5Processor extends Rewriter {
         //   __export(x);
         // This extra variable is necessary in case there's a later import of the
         // same module name.
-        require = this.isExportRequire(call);
-        isExport = require != null;
+        const innerCall = this.isExportRequire(call);
+        if (!innerCall) return false;
+        isExport = true;
+        call = innerCall;  // Update call to point at the require() expression.
+        require = this.isRequire(call);
       }
       if (!require) return false;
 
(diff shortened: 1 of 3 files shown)


src/es5processor.ts
```diff
@@ -220,16 +221,24 @@ class ES5Processor extends Rewriter {
...
 
       this.writeLeadingTrivia(node);
       const varName = this.emitGoogRequire(null, require);
 
       if (isExport) {
-        this.emit(`__export(${varName});`);
+        // node is a statement containing a require() in it, while
+        // requireCall is that call.  We replace the require() call
+        // with the variable we emitted.
+        let fullStatement = node.getText();
```
alexeagle: maybe we should lint tsickle like in google3, ought to be const

but that would be a separate task to change it throughout

test/BUILD
```diff
@@ -1,31 +1,33 @@
...
+
 load("@build_bazel_rules_typescript//:defs.bzl", "ts_library")
 load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test")
 
 ts_library(
     name = "test_support",
     srcs = ["test_support.ts"],
-    deps = ["//src"],
     tsconfig = "//:tsconfig.json",
+    deps = ["//src"],
```
alexeagle: curious, did you run a formatter here? I think buildifier is public but I never hooked it up anywhere

test/es5processor_test.ts
```diff
@@ -113,6 +113,12 @@ __export(require('req/mod'));`)
               `goog.module('a');var module = module || {id: 'a.js'};var mod = goog.require('req.mod');
 __export(mod);`);
     });
+
+    it('converts tslib exportStar usage', () => {
+      expectCommonJs('a.js', `tslib_1.__exportStar(require("./decorators"), exports);`)
```
alexeagle: what if tslib changes in the future? would be nice to have an end-to-end test also

src/es5processor.ts
evmar: Done.

test/BUILD
evmar: My emacs autoformats BUILD files, yeah.

test/es5processor_test.ts
evmar: That is a good observation.  There's a chicken and egg problem here where we need to get tslib working end-to-end before we can test it all, so I think I will save it for later.  Ultimately I think we should drop the code that handles the non-tslib codepath because using tslib is always better.  I considered doing that here but I think it will be hard to get everything converted at once.

Issue #490, opened 2017-05-08
tsickle produces wrong module references when using baseUrl and outDir simultaneously.


rubenlg:
Directory structure:
```
.
├── lib
│   └── module.ts
└── program
    ├── main.ts
    └── tsconfig.json
```

lib/module.ts:
```typescript
export function foo() {
  console.log('foo');
}
```

program/main.ts:
```typescript
import {foo} from 'lib/module';

function main() {
  foo();
}
```

program/tsconfig.json:
```json
{
  "compilerOptions": {
    "module": "commonjs",
    "baseUrl": "../",
    "outDir": "jsout"
  }
}
```

**Expected**: 
The resulting code can be compiled with closure compiler.

**Actual**:
```sh
jsout/program/main.js:3: ERROR - Required namespace "lib.module" never defined.
```
Here are the `module` names generated in the two files:

jsout/lib/module.js:
```javascript
goog.module('jsout.lib.module');
```

jsout/program/main.js:
```
goog.module('jsout.program.main'); // ...
var module_1 = goog.require('lib.module');
```

The name of the `module` module should be the same in both cases.

**NOTE**: The bug still reproduces if `lib` is under `program` and `baseUrl` is ".". I just thought that it was a more realistic case to have libraries outside of your program, perhaps shared with other programs and libraries.

rubenlg:
Note that the original bug listed above is now fixed, but it has reincarnated in a slightly different form, now it affects goog.forwardDeclare instead of goog.require.

rubenlg:
FYI: I created the following test case that reproduces the outstanding issue:
https://github.com/rubenlg/tsickle/commit/f57f03ba94cb7f286572359e9ff343d24b46b941

You can see what happens by running:
```sh
bazel test test:paths_test
```

(If you apply that commit as a patch on the original tsickle, you also need to drop an unused import)

I haven't looked very deeply into the tsickle codebase to figure out what's the right fix. My initial reaction would be to always generate module names from the absolute file paths, but that might break other things. Any guidance would be appreciated.

rubenlg:
I forked tsickle yesterday (April 2) evening, with the GitHub UI, which I assume forks from master.

mprobst:
Then I guess the problem can still happen in some call path, or the rootDir fix should be amended by considering baseUrl in some context.

mprobst:
For reference, the fix I mentioned was a2213f85a20a596e10bfc5066169eadda1c58096.

[...]

Issue #481, opened 2017-04-27
tsickle seemingly breaks modules


simonsarris:
I have an extremely simple TypeScript project, in total it is:

```TypeScript
// classB.ts:
export class ClassB {
  constructor() {  }

  toString(): string {
    return "this is a class B instance";
  }
}
```

```TypeScript
// index.ts:
import { ClassB } from "./classB";

export let myLib: any = {}

// very semantic version
myLib.version = "lalala";

// runtime API
myLib.ClassB = ClassB;
```

```TypeScript
// tsconfig.json:
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "outDir": "./tsout",
    "rootDir": "./src",
    "noEmitHelpers": true, // needed for tsickle
    "strict": true
  }
}
```

Running `tsickle` produces almost valid output, except `index.js` looks like this:

```JavaScript
goog.module('tsout.index'); exports = {}; var module = {id: 'tsout/index.js'};
Object.defineProperty(exports, "__esModule", { value: true });
var classB_1 = goog.require('tsout_classB');
exports.myLib = {};
// very semantic version
exports.myLib.version = "lalala";
// runtime API
exports.myLib.ClassB = classB_1.ClassB;
```

`goog.require('tsout_classB');` is causing the closure compiler to fail. Shouldn't tsickle produce `'tsout.classB'` instead? If I change that line by hand so that the underscore is a period, it works.

```
PS C:\VerySimpleTSProject> java -jar .\closure-compiler.jar --js_output_file=go.js './tsout/*.js' --jscomp_off=deprecatedAnnotations --compilation_level AD
VANCED_OPTIMIZATIONS --formatting=pretty_print
tsout/index.js:3: ERROR - Required namespace "tsout_classB" never defined.
var classB_1 = goog.require('tsout_classB');
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

What explains this error? Running tsc with the above, there are no similar `require` errors.

(Unrelated to the bug, I'm trying to write TS to produce a browser library called myLib. If I'm doing anything obviously wrong on that front please let me know).

evmar:
I think tsickle has a bug in the presence of either outDir or rootDir, not sure which.  I think we don't use rootDir in any of our TS code.  Your help in debugging this would be most helpful!  I would expect the module names should not mention 'tsout' in them, so maybe figuring out why that is would be the place to start.

simonsarris:
[...]

PR #645, opened 2017-11-07
Fix path handling in tsickle tests a bit

mprobst:
This is a prequisite to run the test suite in other environments (g3), but also just a general cleanup of messy code.

Additionally I think I discovered the cause for #481 and #490. Internally and when using bazel, we never run into this, because we emulate one consistent path hierarchy for tsc, i.e. we never use `rootDir` or `outDir` due to clever bazel file layout.

However if you do have `rootDir` or `outDir` set, then we're calling `pathToModuleName` with inconsistent paths, causing inconsistent `goog.require` and `goog.module` calls. This PR fixes that issue, though in a somewhat incomplete way. Ideally, we should not strip `rootDir` but rather call into some TypeScript API to obtain the logical file name. I don't know which or whether it is exposed though.

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -270,6 +270,7 @@ class ES5Processor extends Rewriter {
       modName = nsImport;
       isNamespaceImport = true;
     } else {
+      // TODO(martinprobst): why doesn't this need stripping "rootDir"?
       modName = this.host.pathToModuleName(this.file.fileName, tsImport);
     }
 
@@ -383,7 +384,8 @@ class ES5Processor extends Rewriter {
  * Converts TypeScript's JS+CommonJS output to Closure goog.module etc.
  * For use as a postprocessing step *after* TypeScript emits JavaScript.
  *
- * @param fileName The source file name.
+ * @param fileName The logical JavaScript source file name, i.e. with any
+ *     outDir prefix removed.
  * @param moduleId The "module id", a module-identifying string that is
  *     the value module.id in the scope of the module.
  * @param pathToModuleName A function that maps a filesystem .ts path to a
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1179,7 +1179,7 @@ class Annotator extends ClosureRewriter {
     // Convert the source fileName to a logical fileName.
     const rootDir = (this.tsOpts && this.tsOpts.rootDir) || '';
     const logicalFileName = this.file.fileName.startsWith(rootDir) ?
-        this.file.fileName.substring(rootDir.length) :
+        path.relative(rootDir, this.file.fileName) :
         this.file.fileName;
     const moduleNamespace =
         nsImport !== null ? nsImport : this.host.pathToModuleName(logicalFileName, importPath);
@@ -2013,7 +2013,7 @@ export function emitWithTsickle(
           }
           const outDir = program.getCompilerOptions().outDir || '';
           const logicalFileName =
-              fileName.startsWith(outDir) ? fileName.substring(outDir.length) : fileName;
+              fileName.startsWith(outDir) ? path.relative(outDir, fileName) : fileName;
           content = es5processor.convertCommonJsToGoogModuleIfNeeded(
               host, modulesManifest, logicalFileName, content);
         } else {
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -32,6 +32,10 @@ export interface AnnotatorHost {
    * by default.
    */
   logWarning?: (warning: ts.Diagnostic) => void;
+  /**
+   * Convert an import path in the given context into a Closure module name.
+   * context and import path must have rootDir or outDir prefixes stripped.
+   */
   pathToModuleName: (context: string, importPath: string) => string;
   /**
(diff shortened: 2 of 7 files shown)


test/test_support.ts
```diff
@@ -38,8 +53,17 @@ export const compilerOptions: ts.CompilerOptions = {
   emitDecoratorMetadata: true,
   noEmitHelpers: true,
   jsx: ts.JsxEmit.React,
-  // Flags below are needed to make sure source paths are correctly set on write calls.
-  rootDir: path.resolve(process.cwd()),
+  /**
+   * The 'conceptual path' of source files in TypeScript is their path relative to the current
```
evmar: This term is ill-defined (which I'm sure you're aware of). Is it a bazel concept, a TypeScript concept, or something else?  Maybe a Closure-only concept?

test/test_support.ts
```diff
@@ -95,32 +119,47 @@ export function createProgram(
   return createProgramAndHost(sources, tsCompilerOptions).program;
 }
 
-export function createSourceCachingHost(
+/**
+ * Creates a ts.CompilerHost that only resolves the given source files.
+ *
+ * Map keys in sources should be relative paths. The host simulates an environment where the sources exist immediately underneath the current directory
+ *
+ * Additionally, this host caches
```
evmar: Trailed off.

test/test_support.ts
```diff
@@ -95,32 +119,47 @@ export function createProgram(
...
     sources: Map<string, string>,
     tsCompilerOptions: ts.CompilerOptions = compilerOptions): ts.CompilerHost {
   const host = ts.createCompilerHost(tsCompilerOptions);
 
+  function relativeToBase(p: string) {
+    if (!path.isAbsolute(p)) return p;
+    return path.relative(BASE_PATH, p);
+  }
+
+  // host.getCurrentDirectory = () => BASE_PATH;
```
evmar: XXX

src/es5processor.ts
```diff
@@ -270,6 +270,7 @@ class ES5Processor extends Rewriter {
       modName = nsImport;
       isNamespaceImport = true;
     } else {
+      // TODO(martinprobst): why doesn't this need stripping "rootDir"?
```
evmar: this.file.fileName is a TS object, which should have an absolute path.  But pathToModuleName docs say it expects a relative path?

src/tsickle.ts
```diff
@@ -1163,22 +1167,30 @@ class Annotator extends ClosureRewriter {
...
     // Scripts do not have a symbol. Scripts can still be imported, either as side effect imports or
     // with an empty import set ("{}"). TypeScript does not emit a runtime load for an import with
     // an empty list of symbols, but the import forces any global declarations from the library to
     // be visible, which is what users use this for. No symbols from the script need forward
     // declaration, so just return.
     if (!moduleSymbol) return;
-    const exports = this.typeChecker.getExportsOfModule(moduleSymbol);
+
+    // Convert the source fileName to a logical fileName.
+    const rootDir = (this.tsOpts && this.tsOpts.rootDir) || '';
```
evmar: Is rootDir guaranteed to have a trailing slash?  I think maybe you want path.relative() here.

test/test_support.ts
```diff
@@ -19,6 +19,21 @@ import {toClosureJS} from '../src/main';
...
+ * The BASE_PATH that test source file paths are relative to.
+ *
+ * This is the path that contains the 'test_files' or 'test' folder. This can be the bazel runfiles
+ * directory (i.e. the workspace root), or potentially a subdirectory if the tsickle source
+ * code exists in a subdirectory in a different bazel workspace.
+ */
+const BASE_PATH = (() => {
+  const runfiles = process.env['RUNFILES'];
+  const candidate =
+      path.join(runfiles, 'google3', 'third_party', 'javascript', 'node_modules', 'tsickle');
```
evmar: FWIW google3 will never exist in a non-Unix form, so you might as well write the second arg to this function
'google3/third_party/javascript/node_modules/tsickle'
as that will be easier to grep for (e.g. if we ever move it to a versioned subdir)

src/es5processor.ts
mprobst: Yeah, that's what I'm asking myself. But in tests, this seems to get the right, relative path.

test/test_support.ts
mprobst: Yeah, done.

src/tsickle.ts

src/es5processor.ts
evmar: In the past when this has happened, it's because we fed in a bad path into one of the topmost TS APIs (like the one that parses tsconfig.json).  "Bad" here is a relative path, because TS works with absolute paths.  The symptom is that most things mostly still work up until you get some really crazy path stuff where some files are absolute and others aren't.

It's probably not worth tracking down in this CL.

mprobst:
For the record, this totally fails if root our out dir are relative paths, e.g. `"../../"`. I think we still don't have a good understanding of how we should handle paths, which is why this is broken ...

mprobst:
Obsolete.

PR #646, opened 2017-11-07
Break cyclic dependencies in tsickle.ts

kyliau:
This commit fixes the cyclic dependencies caused by `hasModifierFlag`
and `isDtsFileName` methods in `tsickle` by moving them to `util`.


mprobst:
Ping?

PR #647, opened 2017-11-08
Fix support for peerDependencies

paladox:
With npm 3+ deps in peerDependencies are not installed automatically and installing tsickle is complaining with

npm WARN tsickle@0.23.6 requires a peer of typescript@2.3.4 but none is installed. You must install peer dependencies yourself.

I think this may be causing https://bugs.chromium.org/p/gerrit/issues/detail?id=7630

By putting it in dependencies and peerDependencies it will not complain any more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,9 @@
     "source-map-support": "^0.4.2",
     "typescript": "~2.5.3"
   },
+  "peerDependencies": {
+    "typescript": "~2.5.3"
+  },
   "devDependencies": {
     "@bazel/typescript": "^0.3.1",
     "@types/chai": "^3.4.32",
--- a/package.json
+++ b/package.json
@@ -45,8 +45,7 @@
     "merge2": "^1.0.2",
     "mocha": "^3.2.0",
     "temp": "^0.8.1",
-    "tslint": "^5.4.2",
-    "typescript": "~2.5.3"
+    "tslint": "^5.4.2"
   },
   "scripts": {
     "prepublish": "bazel build ...",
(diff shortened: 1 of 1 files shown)


evmar:
@samccone any thoughts?

wardpeet:
@evmar I believe the this is no bug it's expected behaviour and shouldn't be moved to dependencies

paladox:
It seems it's because we were using the wrong version in peerDependencies.

Reading up more on this here https://codingwithspike.wordpress.com/2016/01/21/dealing-with-the-deprecation-of-peerdependencies-in-npm-3/

alexeagle:
I think this is already fixed by https://github.com/angular/tsickle/commit/eddfb2679dd88af6bc5ce63537fc546ec0b035ee#diff-b9cfc7f2cdf78a7f4b91a753d10865a2 can you confirm?

paladox:
Nope still says

/usr/local/bin/tsickle -> /usr/local/lib/node_modules/tsickle/src/main.js
npm WARN tsickle@0.25.5 requires a peer of typescript@>=2.4.2 <2.6 but none is installed. You must install peer dependencies yourself.

samccone:
@paladox do you have typescript installed on your system globally?

paladox:
yep.

i did npm install typescript -g

which installed typescript@2.6.1

samccone:
`typescript@>=2.4.2 <2.6 ` 

tsickle expects typescript < 2.6 to work, so since you have 2.6.1 installed it will still complain since you are not meeting the dependency constraints.

paladox:
Yep. The dep should probaly be included in tsickle

Since doing npm install typescript installs the latest release.

mprobst:
So I think this is WAI then?

PR #650, opened 2017-11-10
function types: optional and vararg arguments.

mprobst:


test/golden_tsickle_test.ts
```diff
@@ -6,7 +6,9 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-import {expect} from 'chai';
+import {expect, use as chaiUse} from 'chai';
+// tslint:disable-next-line:no-require-imports chai-diff needs a CommonJS import.
```
evmar: How come?  You can
import * as chaiDiff from 'chai-diff';
and then use it as a function.

package.json
```diff
@@ -35,6 +35,7 @@
     "@types/source-map": "^0.5.1",
     "@types/source-map-support": "^0.2.27",
     "chai": "^3.5.0",
+    "chai-diff": "^1.0.1",
```
evmar: This is good, but FWIW in my experience it's easier to just let the thing write out new goldens and examine the output with 'git diff'.

test/golden_tsickle_test.ts
mprobst: There's a wontfix TypeScript issue about it:
https://github.com/Microsoft/TypeScript/issues/5073

package.json
mprobst: When I'm interactively tweaking the emit, I find it much easier to see a good diff there than re-gen the goldens. That also runs a high risk of missing a change. But I understand workflows are different.

PR #652, opened 2017-11-13
Support for TypeScript 2.4

chuckjaz:
This is needed for Angular to switch to TypeScript 2.5.

That's because Angular users don't select a version of tsickle to match their typescript version, rather they get the tsickle dependency from `@angular/compiler-cli`. So we want to release a single tsickle that works with the entire range of typescript versions that Angular supports.


src/BUILD
```diff
@@ -1,6 +1,18 @@
...
+TS24 = "//test_ts24:node_modules/typescript/lib/typescript.d.ts"
+
+# Copy the 2.4 version of typescript's typings into the project.
+# We don't want to copy a file that's licensed under an external license into
+# our source tree, and this mechanism allows the :src library to import it.
+genrule(
+    name = "typescript-2.4",
+    outs = ["typescript-2.4.d.ts"],
+    srcs = [TS24],
+    cmd = "cp $(location %s) $@" % TS24
```
mprobst: I think you can do `$(location $<)` or `$(location $(SRC))` or so, right?

src/es5processor.ts
```diff
@@ -323,11 +322,17 @@ class ES5Processor extends Rewriter {
     switch (call.expression.kind) {
       case ts.SyntaxKind.Identifier:
         const ident = call.expression as ts.Identifier;
-        if (ident.text !== '__export') return null;
+        // TS 2.4 compatibility: accept three leading underscores
```
mprobst: Nit: can we make this comment something that's more obviously a greppable tag? `// TS_24_COMPAT: ...`?

src/typescript.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
mprobst: Can you add a fileoverview comment explaining what this does (i.e. uses 2.4 types on 2.5 API)?

src/typescript.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// tslint:disable:no-any We need to do various unsafe casts between TS versions
+
+export function getEmitFlags(node: ts.Node): ts.EmitFlags|undefined {
+  return (ts as any).getEmitFlags(node);
```
mprobst: add a comment why this works?

mprobst:
Note that the E2E Closure tests are currently broken, see #653.

src/es5processor.ts
alexeagle: sure, but do you think we can remember to follow that convention when we add backwards compatibility logic next time?

src/BUILD
alexeagle: good catch

src/typescript.ts
alexeagle: done

src/typescript.ts
alexeagle: done

src/typescript.ts
```diff
@@ -0,0 +1,46 @@
...
+
+// TS 2.5 typescript.d.ts does not expose getEmitFlags.
```
mprobst: but it still exists? might be worth mentioning.

src/es5processor.ts
mprobst: If we do use a more obvious tag like name, we at least have a fighting chance :-)

PR #653, opened 2017-11-14
Fix e2e tests and support PromiseLike

mprobst:
We accidentally disabled error checking of our E2E tests in #623. This restores the test case, and fixes a number of issues that have crept in.

Some of those problems were in the code even before accidentally disabling the tests, but only exposed when running with `--new_type_inf` (Closure's New Type Inference). The recurring pattern here is that Closure does not actually allow `typedef` aliases of types with generic type arguments.

This also adds support for PromiseLike and PromiseConstructor in our `closure_externs.js`.

### DIFF
--- a/src/closure_externs.js
+++ b/src/closure_externs.js
@@ -29,21 +29,35 @@ var HTMLTableHeaderCellElement;
 
 /**
  * Closure's NodeList is parameterized itself, there is no NodeListOf.
- * @typedef {!NodeList}
+ * @constructor
+ * @template T
+ * @extends {NodeList<T>}
  */
 var NodeListOf;
 
 /**
  * Closure models this as a plain Array.
- * @typedef {Array<string>|null}
+ * @typedef {!IArrayLike<string>|null}
  */
 var RegExpExecArray;
 
-/** @typedef {!Array} */
-var ReadonlyArray;
+/**
+ * @record
+ * @template T
+ * @extends {IArrayLike<T>}
+ */
+function ReadonlyArray() {}
 
-/** @typedef {!Map} */
-var ReadonlyMap;
+/**
+ * @constructor
+ * @template K, V
+ * @extends {Map<K, V>}
+ */
+function ReadonlyMap() {}
 
-/** @typedef {!Set} */
-var ReadonlySet;
+/**
+ * @constructor
+ * @template T
+ * @extends {Set<T>}
+ */
+function ReadonlySet() {}
--- a/src/closure_externs.js
+++ b/src/closure_externs.js
@@ -61,3 +61,14 @@ function ReadonlyMap() {}
  * @extends {Set<T>}
  */
 function ReadonlySet() {}
+
+/**
+ * @record
+ * @template T
+ * @extends {IThenable<T>}
+ */
+function PromiseLike() {};
+
+/** @typedef {function(new:Promise)} */
+var PromiseConstructor;
+
--- a/test/e2e_test.ts
+++ b/test/e2e_test.ts
@@ -20,8 +20,8 @@ export function checkClosureCompile(
   const CLOSURE_COMPILER_OPTS: closure.CompileOptions = {
     'checks_only': true,
     'jscomp_error': 'checkTypes',
-    // TODO(martinprobst): enable when NTI is ready. Currently doesn't support enum union types.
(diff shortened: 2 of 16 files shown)


src/closure_externs.js
```diff
@@ -29,21 +29,35 @@ var HTMLTableHeaderCellElement;
...
+ * @constructor
+ * @template T
+ * @extends {NodeList<T>}
  */
 var NodeListOf;
 
 /**
  * Closure models this as a plain Array.
- * @typedef {Array<string>|null}
+ * @typedef {!IArrayLike<string>|null}
```
alexeagle: rather than copying parts of this file, is it possible we could use the real thing by copying from somewhere during the build?

test/e2e_test.ts
```diff
@@ -49,5 +49,5 @@ describe('golden file tests', () => {
     goldenJs.push('test_files/import_from_goog/closure_OtherModule.js');
     const externs = tests.map(t => t.externsPath).filter(fs.existsSync);
     checkClosureCompile(goldenJs, externs, done);
-  });
+  }, 10000 /* ms timeout */);
```
alexeagle: should this interact with process.env['TEST_TIMEOUT'] in some way? do we really care to have jasmine enforce timeouts as well as bazel?

mprobst: So `small` tests have a 60 second timeout in bazel (go figure...). https://docs.bazel.build/versions/master/test-encyclopedia.html

That seems a bit excessive, and also we don't have a guarantee that this is the only test executing in its suite. So I'd rather keep it tight here and relevant to this local test, not deducted from some outer limitation.

src/closure_externs.js
mprobst: Not sure I follow, this file is the real thing, isn't? It's the bridge aliasing TypeScript types to Closure types, there's no other artefact that has the same functionality.

test/e2e_test.ts
```diff
@@ -11,15 +11,16 @@ import * as closure from 'google-closure-compiler';
...
-    jsFiles: string[], externsFiles: string[], done: (err?: Error) => void) {
+export function checkClosureCompile(jsFiles: string[], externsFiles: string[], done: DoneFn) {
   const startTime = Date.now();
   const total = jsFiles.length;
   if (!total) throw new Error('No JS files in ' + JSON.stringify(jsFiles));
 
   const CLOSURE_COMPILER_OPTS: closure.CompileOptions = {
     'checks_only': true,
     'jscomp_error': 'checkTypes',
+    // TODO(martinprobst): enable when NTI is ready. Currently doesn't support enum union types.
```
evmar: It looks like you worked around the enum thing elsewhere, so adjust this comment.  I kinda suspect we should just avoid NTI until a user demands it.

src/closure_externs.js
alexeagle: Ah, I understand now, I thought these were duplicated from the built-in externs (eg. https://github.com/google/closure-compiler/tree/master/externs/browser)

test_files/promiseconstructor/promiseconstructor.ts
```diff
@@ -0,0 +1,3 @@
+function f(promiseCtor?: typeof Promise): Promise<void> {
```
evmar: Can you add a comment to this file about what the test is concerned with?  It's just trying to use the PromiseCtor type?

test/e2e_test.ts
alexeagle: If this is the only test where we set the timeout, then for other tests we rely on `jasmine.DEFAULT_TIMEOUT_INTERVAL` being the shorter value you like.

In https://github.com/bazelbuild/rules_nodejs/pull/53 I propose to make the default timeout longer, so now all tests other than this one would get 5m (since they are timeout="moderate")

Maybe this is all too nit-picky.

test/e2e_test.ts
mprobst: Whoops, leftover comment. I changed only the comment, I think it's useful for us to run in NTI, just for the more precise error checks.

test_files/promiseconstructor/promiseconstructor.ts
mprobst: Done.

PR #654, opened 2017-11-14
Avoid emitting unsupported template parameters 

mprobst:
Closure restricts the locations where template parameters can be used. Curiously this also depends on the version of the type checker in use, so we have to run our test suite for both.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1391,6 +1391,8 @@ class Annotator extends ClosureRewriter {
     // TypeScript drops exports that are never assigned to (and Closure
     // requires us to not assign to typedef exports).  Instead, emit the
     // "exports.foo;" line directly in that case.
+    this.newTypeTranslator(node).blacklistTypeParameters(node.typeParameters);
+
     this.emit(`\n/** @typedef {${this.typeToClosure(node)}} */\n`);
     if (hasModifierFlag(node, ts.ModifierFlags.Export)) {
       this.emit('exports.');
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -750,7 +750,8 @@ class Annotator extends ClosureRewriter {
         }
 
         this.emitFunctionType([fnDecl], tags);
-        this.newTypeTranslator(fnDecl).blacklistTypeParameters(fnDecl.typeParameters);
+        this.newTypeTranslator(fnDecl).blacklistTypeParameters(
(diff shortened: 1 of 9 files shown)


src/tsickle.ts
```diff
@@ -758,6 +758,7 @@ class Annotator extends ClosureRewriter {
         }
 
         this.emitFunctionType([fnDecl], tags);
+        this.newTypeTranslator(fnDecl).blacklistTypeParameters(fnDecl.typeParameters);
```
evmar: How does this work?  It looks like it creates a type translator, modifies its state, then throws it away...?

mprobst: I've tried to document on the method we're calling. Indeed, this mutates the Map that we're using for aliased symbols. Not super cool, but OTOH pretty harmless, and keeps the blacklisting code in one place. Any other ideas?

evmar: But it only mutates the map on the TypeTranslator instance, so how does the change persist?

mprobst: It's the same map - we're using only one map of aliases per file.

test/e2e_test.ts
```diff
@@ -37,17 +47,10 @@ export function checkClosureCompile(jsFiles: string[], externsFiles: string[], d
...
-    goldenJs.push('test_files/clutz.no_externs/some_name_space.js');
-    goldenJs.push('test_files/clutz.no_externs/some_other.js');
-    goldenJs.push('test_files/import_from_goog/closure_Module.js');
-    goldenJs.push('test_files/import_from_goog/closure_OtherModule.js');
-    const externs = tests.map(t => t.externsPath).filter(fs.existsSync);
-    checkClosureCompile(goldenJs, externs, done);
+  it('compile with Closure (New Type Inference)', (done) => {
+    checkClosureCompile(true /* NTI */, done);
+  }, 15000 /* ms timeout */);
+  it('compile with Closure (Old Type Inference)', (done) => {
```
evmar: We could do these in parallel maybe (?)

src/tsickle.ts
mprobst: I've changed the method to take the map as a parameter, and added more docs.

test/e2e_test.ts
mprobst: I'm kind of scarred from gulp trying to run all kinds of stuff in parallel and spamming the console with the respective outputs. OTOH I think we capture all output here, so that won't be a problem. I'll give it a spin.

PR #656, opened 2017-11-14
Only resolve aliases when emitting typedefs.

mprobst:
This is based on PR #654, please only review the last commit.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -492,8 +492,10 @@ class ClosureRewriter extends Rewriter {
    * @param context The ts.Node containing the type reference; used for resolving symbols
    *     in context.
    * @param type The type to translate; if not provided, the Node's type will be used.
+   * @param resolveAlias If true, do not emit aliases as their symbol, but rather as the resolved
+   *     type underlying the alias. This should be true only when emitting the typedef itself.
    */
-  typeToClosure(context: ts.Node, type?: ts.Type): string {
+  typeToClosure(context: ts.Node, type?: ts.Type, resolveAlias?: boolean): string {
     if (this.host.untyped) {
       return '?';
     }
@@ -502,7 +504,7 @@ class ClosureRewriter extends Rewriter {
     if (!type) {
       type = typeChecker.getTypeAtLocation(context);
     }
-    return this.newTypeTranslator(context).translate(type);
+    return this.newTypeTranslator(context).translate(type, resolveAlias);
   }
 
   newTypeTranslator(context: ts.Node) {
@@ -1396,7 +1398,8 @@ class Annotator extends ClosureRewriter {
     this.newTypeTranslator(node).blacklistTypeParameters(
         this.symbolsToAliasedNames, node.typeParameters);
 
-    this.emit(`\n/** @typedef {${this.typeToClosure(node)}} */\n`);
+    const typeStr = this.typeToClosure(node, undefined, true /* resolveAlias */);
+    this.emit(`\n/** @typedef {${typeStr}} */\n`);
     if (hasModifierFlag(node, ts.ModifierFlags.Export)) {
       this.emit('exports.');
     } else {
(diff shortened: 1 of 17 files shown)


src/type-translator.ts
```diff
@@ -125,10 +125,10 @@ export function symbolToDebugString(sym: ts.Symbol): string {
 /** TypeTranslator translates TypeScript types to Closure types. */
 export class TypeTranslator {
   /**
-   * A list of types we've encountered while emitting; used to avoid getting stuck in recursive
-   * types.
+   * A list of type literals we've encountered while emitting; used to avoid getting stuck in
+   * recursive types.
    */
-  private readonly seenTypes: ts.Type[] = [];
+  private readonly seenTypeLiterals: ts.Type[] = [];
```
evmar: You can use a Set<> here.

test_files/type_alias_imported/type_alias_exporter.ts
```diff
@@ -0,0 +1,3 @@
+export class X { private x: string; }
+export class Y { private x: string; }
+export type XY = X|Y;
```
evmar: Docs in here about what this is looking for are again helpful.

src/type-translator.ts
mprobst: I stumbled over that, and was wondering if you used an array on purpose given these are very likely never more than one or two types.

test_files/type_alias_imported/type_alias_exporter.ts
mprobst: Done.

src/tsickle.ts
```diff
@@ -1128,6 +1128,18 @@ class Annotator extends ClosureRewriter {
...
+      // module because it's only used in type positions, the JSDoc comments still reference a valid
+      // Closure level symbol.
+      const sym = this.typeChecker.getSymbolAtLocation(decl.moduleSpecifier);
+      if (!sym) {
+        return true;  // modules might not have a symbol if they are unused.
+      }
+      // forwardDeclare all symbols that can be imported from the module.
+      const namedSyms =
+          this.typeChecker.getExportsOfModule(sym).map(sym => ({name: sym.name, sym}));
+      this.forwardDeclare(decl.moduleSpecifier, namedSyms, false);
```
evmar: Does this mean any module that imports @angular/core ends up with a forwardDeclare line per function in the angular API?

test_files/import_prefixed/import_prefixed_mixed.ts
```diff
@@ -0,0 +1,4 @@
+import * as exporter from './exporter';
+
+let someVar: exporter.TypeExport;
```
evmar: Comment in these files about what property you're looking for.

test_files/type_and_value/type_and_value.js
```diff
@@ -7,13 +7,14 @@ goog.module('test_files.type_and_value.type_and_value');var module = module || {
  */
 
 var conflict = goog.require('test_files.type_and_value.module');
+const tsickle_forward_declare_1 = goog.forwardDeclare("test_files.type_and_value.module");
```
evmar: I wonder if the relative order of goog.require / goog.forwardDeclare matters.

src/tsickle.ts
mprobst: No. We emit one `forwardDeclare` (as you can see in the test), but then set all symbols exported by that module to be aliased to `tsickle_forward_declare_X.TheSymbolName`.

test_files/type_and_value/type_and_value.js
mprobst: I don't think so. Fingers crossed.

test_files/import_prefixed/import_prefixed_mixed.ts
mprobst: Done.

PR #658, opened 2017-11-20
 Special case union type translation.

mprobst:
TypeScript "flattens" type aliases for union types when it adds
alternatives to a union. This means tsickle cannot reconstruct the
correct syntactical representation of the union type, as any type
aliases constructing it have been resolved.

This PR changes tsickle to pass around the the syntactical form
(i.e. `ts.TypeNode` AST nodes) in addition to the semantic form
(`ts.Type` nodes), so that tsickle can emit intermediate type aliases.
Translation of union types then uses the syntactical form if present.

This required moving around some code, including handling optional
properties, and finding the right generic type arguments for rest
parameters.

Intersection types do not need to be handled as Closure cannot handle
them at all.


mprobst:
I'm not super happy with this change. It seems fragile, and mixing concepts we shouldn't mix.

Maybe the alternative would be emitting additional imports for the union type members? But that would require a two pass emit, and naming symbols whose location we don't know (yet).

mprobst:
Update: I've checked how `.d.ts` declaration emit names types, it's following the same pattern of using the syntactical nodes to generate types. I think this change is the correct approach then.

src/tsickle.ts
```diff
@@ -1360,27 +1374,26 @@ class Annotator extends ClosureRewriter {
...
+    // to mark the property as optional in Closure, it must have "|undefined",
     // so the Closure type must be ?|undefined.
-    if (optional && type === '?') type += '|undefined';
+    // The underlying TypeScript type of the property is e.g. "string|undefined"
+    // when a concrete type is given. However translation doesn't include
+    // "|undefined" in two corner cases:
+    // - for union types, we translate the syntactical type node
+    // - for any types, TS translation only includes ? (as any includes
+    //   "|undefined"), but Closure's ? type does not include absence
+    if (optional) type += '|undefined';
```
LucasSloan: This will add `|undefined` even in cases not explained above, unless there's some logic elsewhere I'm not seeing - maybe say why that's ok in the comment?a

test_files/type/type.js
```diff
@@ -18,8 +18,8 @@ let /** @type {!Array<?>} */ typeTupleTuple = [[1, 2]];
 let /** @type {!Array<?>} */ typeTupleTuple2 = [[1, 2], ''];
 let /** @type {(string|boolean)} */ typeUnion = Math.random() > 0.5 ? false : '';
 let /** @type {(string|boolean)} */ typeUnion2 = Math.random() > 0.5 ? false : '';
-let /** @type {{optional: (undefined|boolean)}} */ typeOptionalField = {};
-let /** @type {{optional: (undefined|string|boolean)}} */ typeOptionalUnion = {};
+let /** @type {{optional: ((undefined|boolean)|undefined)}} */ typeOptionalField = {};
```
LucasSloan: `undefined|boolean|undefined`?

README.md
```diff
@@ -118,6 +118,24 @@ to install the dependencies.
   `clang-format`,
 - `yarn test` runs unit tests, e2e tests and checks the source code formatting.
 
+### Debugging
```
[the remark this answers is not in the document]
mprobst: It's a separate commit, so it retains history including a proper commit message. I think that's fine, given that github allows reviewing in isolation if needed.

src/tsickle.ts
```diff
@@ -495,16 +502,23 @@ class ClosureRewriter extends Rewriter {
...
-  typeToClosure(context: ts.Node, type?: ts.Type, resolveAlias?: boolean): string {
+  typeToClosure(
+      context: ts.Node, typeNode: ts.TypeNode|undefined, type?: ts.Type,
+      resolveAlias?: boolean): string {
     if (this.host.untyped) {
       return '?';
     }
 
     const typeChecker = this.typeChecker;
+    // if (context.type && ts.isUnionTypeNode(context.type)) {
```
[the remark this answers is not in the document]

src/tsickle.ts
```diff
@@ -1360,27 +1374,26 @@ class Annotator extends ClosureRewriter {
...
-    // typeToClosure translation handles it correctly, and string|undefined is
-    // how you write an optional property in Closure.
-    //
-    // But in the special case of an optional property with type any:
-    //   foo?: any;
-    // The TypeScript type of the property is just "any" (because any includes
-    // undefined as well) so our default translation of the type is just "?".
-    // To mark the property as optional in Closure it must have "|undefined",
+    // to mark the property as optional in Closure, it must have "|undefined",
     // so the Closure type must be ?|undefined.
```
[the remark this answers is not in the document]
mprobst: Added to the comment.

test_files/type/type.js
mprobst: Yeah. Doesn't hurt, and makes the code easier/more consistent.

src/type_translator.ts
```diff
@@ -378,6 +378,26 @@ export class TypeTranslator {
     }
   }
 
+  /**
+   * Special-case translation of union types to work off the syntactical type representation
+   * (`ts.Node` object), as opposed to the semantic type representation (`ts.Type` object).
+   *
+   * Unlike with othe types, TypeScript expands aliased union types as they are added into another
+   * union. This means we cannot recover the original type nodes and names from the union type, and
```
evmar: s/the union type/the ts.Type/ perhaps?  It confusing here because when you say "the union type" you mean something other than the 'UnionTypeNode' seen just below.

src/type_translator.ts
```diff
@@ -378,6 +378,26 @@ export class TypeTranslator {
...
+   * must fall back to the syntactical representation.
+   */
+  translateUnionTypeNode(n: ts.UnionTypeNode) {
+    return '(' +
+        n.types
+            .map(tn => {
+              const type = this.typeChecker.getTypeAtLocation(tn);
+              return this.translate(type, false);
```
evmar: What is 'false' here?  Needs a comment.

README.md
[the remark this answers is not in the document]
LucasSloan: Our usual practice is to squash all the commits in a PR - are you planning on squashing everything but this into one?

test_files/declare_class_overloads/externs.js
```diff
@@ -130,9 +130,9 @@ OverloadValueOf.prototype.valueOf = function() {};
 function Merged() {}
 
 /**
- * @param {(string|number|!Array<!OverloadBigMix>)} a_or_c_or_e_or_f
+ * @param {(string|number|!Array<!OverloadBigMix>|!Array<!OverloadBigMix>)} a_or_c_or_e_or_f
```
evmar: I recall at some point in some code we grab the set of union elements and drop duplicates.  I guess it's a different code path than here...

test_files/enum/enum.js
```diff
@@ -27,7 +27,7 @@ enumTestFunction(enumTestValue);
 let /** @type {EnumTest1} */ enumTestLookup = EnumTest1["XYZ"];
 let /** @type {?} */ enumTestLookup2 = EnumTest1["xyz".toUpperCase()];
 // Verify that unions of enum members and other values are handled correctly.
-let /** @type {(boolean|EnumTest1)} */ enumUnionType = EnumTest1.XYZ;
+let /** @type {(EnumTest1|boolean)} */ enumUnionType = EnumTest1.XYZ;
```
evmar: Is this order stable?  That is, what causes this to have boolean on the right, and will it persist across compiles?

test_files/typedef/typedef.ts
```diff
@@ -3,4 +3,15 @@ var y: MyType = 3;
 
 type Recursive = {value: number, next: Recursive};
 
+// TypeScript expands union members when constructing, which drops the
+// distinction in unionUserInUnion between string|number|boolean and
+// UnionTypeAlias|boolean. That's a problem if the constituent types are not
+// valid symbols in the lexical context (e.g. not imported).
```
evmar: You write the problem here, but trail off before describing why the problem isn't a problem.  Either briefly describe what you expect the test to be showing here, or ref the relevant code or something?

README.md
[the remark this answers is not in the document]
mprobst: I've organized this PR such that the commits are individually meaningful. I'll rebase them on master when I push.

test_files/enum/enum.js
mprobst: Previously, this was ordered by the sequence of types in the resolved union type, handled by TS. Now this is ordered by the lexical spelling, i.e. comes out the way the user wrote this. Both are stable, it's entries in an array.

src/type_translator.ts
mprobst: Code removed, no longer applies.

src/type_translator.ts
mprobst: No longer applies.

src/tsickle.ts
mprobst: Added parentheses.

test_files/typedef/typedef.ts
mprobst: done.

mprobst:
Obsolete, superseded by PR #663.

PR #660, opened 2017-11-22
Do not emit generic types on type aliases.

mprobst:
Closure does not support generic types in typedefs. Closure's old type
inference simply ignores the arguments, but its new type inference
reports "incorrect generic type instantiation". This change avoids the
problem by specifically not emitting type arguments if the referenced
type is a type alias (and thus a type definition in Closure).

This change is based on PR #658.


mprobst:
Green in g3.

mprobst:
Obsolete, superseded by PR #663.

PR #662, opened 2017-11-28
exporting main() to ease embedding without fork

PeterHorvathIsento:
Hi,

this trivial change would make embedding into external tools without forking much easier.

Regards,

PH


PeterHorvathIsento:
I signed it!

src/main.ts
```diff
@@ -156,7 +156,7 @@ export function toClosureJS(
       program, transformerHost, compilerHost, options, undefined, writeFile);
 }
 
-function main(args: string[]): number {
+export function main(args: string[]): number {
```
mprobst: We thought that users would call into `toClosureJS` instead of using this method. After all, there's no point in parsing settings from string arguments if you pass them programatically, and the only other interesting thing is writing files and reporting diagnostics, which 3rd party tools probably want to take over.

Can you share your use case?

HorvathAkosPeter: Hello, sorry for the late answer.

The use case is that

1. User made his project working with `./node_modules/.bin/tsickle`
2. He wants to automatize this, but he doesn't like node processes called by another nodes.

I think some functionality should exist in the `main.ts` which makes possible the calling of a tsickle instance, but specifying the command line argument list. To do that, we now have to duplicate the whole `main.ts` functionality, which is dirty.

However, the functionality can be done also with other ways: for example, cloning `process` and overriding its `process.args` and `process.exit` methods. This is what I do now. Also this is dirty, but much lesser.

HorvathAkosPeter: ...finally I've found that this main() isn't so complex, so I rather essentially duplicated it. Thank you anyways!

mprobst: Makes sense, thanks for validating my understanding, and happy to serve as a rubber duck for you ;-)

PR #663, opened 2017-11-28
forward declare symbols not in the local scope.

mprobst:
TypeScript's type checker expands aliases in union types as they are
constructed. This is apparently a design limitation and hard to change
for reasons including performance.

In tsickle, the expanded symbols might however not be available at the
use site, e.g. if a union type is declared in `fileA.ts`, contains
symbols available locally, but then imported in `fileB.ts` where its
constituent types are not available:

    // fileA
    export class LocalType {}
    export type TypeDef = LocalType|number;

    // fileB
    import {TypeDef} from './fileA';
    let useTypeDef: TypeDef|string;

Tsickle would emit `LocalType|number|string` as the type for
`useTypeDef`. This change fixes the issue by ensuring all symbols that
are used have an appropriate forward declare. This turns out to be
easier than mapping back into the original type or trying to emit the
syntactical type.

### DIFF
--- a/src/source_map_utils.ts
+++ b/src/source_map_utils.ts
@@ -109,48 +109,37 @@ export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenera
   return SourceMapGenerator.fromSourceMap(sourceMapTextToConsumer(sourceMapText));
 }
 
+/**
+ * A position in a source map. All offsets are zero-based.
+ */
 export interface SourcePosition {
-  // 0 based
+  /** 0 based */
   column: number;
-  // 0 based
+  /** 0 based */
   line: number;
-  // 0 based
+  /** 0 based full offset in the file. */
   position: number;
 }
 
 export interface SourceMapper {
+  /**
+   * Logically shift all source positions by `offset`.
+   *
+   * This method is useful if code has to prepend additional text to the generated output after
+   * source mappings have already been generated. The source maps are then transparently adjusted
+   * during TypeScript output generation.
+   */
+  shiftByOffset(offset: number): void;
+  /**
+   * Adds a mapping from `originalNode` in `original` position to its new location in the output,
+   * spanning from `generated` (an offset in the file) for `length` characters.
+   */
   addMapping(
       originalNode: ts.Node, original: SourcePosition, generated: SourcePosition,
       length: number): void;
 }
 
 export const NOOP_SOURCE_MAPPER: SourceMapper = {
-  // tslint:disable-next-line:no-empty
-  addMapping() {}
+  shiftByOffset() {/* no-op */},
+  addMapping() {/* no-op */},
 };
-
-export class DefaultSourceMapper implements SourceMapper {
-  /** The source map that's generated while rewriting this file. */
-  public sourceMap = new SourceMapGenerator();
-
-  constructor(private fileName: string) {
-    this.sourceMap.addMapping({
-      // tsc's source maps use 1-indexed lines, 0-indexed columns
-      original: {line: 1, column: 0},
-      generated: {line: 1, column: 0},
-      source: this.fileName,
-    });
-  }
-
-  addMapping(node: ts.Node, original: SourcePosition, generated: SourcePosition, length: number):
-      void {
-    if (length > 0) {
-      this.sourceMap.addMapping({
-        // tsc's source maps use 1-indexed lines, 0-indexed columns
-        original: {line: original.line + 1, column: original.column},
-        generated: {line: generated.line + 1, column: generated.column},
-        source: this.fileName,
-      });
-    }
-  }
-}
--- a/src/transformer_sourcemap.ts
+++ b/src/transformer_sourcemap.ts
@@ -11,17 +11,25 @@ import {isTypeNodeKind, updateSourceFileNode, visitNodeWithSynthesizedComments} 
 import * as ts from './typescript';
 
 /**
- * @fileoverview Creates a TypeScript transformer that parses code into a new `ts.SourceFile`,
- * marks the nodes as synthetic and where possible maps the new nodes back to the original nodes
- * via sourcemap information.
+ * Creates a TypeScript transformer based on a source->text transformation.
+ *
+ * TypeScript transformers operate on AST nodes. Newly created nodes must be marked as replacing an
+ * older AST node. This shim allows running a transformation step that's based on emitting new text
+ * as a node based transformer. It achieves that by running the transformation, collecting a source
+ * mapping in the process, and then afterwards parsing the source text into a new AST and marking
+ * the new nodes as representations of the old nodes based on their source map positions.
+ *
+ * The process marks all nodes as synthesized except for a handful of special cases (identifiers
+ * etc).
  */
 export function createTransformerFromSourceMap(
-    operator: (sourceFile: ts.SourceFile, sourceMapper: SourceMapper) =>
+    sourceBasedTransformer: (sourceFile: ts.SourceFile, sourceMapper: SourceMapper) =>
         string): ts.TransformerFactory<ts.SourceFile> {
   return (context) => (sourceFile) => {
     const sourceMapper = new NodeSourceMapper();
+    const transformedSourceText = sourceBasedTransformer(sourceFile, sourceMapper);
     const newFile = ts.createSourceFile(
-        sourceFile.fileName, operator(sourceFile, sourceMapper), ts.ScriptTarget.Latest, true);
+        sourceFile.fileName, transformedSourceText, ts.ScriptTarget.Latest, true);
     const mappedFile = visitNode(newFile);
     return updateSourceFileNode(sourceFile, mappedFile.statements);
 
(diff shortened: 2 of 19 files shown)


src/tsickle.ts
```diff
@@ -566,7 +578,44 @@ class Annotator extends ClosureRewriter {
...
+    // Because tsickle runs with the --declaration flag, all types referenced from exported types
+    // must be exported, too, so there must either be some declaration that is exported, or the
+    // symbol is actually a global declaration (declared in a script file, not a module).
+    const decl = sym.declarations.find(d => hasModifierFlag(d, ts.ModifierFlags.ExportDefault));
+    if (!decl) return;
+    // Exclude symbols declared in `declare global {...}` blocks, they are glocal and don't need
+    // imports.
+    let current: ts.Node|undefined = decl;
+    while (current) {
+      if (current.flags & ts.NodeFlags.GlobalAugmentation) return;
```
evmar: Is it possible there's some other declaration that satisfies your requirements?  (E.g. that .some picked the wrong one above?)  Might be worth refactoring so you call .some() over all declarations and do this filtering within the .some?

src/tsickle.ts
```diff
@@ -566,7 +578,44 @@ class Annotator extends ClosureRewriter {
...
+      current = current.parent;
+    }
+    // Actually import the symbol.
+    const sf = decl.getSourceFile();
+    const fn = sf.fileName;
```
evmar: Only used once and a bit far away, might be worth just inlining it.

src/tsickle.ts
```diff
@@ -566,7 +578,44 @@ class Annotator extends ClosureRewriter {
...
+    const moduleSymbol = this.typeChecker.getSymbolAtLocation(sf);
+    if (!moduleSymbol) {
+      this.error(sf, `module with exports has no symbol`);
+      return;
+    }
+    // Already imported.
+    if (this.forwardDeclaredModules.has(moduleSymbol)) return;
+
+    const exportedSymbols = this.typeChecker.getExportsOfModule(moduleSymbol);
```
evmar: unused?

src/tsickle.ts
```diff
@@ -1165,28 +1195,34 @@ class Annotator extends ClosureRewriter {
...
-  private forwardDeclare(
-      specifier: ts.Expression, exportedSymbols: NamedSymbol[], isDefaultImport = false) {
-    if (this.host.untyped) return;
+  private forwardDeclare(specifier: ts.Expression, isDefaultImport = false) {
     const importPath = this.resolveModuleSpecifier(specifier);
+    const moduleSymbol = this.typeChecker.getSymbolAtLocation(specifier);
+    this.emit(this.getForwardDeclareText(importPath, moduleSymbol, isDefaultImport));
+  }
+
+  private getForwardDeclareText(
```
evmar: docs

src/tsickle.ts
```diff
@@ -1136,10 +1168,8 @@ class Annotator extends ClosureRewriter {
       if (!sym) {
         return true;  // modules might not have a symbol if they are unused.
       }
-      // forwardDeclare all symbols that can be imported from the module.
-      const namedSyms =
-          this.typeChecker.getExportsOfModule(sym).map(sym => ({name: sym.name, sym}));
-      this.forwardDeclare(decl.moduleSpecifier, namedSyms, false);
+      // forwardDeclare the module.
```
evmar: maybe redundant with the function name?

src/tsickle.ts
```diff
@@ -566,7 +578,44 @@ class Annotator extends ClosureRewriter {
...
+    if (!sym.declarations || sym.declarations.length === 0) return;
+    // A symbol declared in this file does not need to be imported.
+    if (sym.declarations.some(d => d.getSourceFile() === this.file)) return;
+    // Find an exported declaration.
+    // Because tsickle runs with the --declaration flag, all types referenced from exported types
+    // must be exported, too, so there must either be some declaration that is exported, or the
+    // symbol is actually a global declaration (declared in a script file, not a module).
+    const decl = sym.declarations.find(d => hasModifierFlag(d, ts.ModifierFlags.ExportDefault));
+    if (!decl) return;
+    // Exclude symbols declared in `declare global {...}` blocks, they are glocal and don't need
```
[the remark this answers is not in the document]
mprobst: Huh, genuine typo.

src/tsickle.ts

src/tsickle.ts

src/source_map_utils.ts
```diff
@@ -109,48 +109,37 @@ export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenera
...
-  // 0 based
+  /** 0 based full offset in the file. */
   position: number;
 }
 
 export interface SourceMapper {
+  /**
+   * Logically shift all source positions by `offset`.
+   *
+   * This method is useful if code has to prepend addition text to the generated output after source
```
evmar: additional

src/source_map_utils.ts
```diff
@@ -109,48 +109,37 @@ export function sourceMapTextToGenerator(sourceMapText: string): SourceMapGenera
...
+   * mappings have already been generated. The source maps are then transparently adjusted during
```
evmar: This makes me think we got the layering wrong.  We wanted to move all of this to using transformers, so perhaps this new code could try?

src/transformer_sourcemap.ts
```diff
@@ -11,17 +11,24 @@ import {isTypeNodeKind, updateSourceFileNode, visitNodeWithSynthesizedComments}
 import * as ts from './typescript';
 
 /**
- * @fileoverview Creates a TypeScript transformer that parses code into a new `ts.SourceFile`,
- * marks the nodes as synthetic and where possible maps the new nodes back to the original nodes
- * via sourcemap information.
+ * Creates a TypeScript transformer based on a source->text transformation.
+ *
+ * TypeScript transformers operate on AST nodes. Newly create nodes must be marked as replacing an
```
evmar: created

src/transformer_sourcemap.ts
```diff
@@ -11,17 +11,24 @@ import {isTypeNodeKind, updateSourceFileNode, visitNodeWithSynthesizedComments}
...
+ * older AST node. This shim allows running a transformation step that's based on emitting new text
+ * as a node based transformer. It achieves that by running the transformation, collecting a source
+ * mapping in the process, and then afterwards parsing the source text into a new AST and marking
+ * the new nodes as representations of the old nodes based on their source map positions.
+ *
+ * The process marks all nodes as synthesized except for a handful special cases (identifiers etc).
```
evmar: handful *of*

src/source_map_utils.ts
mprobst: Yeah, but this new code is affecting how we annotate, so for this we'd need to move annotate, which is kind of all of tsickle. I considered it, and we really have to do it sooner than later, but it's a bunch of work.

mprobst:
PTAL, addressed all comments I think.

src/source_map_utils.ts
evmar: I think to incrementally move code out of annotate, we will have to make it export data in a form such that the transformer wrapper can use it.  In this case you could export the list of things you want imported, and then the transformer that calls annotate would use that list to generate source nodes for the imports.

Maybe it's not worth worrying about right now, but (1) at our current rate the transformer thing will never happen, and (2) this is maybe an easy one to do it for because you genuinely don't want these imports in the source text (which is why you have to monkey with the source map).

src/tsickle.ts
```diff
@@ -557,16 +561,57 @@ class Annotator extends ClosureRewriter {
...
+    // some builtins (e.g. Symbol)?
+    if (!sym.declarations || sym.declarations.length === 0) return;
+    // A symbol declared in this file does not need to be imported.
+    if (sym.declarations.some(d => d.getSourceFile() === this.file)) return;
+    // Find an exported declaration.
+    // Because tsickle runs with the --declaration flag, all types referenced from exported types
+    // must be exported, too, so there must either be some declaration that is exported, or the
+    // symbol is actually a global declaration (declared in a script file, not a module).
+    const decl = sym.declarations.find(
+        d => hasModifierFlag(d, /* Export | Default */ ts.ModifierFlags.ExportDefault));
```
evmar: "ExportDefault" is weird here(?)

src/tsickle.ts
```diff
@@ -557,16 +561,57 @@ class Annotator extends ClosureRewriter {
...
+    if (!decl) return;
+    // Exclude symbols declared in `declare global {...}` blocks, they are global and don't need
+    // imports.
+    let current: ts.Node|undefined = decl;
+    while (current) {
+      if (current.flags & ts.NodeFlags.GlobalAugmentation) return;
```
evmar: Is it possible there's some other declaration that satisfies your requirements? (E.g. that .find picked the wrong one above?) Might be worth refactoring so you call .some() over all declarations and do this filtering within the .some?

src/source_map_utils.ts
mprobst: I had the same feeling, and looked into what it'd take. The problem is twofold: we'd need to pipe this "extra prefix" through several layers of code, back up to `transformer_sourcemap`. That's a bit nasty, but more importantly, we still need to make sure the effective source map is correct. If we prepended the text in the transformer before parsing, we'd still need to handle the offset as we do now. If we hand the prefix further up, then we'll break source maps because we prepend after those are generated.

I think folding this into the source mapper is OK, it's the abstraction we use to bridge from the text format used in annotate to the transformer mode, so it's the right place.

src/tsickle.ts

src/tsickle.ts

PR #666, opened 2017-12-05
Separate forward declares with a line break.

mprobst:
Previously, if a list of imports would end in a trailing force-load of a
type only module, the line comment would effectively strip the next line
of code. This is very rare (people separate their imports and code), but
gets triggered when forward declaring a symbol from a type only module
which gets prepended before all other imports.


mprobst:
@googlebot I'm myself, and I herewith confirm that I'm OK with me publishing commits I made myself. Maybe you're a bit confused about who's who?

mprobst:
@evmar thanks for the explanation, I was wondering why this was. But I think indeed this is safe, the forward declares do not replace the emit of any other node, so there should not be issues with preceding trivia.

Also, we should really fully move to transformers to escape these kinds of lexical issues...

PR #667, opened 2017-12-06
Upgrade bazel rules to latest

alexeagle:


evmar:
Is there any change worth noting here?

alexeagle:
I opened the PR as a way to do manual CI.

It turns out enabling strict deps breaks tsickle, so I know I shouldn't publish a release from current rules_typescript.

alexeagle:
Ready to review.

Note, I removed the prepublishonly warning in package.json because it was broken - we copy the whole package.json into the `built` folder so the `publish.sh` script fails. I noticed it last time I pushed tsickle to npm.

PR #669, opened 2017-12-07
 Emit goog.module & co after @fileoverview.

mprobst:
This fixes some oddity in our emit, and also makes it possible for
downstream tooling to find fileoverview comments more easily.

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -18,6 +18,17 @@ const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
     new Set(['fileoverview', 'externs', 'modName', 'mods', 'pintomodule']);
 
 /**
+ * Returns true if the given comment is a \@fileoverview style comment in the Closure sense, i.e. a
+ * comment that has JSDoc tags marking it as a fileoverview comment.
+ * Note that this is different from TypeScript's understanding of the concept, where a file comment
+ * is a comment separated from the rest of the file by a double newline.
+ */
+export function isClosureFileoverviewComment(text: string) {
+  const current = jsdoc.parseContents(text);
+  return current !== null && current.tags.some(t => FILEOVERVIEW_COMMENT_MARKERS.has(t.tagName));
+}
+
+/**
  * A transformer that ensures the emitted JS file has an \@fileoverview comment that contains an
  * \@suppress {checkTypes} annotation by either adding or updating an existing comment.
  */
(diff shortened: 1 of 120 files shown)


mprobst:
@alexeagle can you take a look while Evan is out? I've tried to split this in two commits, one with the actual change, and then one with a lot of noise from tests.

src/fileoverview_comment_transformer.ts
```diff
@@ -17,6 +17,17 @@ import * as ts from './typescript';
 const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
     new Set(['fileoverview', 'externs', 'modName', 'mods', 'pintomodule']);
 
+/**
+ * Returns true if the given comment is a \@fileoverview style comment in the Closure sense, i.e. a
+ * comment that has JSDoc tags marking it as a fileoverview comment.
+ * Note that this is different from TypeScript's understanding of the concept, where a file comment
+ * is a comment separated from the rest of the file by a double newline.
+ */
+export function isFileoverviewComment(text: string) {
```
alexeagle: maybe name it isClosureFileoverviewComment?

mprobst:
> does this PR have an effect on the requirement to have a blank line after the /** @fileoverview */ ?

No - that's a TypeScript thing, and this one is dealing in Closure fileoverview comments, only in JS after all TS has already happened.

photz:
Now when I run the compiler, I always get this error:
`ERROR - goog.module() call must be the first statement in a module.`

mprobst:
@photz can you share the generated `.js` code that Closure compains on?

photz:
@mprobst 
```
var __assign = (this && this.__assign) || Object.assign || function(t) {
    for (var s, i = 1, n = arguments.length; i < n; i++) {
        s = arguments[i];
        for (var p in s) if (Object.prototype.hasOwnProperty.call(s, p))
            t[p] = s[p];
    }
    return t;
};
/**
 */
goog.module('mouse');
var module = module || { id: 'mouse.ts' };
exports.__esModule = true;
var tsickle_forward_declare_1 = goog.forwardDeclare("update");
goog.require("update"); // force type-only module to be loaded
```
The verson string returned by the compiler is 20180610.0.0.

mprobst:
You have to run tsc with --importHelpers (or --noEmitHelpers and then make
the helpers available globally). Your app will then import the tslib npm
package for functions like __assign.

Am Mo., 18. Juni 2018 um 22:05 Uhr schrieb photz <notifications@github.com>:

PR #670, opened 2017-12-07
Don't emit clutz aliases for reexported symbols.

LucasSloan:
Tsickle should only emit clutz aliases in the original exporting file, not any files that reexport.  Reexporting a symbol doesn't bring the symbol into the file, so the alias would refer to a symbol that isn't available.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -2130,6 +2130,23 @@ export function emitWithTsickle(
   };
 }
 
+function areAnyDeclarationsFromSourceFile(
+    declarations: ts.Declaration[], sourceFile: ts.SourceFile) {
+  for (const decl of declarations) {
+    if (decl.getSourceFile() === sourceFile) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
+/**
+ * A tsickle produced declaration file might be consumed be referenced by Clutz
+ * produced .d.ts files, which use symbol names based on Closure's internal
+ * naming conventions, so we need to provide aliases for all the exported symbols
+ * in the Clutz naming convention.
+ */
 function addClutzAliases(
     fileName: string, dtsFileContent: string, sourceFiles: ts.SourceFile[],
     typeChecker: ts.TypeChecker, host: TsickleHost): string {
@@ -2150,6 +2167,14 @@ function addClutzAliases(
     // moduleExports is a ts.Map<ts.Symbol> which is an es6 Map, but has a
     // different type for no reason
     for (const symbol of moduleExports) {
+      // We only want to add clutz aliases in the file the symbol was originally
+      // exported from, not in any files where the symbol was reexported, since
+      // the alias will refer to a symbol that might not be present in the reexporting
+      // file.  If there are no declarations, be conservative and emit the aliases.
+      const declarations = symbol.getDeclarations();
+      if (declarations && !areAnyDeclarationsFromSourceFile(declarations, sf)) {
+        continue;
+      }
       // Want to alias the symbol to match what clutz would produce, so clutz .d.ts's
       // can reference symbols from typescript .d.ts's. See examples at:
       // https://github.com/angular/clutz/tree/master/src/test/java/com/google/javascript/clutz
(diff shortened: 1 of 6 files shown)


mprobst:
@LucasSloan is this PR no longer necessary? What happened? :-)

LucasSloan:
Whoops, I meant to close the other PR, since it didn't pass tests and needed a major rethink.  Please review!

src/tsickle.ts
```diff
@@ -2130,6 +2130,23 @@ export function emitWithTsickle(
   };
 }
 
+function isSymbolDeclaredInSourceFile(symbol: ts.Symbol, sourceFile: ts.SourceFile) {
+  const declarations = symbol.getDeclarations();
+  if (!declarations) {
+    // If there are no declarations, err on the side of safety and emit the clutz
+    // aliases
```
evmar: This comment is a bit weird because it comes out of nowhere (it looks like the function is just checking where a symbol lives, but now it's talking about clutz).  Maybe a better comment is a function doc that says "if there are no declarations, returns true" without talking about clutz and then documenting the clutz thing at the usage site?  Not sure.

src/tsickle.ts
```diff
@@ -2130,6 +2130,23 @@ export function emitWithTsickle(
...
+  for (const decl of declarations) {
+    if (decl.getSourceFile() === sourceFile) {
+      return true;
+    }
+  }
+
+  return false;
+}
+
 function addClutzAliases(
```
evmar: I have no idea what this function is supposed to do, heh.

src/tsickle.ts
evmar: Another idea: make this function take the declaration list rather than the symbol itself, and have the caller pre-filter out symbols that have no declaration.

test_files/reexport.declaration/reexport.ts
```diff
@@ -0,0 +1 @@
+export * from './export';
```
evmar: Can you add some comments to these test files about what they're testing, like what property of the output is important?

test/test_support.ts
```diff
@@ -120,6 +120,12 @@ export function createSourceCachingHost(
     if (sources.has(fileName)) {
       return true;
     }
+    // Typescript occasionally needs to look on disk for files we don't pass into
+    // the program as a source (eg to resolve a module that's in node_modules),
+    // but only .ts files explicitly passed in should be findable
+    if (/.*\.ts$/.test(fileName)) {
+      return false;
+    }
```
evmar: Why is this necessary in this change?

src/tsickle.ts
LucasSloan: Good suggestion, that was bugging me!

src/tsickle.ts
LucasSloan: Added a comment - does it make sense?

test/test_support.ts
LucasSloan: Test support provides a kind of hermetic way to run a typescript compilation, where you explicitly pass in the sources as strings.  However, we have one test that needs to check if a file exists on disk to do module resolution inside node_modules (even though it doesn't actually read the contents of the file).

I added a test case where I wanted to compile 2 .d.ts's to check them for consistency, but without this line, tsc would notice that there was a .ts file around and try to grab that instead of the .d.ts file I wanted to compile.

This was the easiest way I could find to resolve the 2 use cases.

test_files/reexport.declaration/reexport.ts
LucasSloan: Added a comment.

test/test_support.ts
```diff
@@ -120,6 +120,12 @@ export function createSourceCachingHost(
     if (sources.has(fileName)) {
       return true;
     }
+    // Typescript occasionally needs to look on disk for files we don't pass into
+    // the program as a source (eg to resolve a module that's in node_modules),
+    // but only .ts files explicitly passed in should be findable
+    if (/.*\.ts$/.test(fileName)) {
```
evmar: Can write it:
  /\.ts$/

LucasSloan: Great, thanks for the tip!

PR #671, opened 2017-12-08
Emit clutz aliases in the style that incremental clutz expects.

LucasSloan:
Global clutz runs allow symbols to be resolved to the name they have in their original contexts, even if they're exported under another name.  Incremental clutz is forced to use the exporting name, so tsickle needs to emit both until global clutz goes away.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -2141,6 +2141,15 @@ function areAnyDeclarationsFromSourceFile(
   return false;
 }
 
+function addToMultiMap<T, U>(map: Map<T, U[]>, key: T, value: U) {
+  const array = map.get(key);
+  if (array) {
+    array.push(value);
+  } else {
+    map.set(key, [value]);
+  }
+}
+
 /**
  * A tsickle produced declaration file might be consumed be referenced by Clutz
  * produced .d.ts files, which use symbol names based on Closure's internal
@@ -2150,7 +2159,7 @@ function areAnyDeclarationsFromSourceFile(
 function addClutzAliases(
     fileName: string, dtsFileContent: string, sourceFiles: ts.SourceFile[],
     typeChecker: ts.TypeChecker, host: TsickleHost): string {
-  const reexports: string[] = [];
+  const reexportsByNamespace: Map<string, string[]> = new Map();
   for (const sf of sourceFiles) {
     const moduleSymbol = typeChecker.getSymbolAtLocation(sf);
     const moduleExports = moduleSymbol && typeChecker.getExportsOfModule(moduleSymbol);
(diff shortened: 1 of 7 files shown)


src/tsickle.ts
```diff
@@ -2153,26 +2153,39 @@ function addClutzAliases(
...
 
       if (symbol.flags & ts.SymbolFlags.Class) {
         // classes need special care to match clutz, which seperates class types into a
         // type for the static properties and a type for the instance properties
-        reexports.push(`type ${clutzSymbolName}${paramsWithContraint} = ${symbol.name}${params};`);
-        reexports.push(`const ${clutzSymbolName}: typeof ${symbol.name};`);
-        reexports.push(
-            `type ${clutzSymbolName}_Instance${paramsWithContraint} = ${symbol.name}${params};`);
-        reexports.push(`const ${clutzSymbolName}_Instance: typeof ${symbol.name};`);
+        for (const clutzSymbolName of clutzSymbolNames) {
```
mprobst: these loops are a bit odd for N=2 (and never more). Consider (optional) extracting the 4 push statements here into a function declare in this scope that you call twice with the two symbols, and for the code below just duplicating the `reexports.push` line. Roughly the same amount of code, but arguably more clear than having a loop for two items.

LucasSloan:
Did some major surgery to get the desired output.  PTAL.

src/tsickle.ts
```diff
@@ -2153,37 +2162,62 @@ function addClutzAliases(
...
       // can reference symbols from typescript .d.ts's. See examples at:
       // https://github.com/angular/clutz/tree/master/src/test/java/com/google/javascript/clutz
-      const clutzSymbolName = 'module$contents$' + clutzModuleName + '_' + symbol.name;
-
+      // The first symbol name is that currently produced by clutz, and the second
+      // is what incremental clutz will produce.
+      const reexportData = [];
+      reexportData.push({
+        namespace: 'ಠ_ಠ.clutz',
+        clutzSymbolName: 'module$contents$' + clutzModuleName + '_' + symbol.name,
```
evmar: Maybe use
```
`foo${clutzModuleName}bar_${...}`
```
style strings?

src/tsickle.ts
```diff
@@ -2153,37 +2162,62 @@ function addClutzAliases(
...
+        for (const {namespace, clutzSymbolName, aliasedSymbolName} of reexportData) {
+          addToMultiMap(
+              reexports, namespace, `const ${clutzSymbolName}: typeof ${aliasedSymbolName};`);
+        }
       }
     }
   }
 
-  if (reexports.length) {
+  if (reexports.size) {
```
evmar: I think .length was correct.

src/tsickle.ts
LucasSloan: Done.

src/tsickle.ts
LucasSloan: `reexports` is a map, not a list.  I think your confusion come from `reexports` and `reexportsData` both existing.  I could rename them to `namespaceToReexportLineMultimap` and `reexportsDataList`, but that seems terrible.  Do you have any thoughts on improving the names?

evmar: I might give the inner one the shorter name and the outer one the longer name.  Like maybe reexports for the inner one, and allReexports or reexportsBySymbol for the outer one.  ("Data" is usually a filler word so you can leave it out.)

LucasSloan: Renamed the outer one to `reexportsByNamespace` and the inner on to just `reexports`.

LucasSloan:
Anything else?

PR #672, opened 2017-12-14
Upgrade tsickle to TypeScript 2.6

mprobst:
See the individual commits for more details. The major bit here is how we handle imports and exports.


mprobst:
Admittedly the `options` and `host` fields in the flags are named suboptimal ;-) Taking ideas, maybe I should just say `compilerOptions` and `moduleResolutionHost`? OTOH this actually runs out to line up with the fields that `CompilerHost` in rules_typescript already defines.

src/typescript.ts
```diff
@@ -20,7 +20,7 @@ import * as ts from 'typescript';
 // Note, this import depends on a genrule copying the .d.ts file to this package
 import * as ts24 from './typescript-2.4';
```
alexeagle: I think we now need to add a test case for ts25 as well

src/es5processor.ts
```diff
@@ -119,6 +149,7 @@ class ES5Processor extends Rewriter {
 
   /**
    * Emits file comments for the current source file, if any.
+   * oddityasds
```
rkirov: oddity indeed

src/transformer_sourcemap.ts
```diff
@@ -49,10 +49,33 @@ export function createTransformerFromSourceMap(
...
+        if (ts.isExportDeclaration(node)) {
+          // Return the original nodes for export declarations, unless they were expanded from an
+          // export * to specific exported symbols.
+          const originalExport = originalNode as ts.ExportDeclaration;
+          if (!node.moduleSpecifier) {
+            // export {a, b, c};
+            return originalNode;
+          }
+          if (!!originalExport.exportClause === !!node.exportClause) {
+            // This alread was export with symbols (export {...}) or was not expanded.
```
rkirov: already

src/typescript.ts
mprobst: Done.

src/transformer_sourcemap.ts
mprobst: Done.

src/es5processor.ts
mprobst: Fixed.

Issue #674, opened 2017-12-15
Tsickle decorator downl-eveling removes all decorators


chuckjaz:
When any decorator is down-leveled, all decorators are removed instead of just the ones that were down-leveled.

See https://github.com/angular/angular/issues/20864

To reproduce see https://github.com/alxhub/angular/commit/b4835e332998542ea594ccb776c92615640594a1

The issue appears to be here: https://github.com/angular/tsickle/blob/f3860a588361ca8cae2727e4691c0cd27a946a02/src/class_decorator_downlevel_transformer.ts#L138

This unconditionally removes all decorators from the class instead of just removing the decorator that was down-leveled.

PR #675, opened 2017-12-15
Preserve decorators that are not down-leveled

chuckjaz:
Fixes: #674 

Preserves decorators that are not down-leveled. This fixes https://github.com/angular/angular/issues/20864


test/decorator-annotator_test.ts
```diff
@@ -442,6 +442,21 @@ class ViewUtils {
...
+        /** @Annotation */ var RemoveMe: Function = undefined as any;
+
+        var KeepMe: Function = undefined as any;
+
+        @KeepMe()
+        @RemoveMe()
+        class ViewUtils {
+          constructor() {}
+        }
+        `).to.contain('@KeepMe');
```
LucasSloan: Please follow the pattern in the rest of the file and compare to the actual output.

PR #678, opened 2017-12-20
emit typedefs in untyped mode

evmar:
Even in untyped mode, it's useful to define the typedef as {?}.
This allows downstream JS to continue to refer to the typedef.


evmar:
[note to self: this fails in g3 for reasons I need to track down]

evmar:
@mprobst I think I've gotten all the google3-side breakages from this fixed.  Do you think it's worth landing?  It is a small step towards tsickle typed.

mprobst:
Yes, go for it.

PR #680, opened 2017-12-21
don't emit types like "...?=" in externs, which are illegal

evmar:
If we think a param is both a rest param and optional, prefer the
rest param (as that subsumes optional).


evmar:
In particular, I expect this to fix the common warning we have with jquery.

evmar:
Green in google3 for a sample run (think this is pretty low-risk, given that the old output was illegal).

PR #682, opened 2017-12-21
avoid a warning when 'module' is unused

evmar:
Depending on the user's compilation settings, not using the implied
'module' variable will cause compilation errors.  Avoid this.


rkirov:
Good luck on the global presubmit

evmar:
Green in g3 sample.

Issue #683, opened 2017-12-22
[Feature] Option to disable automatic property quoting


theRobinator:
I'm working on transitioning a large codebase that uses Closure Compiler+Templates over to Typescript. Currently, we have code that looks like this:
```
/** @type {!Object<string, number>} */
let params = {};
if (something) {
  params.foo = 5;
}

soy.renderAsElement(SomeTemplate, params);
```

The Typescript version of this is to declare `params` as `{[key: string]: number}`, which works well. However, tsickle emits a warning on seeing this:
```
{ [key: string]: number; } has a string index type but is accessed using dotted access. Quoting the access.
```
This changes the setting code into `params['foo'] = 5`, which no longer works because any usages of `params.foo` in soy templates are minified!

To address this situation, I propose adding a `--disableAutoQuoting` option to tsickle. This would disable the transformation above, making code that currently compiles under CC continue working.

I will submit a PR for this in just a bit.

evmar:
Do you have other code where an `Object<string, ...>` where it *should* quote?  How do you represent a map of strings?

theRobinator:
We have code where we don't want to minify object properties (passing to external libs, etc), and in those situations, we have quoted the object keys ourselves. We're currently compiling the whole thing in advanced mode.

evmar:
Have you considered using types like
```
/** @type {{foo: number}} */
```
?

(Not trying to be difficult here, just making sure I understand the design space!)

theRobinator:
We do have several places where we've done that (mostly with `@typedef`s), and that solves the soy template case above pretty well. It doesn't get everything, however, like this method for global soy parameters:

```
function addGlobalParameters(user, soyParams) {
  soyParams.global_userName = user.name;
  .. etc
}
```

Actually it was upon discovering this function that I decided to submit the PR. It seemed like I was playing whack-a-mole with bugs that the quoting caused, and that it would be a better solution to optionally disable the quoting. I'm especially worried because CC doesn't ever emit errors about bad quoting, and since our codebase is ~500K lines it would take forever to find something.

mprobst:
Sorry for the radio silence, I added this and have been on vacation the
last three weeks. I'll get back to you on this tomorrow.

theRobinator:
This is fixed by https://github.com/angular/tsickle/pull/684.

PR #684, opened 2017-12-22
Add disableAutoQuoting option to stop quoting of string object properties

theRobinator:
See https://github.com/angular/tsickle/issues/683 for motivation.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -72,6 +72,7 @@ function loadSettingsFromArgs(args: string[]): {settings: Settings, tscArgs: str
         break;
       case 'disableAutoQuoting':
         settings.disableAutoQuoting = true;
+        break;
       case '_':
         // This is part of the minimist API, and holds args after the '--'.
         break;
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -66,6 +70,8 @@ function loadSettingsFromArgs(args: string[]): {settings: Settings, tscArgs: str
       case 'verbose':
         settings.verbose = true;
         break;
+      case 'disableAutoQuoting':
+        settings.disableAutoQuoting = true;
```
mprobst: don't forget the break!

theRobinator: Whoops! Good catch. Done.

mprobst:
@theRobinator could you rebase the CL on `origin/master` and force push? That should get rid of the commit by me in the history and make the CLA bot happy.

```
git rebase origin/master
git push -f
```

theRobinator:
Will do.

theRobinator:
@mprobst done!

mprobst:
merged.

alexeagle:
Note, we now think the option was reversed - autoQuoting should be opt-in, not opt-out (at least until it works almost all the time)

https://github.com/angular/tsickle/pull/895 will switch the logic to `enableAutoQuoting` instead.

PR #685, opened 2017-12-30
Emit goog.require only for imported types.

mprobst:
Emitting forced imports can cause circular dependency issues in Closure
Compiler. Types that are only imported because they are used in resolved
union types must have already been imported by some other file in the
compilation unit, so skipping these avoids some circularity issues
without dropping necessary files from the compilation.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -625,7 +625,8 @@ class Annotator extends ClosureRewriter {
     // Already imported?
     if (this.forwardDeclaredModules.has(moduleSymbol)) return;
     // TODO(martinprobst): this should possibly use fileNameToModuleId.
-    const text = this.getForwardDeclareText(sf.fileName, moduleSymbol);
+    const text =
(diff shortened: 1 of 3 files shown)


src/tsickle.ts
```diff
@@ -1136,15 +1137,18 @@ class Annotator extends ClosureRewriter {
         {options: this.tsOpts, host: this.tsHost}, this.file.fileName,
         (specifier as ts.StringLiteral).text);
     const moduleSymbol = this.typeChecker.getSymbolAtLocation(specifier);
-    this.emit(this.getForwardDeclareText(importPath, moduleSymbol, isDefaultImport));
+    this.emit(this.getForwardDeclareText(
+        importPath, moduleSymbol, /*forceTypeOnlyLoads*/ true, isDefaultImport));
```
evmar: Super minor nit: you use different whitespace style in this comment vs the previous one.

mprobst:
Turns out this actually breaks a ton of stuff and doesn't fix my issue (but on the bright side, Jeremy promised to fix my issue!).

PR #704, opened 2018-01-09
chore(renovate): rebase stale PRs

rarkins:
Configures Renovate to automatically rebase its PRs whenever they are not up-to-date with master.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,6 +1,7 @@
 {
   "extends": [
-    "config:base"
+    "config:base",
+    ":rebaseStalePrs"
   ],
   "ignorePaths": ["test*/**"]
 }


rarkins:
@alexeagle 

Renovate normally can detect if a repository requires updates on PRs, i.e. PRs must be up-to-date with `master`. However for this "Forking Renovate" version I chose to be cautious and not ask for the necessary read-only rights to "administration" data about repositories, which means Forking Renovate cannot automatically detect the repository setting and needs to be configured. BTW I pointed out to GitHub that it's a bit silly that the setting is visible via the open website yet requires special rights to see via API, but they did not consider it to be a bug.

Merging this PR will mean you should no longer need to click the "Update branch" button on remaining Renovate PRs every time you merge another one. Renovate will run every time you merge one of its PRs, so you should see them updated within seconds/minutes (whatever the time takes to regenerate all those lock files). For *other* commits to master (non-Renovate PRs), Renovate runs hourly to catch up with those and rebase.

alexeagle:
I'm not sure about this solution, it guarantees quadratic CI resource consumption. Let's keep chatting on the other PR first

alexeagle:
I think it's easier to understand if renovate acts like other PR authors, and doesn't re-base its PRs. Should be less of a problem now that we've on-boarded.

Issue #714, opened 2018-01-19
Some enum files are skipped


theRobinator:
I'm working on migrating a large project that has a _lot_ of enums. For some reason I'm seeing that tsickle is only producing the correct output in some cases.

As an example, I have two enums, `ApiCacheStrategy` and `AnimationState`, both in the same folder (but separate files):
```
// apicachestrategy.ts
export enum ApiCacheStrategy {
    ENDPOINT,
    PARAMS,
}

// animationstate.ts
export enum AnimationState {
    PLAYING,
    PAUSED,
    STOPPED,
}
```

When I run tsickle, `animationstate.ts` is translated properly:
```
/** @enum {number} */
const AnimationState = {
    PLAYING: 0,
    PAUSED: 1,
    STOPPED: 2,
};
exports.AnimationState = AnimationState;
AnimationState[AnimationState.PLAYING] = "PLAYING";
AnimationState[AnimationState.PAUSED] = "PAUSED";
AnimationState[AnimationState.STOPPED] = "STOPPED";
```
But, `apicachestrategy.ts` gets no love, seemingly being skipped:
```
var ApiCacheStrategy;
(function (ApiCacheStrategy) {
    ApiCacheStrategy[ApiCacheStrategy["ENDPOINT"] = 0] = "ENDPOINT";
    ApiCacheStrategy[ApiCacheStrategy["PARAMS"] = 1] = "PARAMS";
})(ApiCacheStrategy = exports.ApiCacheStrategy || (exports.ApiCacheStrategy = {}));
```

I started trying to make a minimal repro case to show off in this issue, but running tsickle with only `apicachestrategy.ts` actually worked fine! So then I went after my tsconfig.json, assuming it was some weird `@export` collision issue, and found out that changing this:
```
"include": [
    "project/src/**/*"
]
```
to this:
```
"include": [
    "project/src/every/**/*",
    "project/src/subfolder/**/*",
    "project/src/under/**/*",
    "project/src/src/**/*",
    ...
]
```
also fixes it, though the two globs _should_ evaluate to the same list of files!

So I'm pretty sure that this issue is very specific to my project layout and tsconfig. I'm going to start debugging this in tsickle tomorrow morning, but in the meantime I'd appreciate any input or ideas that anybody has, especially around where in the code to start looking.

theRobinator:
Figured it out! The glob change changed how the path was supplied to the `shouldSkipTsickleProcessing`, and that function was always comparing against a relative path. I've submitted a PR.

PR #715, opened 2018-01-19
Resolve all paths before determining whether or not to skip files

theRobinator:
Fixes https://github.com/angular/tsickle/issues/714

It turns out that Typescript supplies a mixture of relative and absolute paths to the `shouldSkipTsickleProcessing` function. In particular, files that are imported in TS using `../` appear using absolute paths. However, the `fileNames` array ends up being all relative paths (at least in my project).

As far as I can tell we can't guarantee any of the path formats, so this change resolves all paths to their absolute versions before checking if a file should be excluded.


theRobinator:
Also, I noticed while doing this that the golden tests are failing in master. I didn't introduce any failing tests :)

mprobst:
Thanks for the fix!

PR #717, opened 2018-01-23
Delete package-lock.json.

mprobst:
The file is unused and confusing, given there is also the yarn lockfile.


mprobst:
This supersedes #712 (sorry, I made that non-mergeable due to conflict).

PR #719, opened 2018-01-23
Name goog.modules after absolute paths

theRobinator:
As mentioned in https://github.com/angular/tsickle/pull/715, file paths are sometimes supplied to tsickle as relative, and sometimes as absolute. I originally thought that this only affected file ignoring, but it apparently affects module naming as well. Specifically, it is possible for tsickle to emit `goog.module` and `goog.require` declarations that don't match.

To fix this, I've modified the `pathToModuleName` function to always emit absolute path module names. This allows my codebase to compile.


src/cli_support.ts
```diff
@@ -10,13 +10,17 @@ import * as path from 'path';
...
+  fileName = fileName.replace(/\.[tj]s$/, '');
 
   if (fileName[0] === '.') {
     // './foo' or '../foo'.
     // Resolve the path against the dirname of the current module.
     fileName = path.join(path.dirname(context), fileName);
   }
+
+  // Ensure consistency by naming all modules after their absolute paths
+  fileName = path.resolve(fileName);
```
mprobst: This is indeed a problem, and your fix is pretty close I think.

We cannot quite use absolute path names I think, because then you'll end up with non-reproducible names across different developer machines, build systems, and so on.

Maybe you could try to first create an absolute path, and then relativize against the project's source roots?

theRobinator: Sorry for the delay on this -- I've been sidetracked by other projects. This sounds like a better idea, and as a bonus all the module names will be easier to read. I'll make this change either tomorrow or Monday.

theRobinator: @mprobst I've updated this. I had to make it a bit more complex than just resolving against the `sourceRoot` property since it's possible to specify files on the command line.

mprobst:
Friendly ping, is this still current? :-)

theRobinator:
Whoops, I totally forgot that this hadn't been merged. I'm sorry about that! I just updated this with a refactor to pass the property in.

I'm having some issues making the tests do the same thing on my machine as they do in Travis. The master tests are failing for me, including after a `bazel clean`, but the failures are different than the ones that show up here on a push. Does anyone have any ideas about that?

In the meantime I've tested this by running the built version on my codebase and running those unit tests, which pass.

theRobinator:
I nuked my dependencies and reinstalled all of them and it works now. Guess something updated how it indents the code. @mprobst this is now all ready!

src/main.ts
```diff
@@ -131,18 +131,21 @@ function loadTscConfig(args: string[]):
...
+
+  const compilerHost = ts.createCompilerHost(options);
+  const program = ts.createProgram(fileNames, options, compilerHost);
+  const filesToProcess = new Set(absoluteFileNames);
   const transformerHost: tsickle.TsickleHost = {
     shouldSkipTsickleProcessing: (fileName: string) => {
       return !filesToProcess.has(path.resolve(fileName));
     },
     shouldIgnoreWarningsForPath: (fileName: string) => false,
     pathToModuleName: cliSupport.pathToModuleName,
```
mprobst: what about binding the parameter here, so we don't have to pipe rootModulePath through all the system?

```
const rootModulePath = getCommon...;
// ...
  pathToModuleName: (context: string, importPath: string) => cliSupport.pathToModuleName(context, importPath, rootModulePath),
```

That saves us a lot of piping things around, but achieves the same thing I think?

src/util.ts
```diff
@@ -62,3 +63,24 @@ export function hasModifierFlag(node: ts.Node, flag: ts.ModifierFlags): boolean
...
+export function getCommonParentDirectory(fileNames: string[]): string {
+  const pathSplitter = /[\/\\]+/;
+  const commonParent = fileNames[0].split(pathSplitter);
+  for (let i = 1; i < fileNames.length; i++) {
+    const thisPath = fileNames[i].split(pathSplitter);
+    let j = 0;
+    while (thisPath[j] === commonParent[j]) {
+      j++;
+    }
+    commonParent.splice(j, Infinity);  // Truncate without copying the array
```
mprobst: I think you can just do `commonParent.length = j;`

src/main.ts
```diff
@@ -131,18 +131,21 @@ function loadTscConfig(args: string[]):
...
     fileNameToModuleId: (fileName) => fileName,
+    rootModulePath: getCommonParentDirectory(absoluteFileNames),
```
mprobst: can you use `options.rootDir || getCommon...`? `rootDir` is the root of the source paths to generate, so it matches this well and should be preferred if the user sets it.

mprobst:
@theRobinator sorry about the dependencies issue. You do need to run yarn manually despite bazel, and if you hadn't done that or got into an odd state it's possible to have an unclean build still. It's a bit of a compromise between bazel (for reproducible builds) and using npm (because that fits the node ecosystem).

mprobst:
@googlebot I confirm that this code is fine, it's simply a merge from master to resolve conflicts.

mprobst:
Thanks for the contribution @theRobinator !

PR #727, opened 2018-01-29
feat: emit template types even in untyped mode.

mprobst:
For historical reasons, Closure Compiler allows specifying template
arguments even on types that do not declare any template arguments, and
just silently ignores them, including the types inside of them.

Emitting template arguments even when run in untyped mode allows Closure
Compiler to migrate to enforcing type argument's presence, and makes
sure type arguments are resolved when using templated TS types in
untyped mode from Closure code.


mprobst:
@blickly care to take a look/

PR #730, opened 2018-01-30
Configure renovate to be less noisy:

alexeagle:
- For all dependencies other than Closure compiler:
- update only once a month
- combine all the updates into a single PR

/cc @rarkins to see if I'm doing this right

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,13 @@
   "extends": [
     "config:base"
   ],
+  "packageRules": [
+    {
+      "extends": ["group:all"],
+      "excludePackagePatterns": ["google-closure-compiler"],
+      "groupName": "deps",
+      "schedule": ["on the first day of the month"]
+    }
+  ],
   "ignorePaths": ["test*/**"]
 }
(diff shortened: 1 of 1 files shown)


rarkins:
@alexeagle I'll get back to you today

alexeagle:
thanks - I didn't see any negative patterns like this in the docs and I don't know how to verify what will be the effect of this change (so it's likely just wrong)

rarkins:
@alexeagle I updated Renovate's logic so that a purely "negating" `packageRule` like you wrote is now valid. I think that's intuitive and others would have the same expectation. I recommend to update it slightly to this though:

```json
{
  "extends": [
    "config:base",
    ":semanticCommitsDisabled"
  ],
  "packageRules": [
    {
      "excludePackageNames": ["google-closure-compiler"],
      "extends": [
        "group:all",
        "schedule:monthly"
      ]
    }
  ],
  "ignorePaths": ["test*/**"]
}
```

This way your monthly PR will be called "Update all packages".

alexeagle:
FYI @mprobst it's possible we'll just get a monthly red PR, and need to dial this back a bit.

PR #734, opened 2018-01-30
Fix crash when class declaration extends an expression.

### LINKED ISSUE #679: Crash when class extends <expression>
tsickle crashes if the RHS of a `class extends` is an expression instead of a symbol. This pattern is useful when declaring classes built up using class mixin functions as described here: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html

```ts
declare function MyMixin<T extends new (...args: any[]) => {}>(base: T):
    T & MixinConstructor;

interface MixinConstructor {
  new(...args: any[]): MyMixin;
}

interface MyMixin {
  mixinProp: string;
}

declare class MyClass extends MyMixin(HTMLElement) {
  classProp: string;
}
```

```
$ tsickle --externs=tsickle.js
Warning at mixin.d.ts: unhandled type flags: Intersection
/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1700
                        finally { if (e_18) throw e_18.error; }
                                            ^

Error: no symbol
    at ExternsWriter.ClosureRewriter.mustGetSymbolAtLocation (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:202:23)
    at ExternsWriter.ClosureRewriter.maybeAddHeritageClauses (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:488:44)
    at ExternsWriter.writeExternsType (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1861:22)
    at ExternsWriter.visit (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1788:26)
    at ExternsWriter.visit (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1692:34)
    at /Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1669:75
    at Array.forEach (<anonymous>)
    at ExternsWriter.process (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:1669:36)
    at writeExterns (/Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:2071:59)
    at /Users/aomarks/.nvm/versions/node/v9.0.0/lib/node_modules/tsickle/src/tsickle.js:2184:26```

### PR DESCRIPTION
**Pull request by aomarks**
Fixes #679.

It's possible for a class declaration to extend an expression that is not a symbol, for example when a mixin function is used to build a base class, as in `declare MyClass extends MyMixin(MyBaseClass)`.

Handling this correctly is tricky. Closure also doesn't handle this `extends <expression>` syntax. We would probably need to generate an intermediate class declaration and extend that. For now, just omit the `extends` annotation.

### CODE REVIEW

File: src/tsickle.ts [hunk 1]
```diff
@@ -476,7 +476,18 @@ abstract class ClosureRewriter extends Rewriter {
         // But it's fine to translate TS "implements Class" into Closure
         // "@extends {Class}" because this is just a type hint.
         const typeChecker = this.typeChecker;
-        const sym = this.mustGetSymbolAtLocation(impl.expression);
+        const sym = this.typeChecker.getSymbolAtLocation(impl.expression);
+        if (!sym) {
+          // It's possible for a class declaration to extend an expression that
```
mprobst: `... that does not *have* a symbol`, right?

src/tsickle.ts
```diff
@@ -476,7 +476,18 @@ abstract class ClosureRewriter extends Rewriter {
...
+          // is not a symbol, for example when a mixin function is used to build
+          // a base class, as in `declare MyClass extends MyMixin(MyBaseClass)`.
+          //
+          // Handling this correctly is tricky. Closure also doesn't handle this
```
mprobst: Can you expand on how Closure doesn't handle the extends syntax? Does it crash, does it report errors?

src/tsickle.ts
```diff
@@ -476,7 +476,18 @@ abstract class ClosureRewriter extends Rewriter {
...
+          // `extends <expression>` syntax. We would probably need to generate
+          // an intermediate class declaration and extend that. For now, just
+          // omit the `extends` annotation.
+          continue;
```
mprobst: Can you emit a comment a la `// taze: cannot resolve extended supertype` to make debugging easier?

mprobst:
There's a bigger question, which is what Closure Compiler is going to do with code like this. That is, if Closure sees `foo.bar()` with `bar` being defined on a mixin into `foo`, I suspect it'll back off its optimizations, but that's not certain.

I really wish Polymer wouldn't use mixins, life would be a lot easier :-/

src/tsickle.ts
aomarks: Done.

src/tsickle.ts
aomarks: Done (it throws, added issue link).

src/tsickle.ts
aomarks: I added a call to `debugWarn` so ti will show up in the diagnostics. Looks like I could also/instead do `this.emit('* ...')` from inside `maybeAddHeritageClauses`, which would add a comment to the class jsdoc, but I wasn't sure if that was a good idea (what if the jsdoc comment hadn't started yet?). Let me know which you prefer, and if just emitting a string directly from that function is ok.

aomarks:
> There's a bigger question, which is what Closure Compiler is going to do with code like this. That is, if Closure sees foo.bar() with bar being defined on a mixin into foo, I suspect it'll back off its optimizations, but that's not certain.

Yep! Actually in the Closure-compilable Polymer 2 source we assign the base class to a variable first, annotate it with the interfaces it gets from the mixin (where those interfaces are generated by a script that does static analysis of the mixin function), and then extend that. Like this: https://github.com/Polymer/polymer/blob/master/lib/elements/array-selector.html#L351.

> I really wish Polymer wouldn't use mixins, life would be a lot easier :-/

It is kinda painful. TypeScript's support for mixins is really nice, but it has definitely hurt our ergonomics with Closure.

aomarks:
Thanks for the review! Addressed comments, PTAL.

src/tsickle.ts
mprobst: I think it'd be nice to leave a comment in the generated JS. If something goes wrong during compilation, it'll be very helpful to have some documentation of what happened in the code the user sees.

aomarks: Done, PTAL.

aomarks:
> Can you rebase onto master?

Done.

PR #737, opened 2018-02-05
 feat(modules): support `export =`.

mprobst:
TypeScript natively emits these as `module.exports = ...` (CommonJS style), where Closure expects a bare `exports =` assignment.


src/es5processor.ts
```diff
@@ -186,10 +187,18 @@ class ES5Processor extends Rewriter {
         }
         // Check for
         //   Object.defineProperty(exports, "__esModule", ...);
-        if (this.isEsModuleProperty(node as ts.ExpressionStatement)) {
+        if (this.isEsModuleProperty(exprStatement)) {
           this.emitCommentWithoutStatementBody(node);
           return;
         }
+        // Check for
+        //    module.exports = ...;
```
evmar: super nit: use a 2-space tab here and below

mprobst: Done.

PR #744, opened 2018-02-09
update to latest bazel rules

alexeagle:


evmar:
Any expected change?

alexeagle:
Nope, just making sure this is green as part of my usual release process - note I had to revert an interesting commit from rules_nodejs because it broke tsickle...
Will update this to point to tags before merging.

evmar:
Pre-rubber stamp on importing adjustments to build rules done in other repos ;)

PR #749, opened 2018-02-16
Switch publish instructions to npm_package rule

alexeagle:
Also clean up fields in the tsconfig.json that were triggering warnings from ts_library

Signed-off-by: Alex Eagle <alexeagle@google.com>

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,9 +2,6 @@
   "compilerOptions": {
     "experimentalDecorators": true,
     "module": "commonjs",
-    "declaration": true,
-    "noEmitOnError": true,
-    "target": "es5",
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -2,9 +2,6 @@
   "compilerOptions": {
     "experimentalDecorators": true,
     "module": "commonjs",
-    "declaration": true,
```
LucasSloan: Why did tsconfig change?

alexeagle: I mentioned it in the commit message

This is to avoid warnings from `ts_library`

alexeagle:
Yes, I already published `tsickle@0.27.1` with this, and I did a deep-diff against the prior output to ensure they're identical.

PR #751, opened 2018-02-19
Update source-map dependency.

### LINKED ISSUE #750: Update source-map dependency
Renovate wasn't able to do this because there was a bug upstream with the typings for source-map, see https://github.com/angular/tsickle/pull/707#issuecomment-356656843

Now I really need this fix, because that same typings bug affects TypeScript compilations, eg. breakage on Angular Devkit https://circleci.com/gh/angular/devkit/7333 /cc @filipesilva

However there are breaking changes so this is not trivial:
https://github.com/mozilla/source-map/compare/0.5.7...0.6.1

alexeagle:
This is needed to pick up the fix for https://github.com/mozilla/source-map/issues/273

Closes #750


src/source_map_utils.ts
```diff
@@ -6,9 +6,31 @@
...
 
-import {BasicSourceMapConsumer, RawSourceMap, SourceMapConsumer, SourceMapGenerator} from 'source-map';
+import {RawSourceMap, SourceMapConsumer, SourceMapGenerator} from 'source-map';
 import * as ts from './typescript';
 
+/**
+ * This interface was defined in @types/source-map but is absent from the typings
+ * distributed in the source-map package.
+ * Copied from https://unpkg.com/@types/source-map@0.5.2/index.d.ts
+ * see https://github.com/angular/tsickle/issues/750
```
mprobst: leave a TODO with instructions on how to fix eventually?

alexeagle: I don't think there's any action to take, probably not worth making a fix upstream for their typings? Maybe they prefer these APIs be undocumented. I'll open an issue at least...

alexeagle: Ah, I'm wrong, https://github.com/mozilla/source-map/commit/7c06ac83dd6d75e65f71727184a2d8630a15bf58#diff-7945f6bb445d956794564e098ef20bb3 adds the `toJSON` method among other things.
However that's only in the 0.7 release which adds more breaking changes, eg. the `SourceMapConsumer` constructor is now a sync (???)
So I think we should land this, I'll update the comments.

PR #754, opened 2018-02-20
Tell renovate not to bother with source-map

alexeagle:
The breaking change from 0.6 to 0.7 is too breaky for us.

/cc @rarkins


rarkins:
Lgtm

PR #757, opened 2018-03-03
Add option to disable goog.modules conversion

choumx:
Add `--disableGoogModule` option and setting to disable Closure module conversion.

Context: Our project uses Closure Compiler with ES6 modules. Closure should [normalize disparate modules during compilation](https://github.com/google/closure-compiler/wiki/JS-Modules) but this doesn't seem to work for us (maybe due to our custom CC runner).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -24,6 +24,9 @@ export interface Settings {
   /** If provided, do not modify quoting of property accesses. */
   disableAutoQuoting?: boolean;
 
+  /** If provided, do not convert to 'goog.module' Closure modules. */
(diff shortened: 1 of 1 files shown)


choumx:
Actually we don't need this PR if we programmatically configure a `TsickleHost` instance and pass `false` to the `googmodule` setting. Feel free to merge or close. 😄

evmar:
Alex can maybe comment on the situation with ES6 and Closure modules interop.

mprobst:
Closing this given you're not using it, and as a professional software engineer, I want as little software as possible in my projects ;-)

PR #759, opened 2018-03-07
fix(externs): add SymbolConstructor.

bowenni:
https://github.com/Microsoft/TypeScript/blob/v2.6.2/lib/lib.es2015.symbol.d.ts#L29


bowenni:
I'm working on the failing test, might need to add an `es3.js` extern to the e2e tests.

mprobst:
The e2e test should have all externs available, that's how it also has `Promise` and so on from this file. I believe the `isInTestMode` is a red herring here, we're running JSCompiler in a regular execution mode for our tests, and that flag is only for JSCompiler internal testing.

Are you maybe confusing `Symbol` (uppercase) and `symbol` lowercase? `Symbol` in `es3.js` is only a function, whereas `symbol` is the typedef. Maybe you just need to use `symbol` (lower case) in your CL?

Though if that's the case, I wonder what `Symbol` (upper) emits as, and what JSCompiler makes out of it. Could you consider adding a bit of TypeScript code that uses `Symbol` and `SymbolConstructor` so we can see this working in action?

mprobst:
Your definition:

```javascript
/** @typedef {function(new:Symbol)} */
var SymbolConstructor;
```

Defines `SymbolConstructor` as a function that returns `Symbol` when called with `new`. I think there is no type called `Symbol` in Closure, there's only a type called `symbol`. So you might need to change this to:

```javascript
// Note: lowerCase symbol below
/** @typedef {function(new:symbol)} */
var SymbolConstructor;
```

bowenni:
Now I see what you mean. 
I confused `Symbol` with `symbol`, and yes `symbol` is the right typedef.

Issue #761, opened 2018-03-12
Remove duplicate declarations of static methods


shicks:
Currently tsickle emits code like the following:

```js
class Observable {}
// ...
/** @owner Observable */
Observable.create = (subscribe) => { /* ... */ };
exports.Observable = Observable;
function Observable_tsickle_Closure_declarations() {
    /**
     * @owner Observable
     * @type {!Function}
     */
    Observable.create;
}
```

The Closure compiler doesn't understand this redeclaration properly, and it will soon be an error.

The extra function wrapper shouldn't be necessary.  The declarations need to be moved to be above an unannotated assignment.

evmar:
I took a tour through the relevant code.

I removed the extra function parameter here:
https://github.com/angular/tsickle/pull/762

Moving the location declaration is trickier.  tsickle works by modifying the input to the compiler, but the emit of `class X { }` and the following `X.foo = ...` for a static is done by the compiler after our input.  We might have to move statics out of the class ourselves, like a sort of pre-downlevelling.

mprobst:
I think the fix to remove the function wrapper will cause other problems; we introduced that to avoid issues with base classes that have a getter with the same name as a property that then gets executed when the browser sees the declaration.

But taking a step back, there are two problems here:

1. we implicitly declare `create` twice
2. the declaration's types do not match

The code's intention is actually not to declare twice, we only want one declaration to count, the one in the `declarations` method. The other assignment should really just be an assignment and not count as a declaration at all. Is it possible to tell that to Closure Compiler, shicks@?

Similarly for 2, our intention is for `Observable.create` to be of type `Function`. Closure Compiler infers a tighter type here, and then complains that it doesn't match with the declaration. I wonder if there's a way for it to have the human written type take precedence over its implied inference?

@evmar re ordering, and @shicks might be able to confirm, but would it be possible to move the `Observable_tsickle_Closure_declarations` _before_ the class declaration? ES6 classes are not hoisted, but we're not running the function, so pure JS wise, this should be legal:

```ts
function Observable_tsickle_Closure_declarations() { /** @type ... */ Observable.create; }
class Observable {}
```

evmar:
I tried the reordering you suggested in a playground and it worked.  I'll try a larger patch.

evmar:
Edit: I take that back, the earlier type declaration seems to be ignored.  I'll experiment some more.

shicks:
[...]

PR #762, opened 2018-03-12
don't wrap a function around the declarations of class members

evmar:
I think the use of a function here dates back to when we used to
emit the typeAnnotationsHelper as a method on the class.  And we
did that because we needed the resulting code to TS type-check.

None of that is true anymore, and Closure style is to declare members
etc. at the top level, not in a helper.


mprobst:
The problem the function wrapper fixes is that a mere property access can have side effects:

```ts
class Foo {
  static get bar { throw new Error() };
}

class MyFoo extends Foo {
}
/** @type {blergh} */
MyFoo.prototype.bar;  // throws new Error() at runtime
```

mprobst:
This is the potential fix for #761 (linking the issues together).

evmar:
Dropping this, we'd need to figure out what works against google3 before we try a public change.

PR #763, opened 2018-03-13
add a test and note about why we have a wrapper

evmar:
Include a demo of the getter pattern that made our
  function ${className}_tsickle_Closure_declarations() {
wrapper necessary.

Note that the test doesn't actually fail if we break this, but
at least there's a big comment in the output that will hopefully
make it more obvious.

### DIFF
--- a/test_files/fields/fields.ts
+++ b/test_files/fields/fields.ts
@@ -22,3 +22,17 @@ fieldsTest.field1 = 'hi';
 let AnonymousClass = class {
   field: number;
 };
+
+
+class BaseThatThrows {
+  get throwMe(): number { throw new Error(); }
+}
+class Derived extends BaseThatThrows {
+  /**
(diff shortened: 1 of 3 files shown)


test_files/fields/fields.ts
```diff
@@ -22,3 +22,17 @@ fieldsTest.field1 = 'hi';
 let AnonymousClass = class {
   field: number;
 };
+
+
+class BaseThatThrows {
+  get throwMe(): number { throw new Error(); }
```
mprobst: have you tried and seen this fail? I'm always unsure on whether this happens if you declare it as as static or as an instance property.

evmar: Yep, I produced this case by verifying it throws.  Just to double-check I pasted the JS code in the test case into a a JS repl and verified that it throws if I remove the function wrapper.

PR #765, opened 2018-03-14
Skip emitting clutz aliases for default exports.

LucasSloan:
The current emit for default exports is broken and the alternative path from tsickle to clutz is also broken, so no one is currently relying on the current behaviour.  Since the style guide forbids default exports, the easiest way to get error free .d.ts files is to skip emitting aliases for default exports.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -2156,6 +2156,15 @@ function addClutzAliases(
       if (declarations && !areAnyDeclarationsFromSourceFile(declarations, sf)) {
         continue;
       }
+
+      // default is a keyword in typescript, so the name of the export being default
+      // means that it's a default export
+      if (symbol.name === 'default') {
+        dtsFileContent +=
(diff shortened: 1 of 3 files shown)


src/tsickle.ts
```diff
@@ -2156,6 +2156,14 @@ function addClutzAliases(
       if (declarations && !areAnyDeclarationsFromSourceFile(declarations, sf)) {
         continue;
       }
+
+      // default is a keyword in typescript, so the name of the export being default
+      // means that it's a default export
+      if (symbol.name === 'default') {
+        dtsFileContent += '// skipped emitting clutz aliases for a default export';
```
evmar: Maybe make this comment say why? (e.g. "not supported yet" or some such?)

LucasSloan: Done!

evmar: Looks like you need to rebaseline the test too.  (may as well add in the missing trailing newline while you're at it)

PR #767, opened 2018-03-16
 Emit same module stanza in ES5 and ES6 modes.

mprobst:
Previously, tsickle would emit different module stanzas in ES5 vs ES6:

    var module = module || {id: '...'};  // ES5
    var module = {id: '...'};            // ES6

Using `module ||` will allow the ES6 code to run in NodeJS and still
receive the correct `module` object.

This code was added as is from the very get go. It is unclear why the
difference was important, maybe the author wanted to avoid accidentally
exposing a `module` object in ES6 mode.

I have confirmed Closure Compiler fully removes the module assignment
in either mode, and does not produce extra warnings, so this seems to
have been unnecessary.

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -137,7 +137,7 @@ class ES5Processor extends Rewriter {
       }
       // The module=module bit suppresses an unused variable warning which
       // may trigger depending on the compilation flags.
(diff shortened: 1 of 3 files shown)


mprobst:
@vikerman you added the code originally, do you happen to remember why?

Issue #768, opened 2018-03-17
Const enums when using preservConstEnums are not being exported


alan-agius4:
I have a `const enum` in `TypeScript` and `preservConstEnums` is set to true.

When using NGC and afterwards tsickle kicks in the output of the file is correct example (As one can see the const enum is preserved an emitted in the final bundle)

```ts
/**
 * @fileoverview added by tsickle
 * @suppress {checkTypes} checked by tsc
 */
export { SkeletonComponent } from "./src/skeleton/skeleton.component";
export { Package1Module } from "./src/module";
/** @enum {string} */
const Foo = {
    bar: "bar",
};
export { Foo };
//# sourceMappingURL=public-api.js.map
```

However this is not exported from the entry file.
```ts
/**
 * @fileoverview added by tsickle
 * @suppress {checkTypes} checked by tsc
 */
/**
 * Generated bundle index. Do not edit.
 */
export { SkeletonComponent, Package1Module } from './public-api';
//# sourceMappingURL=speedy-package-1.js.map
```

Note: My public-api file I am using barrels. ex:
```ts
export * from "./src/skeleton/skeleton.component";
export * from "./src/module";
```

It looks like `const enums` are never included in the export, Even though they are preserved.

I would expect that preserved const enums are treated the same as enums.

mprobst:
Can you try adding an explicit export for the enum from your barrel file
and see what that does? It could be that by rewriting the enum into an
@enum, TypeScript looses track of the symbol and then doesn't export in the
export * line.

PR #769, opened 2018-03-17
fix: `const enums` when using `preserveConstEnums` are not being exported

alan-agius4:
At the moment `const enums` are not exported when the user specifically chooses to preserve them by setting`preserveConstEnums` to `true` in the `CompilerOptions`.

Closes: #768


mprobst:
Thanks for the contribution!

PR #771, opened 2018-03-19
Break up variable statements into declarations.

mprobst:
tsickle would previously emit variable types inline:

    var /** @type {string} */ x, /** @type {number} y;

This has two issues:
* For exported variables, the type is dropped entirely by TS' emit.
* JSDoc preceding the variable declaration is not merged with the JSDoc
  containing the @type tag.

This change breaks up variable declaration lists into the individual
declarations contained in them, and emits any preceding JSDoc on each
of them, merged with the `@type` declaration.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -777,19 +777,6 @@ class Annotator extends ClosureRewriter {
         // Emit the TS interface verbatim, with no tsickle processing of properties.
         this.writeRange(node, node.getFullStart(), node.getEnd());
         return true;
-      case ts.SyntaxKind.VariableDeclaration:
-        const varDecl = node as ts.VariableDeclaration;
-        // Only emit a type annotation when it's a plain variable and
-        // not a binding pattern, as Closure doesn't(?) have a syntax
-        // for annotating binding patterns.  See issue #128.
-        // Don't emit type annotation when the variable statement is a @polymerBehavior,
-        // as otherwise the polymer closure checker will fail.
-        // See b/64389806
-        if (this.polymerBehaviorStackCount === 0 &&
-            varDecl.name.kind === ts.SyntaxKind.Identifier) {
-          this.emitJSDocType(varDecl);
-        }
-        return false;
       case ts.SyntaxKind.ClassDeclaration:
         const classNode = node as ts.ClassDeclaration;
         this.visitClassDeclaration(classNode);
@@ -914,18 +901,27 @@ class Annotator extends ClosureRewriter {
         }
 
         if (docTags.length > 0 && node.getFirstToken()) {
-          this.emit('\n');
-          this.emit(jsdoc.toString(docTags));
           const isPolymerBehavior = docTags.some(t => t.tagName === 'polymerBehavior');
           if (isPolymerBehavior) {
             this.polymerBehaviorStackCount++;
           }
-          this.writeNodeFrom(node, node.getStart());
+          if (ts.isVariableStatement(node)) {
+            this.visitVariableStatement(node, docTags);
+          } else {
+            // Property declaration.
+            this.emit('\n');
+            this.emit(jsdoc.toString(docTags));
+            this.writeNodeFrom(node, node.getStart());
+          }
           if (isPolymerBehavior) {
             this.polymerBehaviorStackCount--;
           }
           return true;
+        } else if (ts.isVariableStatement(node)) {
+          this.visitVariableStatement(node, docTags);
+          return true;
         }
+        // Property declaration without doc tags.
         break;
       case ts.SyntaxKind.PropertyAssignment:
         const pa = node as ts.PropertyAssignment;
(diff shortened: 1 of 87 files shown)


mprobst:
This fixes ##776.

mprobst:
@evmar please take a look.

I took a look and decided that this PR will make it a lot easier to move the rest of the code to transformers, as it'll also allow us to drop hacks such as the polymer behaviour stack count. So I think landing this first has value.

I think I found a reasonable workaround for the question mark/conformance issue by not emitting any type at all for variables that have a blacklisted type that are also initialized (so Closure will likely infer). This PR is causing some compilation failures internally due to tightened checks, but I'm going through those and cleaning them up - they seem fixable (and worth fixing).

This PR required some source map gymnastics and removes one source map related test, as with the transformers source map hack, it's impossible to produce correct mappings for comments outside of nodes. The good news is that Closure still reports everything correctly as it uses the nodes, not the comments. The other good news is that we'll hopefully be able to drop the entire source map stuff soonish.

src/tsickle.ts
```diff
@@ -1209,6 +1223,46 @@ class Annotator extends ClosureRewriter {
     return emitText;
   }
 
+  /** Emits a variable statement per declaration in this variable statement, duplicating JSDoc. */
+  private visitVariableStatement(varStmt: ts.VariableStatement, docTags: jsdoc.Tag[]) {
+    // Previously tsickle would emit inline types (`var /** @type {string} */ x;`), but TypeScript's
+    // emit strips those comments for exports.
```
evmar: Can you complete this comment with what we do now?  E.g.
// So instead we now emit [...].

src/tsickle.ts
```diff
@@ -1209,6 +1223,46 @@ class Annotator extends ClosureRewriter {
...
+    const keyword = varStmt.declarationList.getFirstToken().getText();
+    const additionalTags = jsdoc.toStringWithoutStartEnd(docTags, jsdoc.TAGS_CONFLICTING_WITH_TYPE);
+    for (const decl of varStmt.declarationList.declarations) {
+      this.emit('\n');
+      this.addSourceMapping(decl);
+      // Only emit a type for plain identifiers. Binding patterns have no type syntax in Closure
+      // Compiler, and TypeScript 2.6 crashes on "getTypeAtLocation(bindingpattern)". By emitting a
```
evmar: s/emitting/omitting?

test/BUILD
```diff
@@ -58,6 +56,33 @@ jasmine_node_test(
...
+    ],
+    tsconfig = "//:tsconfig.json",
+    deps = [
+        ":test_support",
+        "//src",
+    ],
+)
+
+jasmine_node_test(
+    name = "closure_e2e_test",
```
evmar: You'll need to duplicate this in the internal test/BUILD file when you import it.

test_files/arrow_fn/arrow_fn.ts
```diff
@@ -1,3 +1,6 @@
 var fn3 = (a: number): number => 12;
 var fn4 = (a) => a + 12;
 export var fn5 = (a = 10) => a;
+
+/** @param a this must be escaped, as Closure bails on it. */
```
evmar: Clarify what 'this' refers to(?)

test_files/duplicated_externs.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @fileoverview A file containing externs duplicating the symbols generated from
+ * declare/declare.d.ts (and thus declare/externs.js). This is used to ensure generated externs do
+ * not conflict with pre-existing externs.
+ * @externs
+ */
+
+/** @type {number} */
+var declareGlobalVar;
```
evmar: I vaguely recall that you need to explicitly add these files in the test suite somewhere, but I don't see a corresponding change here that does that.

src/tsickle.ts

src/tsickle.ts

test/BUILD
mprobst: Ack.

test_files/arrow_fn/arrow_fn.ts
mprobst: Done.

test_files/duplicated_externs.js
mprobst: This is a bit embarrassing, I think I wanted to add this in a PR long past. I updated the test runner to include it.

mprobst:
@evmar PTAL.

test_files/duplicated_externs.js
evmar: It seems there is no affected test (?).  Maybe this belongs in some other CL?

mprobst:
> It seems there is no affected test (?). Maybe this belongs in some other CL?

Yeah it arguably does, but I seem to have lost the initial context for this. I'm piggybacking on this PR.

PR #772, opened 2018-03-19
Remove renovate configuration.

mprobst:
In summary, this overall seems to generate more work than the benefit
in this particular repository.


rarkins:
@mprobst I've now disabled the bot server-side for this repository. If you delete your custom config first while keeping the bot installed then you'd just end up with the default bot settings instead.

I'm curious as to whether you have a "dream scenario" where Renovate should be doing something differently?

Absence of dependency automation usually leads to dependency rot, e.g. this repository itself had two breaking in-range dependency updates prior to Renovate, which I thought at the time was a perfect case study for why automation helps.

I assume earlier today you were maybe frustrated by the "Update branch" requirement? Rebasing is something Renovate could do on its own if allowed. Otherwise I don't see how 1 PR per month ("Update all dependencies") with conveniently embedded changelogs to save you looking them up is pretty light work.

alexeagle:
@rarkins a bit of context - we used angular/tsickle as a dry-run before enabling Renovate in angular/angular - I think this was a success and we plan to start with Renovate as soon as we get Angular 6 through Release Candidate.
The fact that we don't continue using it for tsickle is not a big deal.

rarkins:
FYI: I have added this tsickle project to a “Renovate Mirrors” org here: https://github.com/renovate-mirrors/tsickle/pulls

That fork shadows/mirrors this source repository (syncing its master with this upstream repo hourly) and behaves essentially like before except that it submits PRs to its *own* fork and not back to this upstream repo. You can browse the PRs at any time to see what’s out of date and can raise PRs back to here yourself fairly easily by selecting any of the Renovate branches there and raising a new (second) PR to here.

PR #774, opened 2018-03-20
Do not emit unknown types.

mprobst:
Explicitly emitting unknown types can cause issues with Closure Compiler, where it would normally infer a known type, and the unknown type causes a warning later down (e.g. due to a conformance check for goog.debug.Logger or similar). Skipping the type altogether allows the type inference to kick in and often do a better job. If type inference fails, just declaring the variable or property ("X.prototype.foo;") has the same effect as a ? declaration.

This has a chance of causing stricter type checks for downstream code. For TypeScript code, type checks should be disabled, so this should not cause additional errors. For Closure JS code using TS code, this should hopefully be rare enough to be manageable.


mprobst:
The context is that #771 causes more `/** @type {?} */` to be emitted for exported variables, which then cause worse type inference, which then causes conformance errors. This is a bit of a long shot and could have all kinds of unintended side effects, let's see if it works out.

mprobst:
This does not work, as it incorrectly triggers Closure Compiler's heuristics on what is a declaration and what is not.

PR #777, opened 2018-03-27
Handle inferred types.

mprobst:
Previously, tsickle would only handle defined literal types, e.g.:
    const x: {a: number} = {a: 1};

With this change, tsickle will also emit a type for inferred types:
    const x = {a: 1};

Both of these convert to the same type.

### DIFF
--- a/src/type-translator.ts
+++ b/src/type-translator.ts
@@ -135,7 +135,7 @@ export class TypeTranslator {
    * A list of type literals we've encountered while emitting; used to avoid getting stuck in
    * recursive types.
    */
-  private readonly seenTypeLiterals = new Set<ts.Type>();
+  private readonly seenAnonymousTypes = new Set<ts.Type>();
 
   /**
    * Whether to write types suitable for an \@externs file. Externs types must not refer to
@@ -279,7 +279,7 @@ export class TypeTranslator {
     // Avoid infinite loops on recursive type literals.
     // It would be nice to just emit the name of the recursive type here (in type.aliasSymbol
     // below), but Closure Compiler does not allow recursive type definitions.
-    if (this.seenTypeLiterals.has(type)) return '?';
+    if (this.seenAnonymousTypes.has(type)) return '?';
 
     // If type is an alias, e.g. from type X = A|B, then always emit the alias, not the underlying
     // union type, as the alias is the user visible, imported symbol.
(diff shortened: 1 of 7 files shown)


src/type-translator.ts
```diff
@@ -488,18 +488,17 @@ export class TypeTranslator {
...
+      if (type.symbol.flags & ts.SymbolFlags.Function ||
           type.symbol.flags & ts.SymbolFlags.Method) {
         const sigs = this.typeChecker.getSignaturesOfType(type, ts.SignatureKind.Call);
         if (sigs.length === 1) {
           return this.signatureToClosure(sigs[0]);
         }
+        this.warn('unhandled anonymous type with multiple call signatures');
+        return '?';
+      } else {
+        return this.translateTypeLiteral(type);
```
evmar: Is there any other case (other than function, method, type literal)?  Maybe it's worth making this an 'else if' to allow an 'else' bailout?

test_files/variables/variables.ts
```diff
@@ -1,2 +1,4 @@
 var v1: string;
 var v2: string, v3: number;
+
+const inferred = {a: 1, b: [{c: '2'}]};
```
evmar: Can you add a comment to this about what property we're looking for in the output?  (I know we haven't done this in the past, but I have been trying to do this on new tests for future sanity.)

src/type-translator.ts
mprobst: All object types are essentially the empty type plus some properties defined on it. We special case `function` and `method` here because we know how to emit call signatures in a way that fits Closure, and we special case construct signatures inside of translate below for the same reason (I suspect we could simplify this, e.g. shouldn't all functions in the end of the day be types with just one call signature? we handle that below in translate).

So I believe that translateTypeLiteral actually implements an arbitrary translateAnonymousType. We still fall back inside of it if we can't figure out the properties. I think that should be enough?

test_files/variables/variables.ts
mprobst: Done.

mprobst:
PTAL.

mprobst:
By the way, of course the global presubmit will prove me wrong...

mprobst:
In a surprise twist, TGP actually looks green with two or three fiex.

PR #783, opened 2018-04-04
release: prevent publishing sources

alexeagle:


package.json
```diff
@@ -70,5 +70,7 @@
   "bugs": {
     "url": "https://github.com/angular/tsickle/issues"
   },
-  "homepage": "https://github.com/angular/tsickle"
+  "homepage": "https://github.com/angular/tsickle",
+  "private": "See release instructions in the README.md",
+  "private": true
```
evmar: Duplicated line here intentional?

alexeagle: yes, that's the poor-man's JSON comment (there are no wealthy JSON users)

Issue #352, opened 2017-01-31
Namespace / mangle type names made global in externs


mprobst:
Currently, a TypeScript file containing ambient declarations writes out a secondary externs file:

```ts
declare interface Foo { ... }
```

Generates an `externs.js`:

```js
/** @interface */
function Foo() {}
```

The problem is that `Foo` in the externs file is a global type, whereas the `Foo` in the TypeScript file was scoped to the local module. Thus, multiple `Foo`s in multiple files can conflict.

A potential fix would be to namespace/mangle `Foo` as e.g. `_tsickle_extern_path_to_the_file_Foo` and then include an alias in the Closure output (something like `var Foo = _tsickle_extern_path_to_the_file_Foo;`).

evmar:
I agree.  Note that it's possible for file A to `export declare interface Foo` then for file B to `import {Foo} from ''a'`, which means that B needs some way of knowing whether `Foo` is a mangled Closure name or not.  I *think* you can get that from looking at `symbol.declarations`.  (It's similar to the problem where we decide to not emit the Closure `var` for a TS `interface` because of a conflict with an existing TS `var`; we need to know to not use the interface from other files as well.  And that can also be solved by name-mangling so maybe we'll have some more generic `closureNameForTypeScriptType` that knows these mangling rules for use in JSDoc.)

mprobst:
If we just re export the names from the module, we don't need to do
anything special at the use site.

Evan Martin <notifications@github.com> schrieb am Di., 31. Jan. 2017, 16:28:

evmar:
Oh yeah, duh.  OK.

b-strauss:
Is this only a problem with externs? I would also think that closure has problems with duplicate types in "internal" code.

mprobst:
@b-strauss within a `goog.module`, all symbols, including types, are scoped to the local file, and can only be seen "on the outside" through explicitly exporting them. So there's no conflict there.

evmar:
This affected google3: someone declared a local type named Location, which ended up affecting the browser global type named Location.

mprobst:
Fixed a while ago.

PR #785, opened 2018-04-09
extend the declare_export test to showcase a bug

evmar:
This doesn't change any behavior, but extends the test suite
to include some more comments about a particular bug.  This will
make the diff from the fix easier to read.

A step on #352.


evmar:
CC @mprobst  FYI

PR #786, opened 2018-04-10
Make sure to ts.setTextRange on default imports.

mprobst:


mprobst:
Sorry, hold off on review, this isn't ready.

mprobst:
@evmar ready to review.

test/golden_tsickle_test.ts
```diff
@@ -106,7 +106,7 @@ const testFn = TEST_FILTER ? fdescribe : describe;
 testFn('golden tests with transformer', () => {
   testSupport.goldenTests().forEach((test) => {
     if (TEST_FILTER && !TEST_FILTER.testName.test(test.name)) {
-      xit(test.name);
+      // do not xit(test.name) as that spams a lot of useless console msgs.
```
evmar: Hooray!

test_files/file_comment/export_star.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * @fileoverview fileoverview comment before export.
+ */
+
+export * from './comment_before_var';
```
evmar: Add some docs (below the export if the comments matter) about what this is testing for.

test_files/file_comment/side_effect_import.ts
```diff
@@ -0,0 +1,5 @@
+/**
+ * @fileoverview This is a fileoverview comment preceding a side-effect import.
+ */
+
+import './file_comment';
```
evmar: Here too.

src/transformer_util.ts
```diff
@@ -183,7 +183,7 @@ function emitMissingSyntheticCommentsAfterTypescriptTransform(context: ts.Transf
         }
       }
       // TypeScript ignores synthetic comments on reexport / import statements.
-      const moduleName = extractModuleNameFromRequireVariableStatement(node);
+      const moduleName = isCommonJsRequireStatement(node);
       if (moduleName && fileContext.importOrReexportDeclarations) {
```
evmar: No longer "moduleName", needs a rename.

mprobst:
@evmar PTAL, all done.

src/transformer_util.ts
```diff
@@ -206,28 +207,31 @@ function emitMissingSyntheticCommentsAfterTypescriptTransform(context: ts.Transf
...
-  }
-  const varStmt = node as ts.VariableStatement;
-  const decls = varStmt.declarationList.declarations;
-  let init: ts.Expression|undefined;
-  if (decls.length !== 1 || !(init = decls[0].initializer) ||
-      init.kind !== ts.SyntaxKind.CallExpression) {
-    return null;
+function isCommonJsRequireStatement(node: ts.Node): boolean {
+  // CommonJS requires can be either "var x = require('...');" or (for side effect imports), just
+  // "require('...');".
```
evmar: By the way, I think this code might be duplicated with the ES5 rewriter, which also must find require statements.  At least, it feels pretty familiar.

evmar: https://github.com/angular/tsickle/blob/master/src/es5processor.ts#L247

evmar: Yeah it's fine I think.  It's worth keeping in mind though (the other code I linked has a lot more stuff around export * that I think maybe doesn't apply here?).

mprobst: Ack, I actually added a test for that because I was worried. It turns out not to be a problem though, because we actually do rewrite those from `export *` to explicit exports for each symbol.

PR #787, opened 2018-04-11
Add support for ES transpilation

drjanitor:
This mode prevents tsickle from adding spurious stuff into TSC output
coming from JS-to-JS transpilation.

We still need to run tsickle in order to turn require('tslib') into
goog.require('tslib'). Also, some downstream dependencies only emit code
through tsickle, so bypassing tsickle requires shaving yaks.

### DIFF
--- a/src/es5processor.ts
+++ b/src/es5processor.ts
@@ -31,6 +31,8 @@ export interface Es5ProcessorHost {
   fileNameToModuleId(fileName: string): string;
   /** Whether to convert CommonJS module syntax to `goog.module` Closure imports. */
   googmodule?: boolean;
+  /** Identifies whether this file is the result of a JS transpilation. */
+  isJsTranspilation?: boolean;
   /** Whether the emit targets ES5 or ES6+. */
   es5Mode?: boolean;
   /** expand "import 'foo';" to "import 'foo/index';" if it points to an index file. */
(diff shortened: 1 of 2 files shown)


src/es5processor.ts
```diff
@@ -260,6 +269,10 @@ class ES5Processor extends Rewriter {
       const call = decl.initializer as ts.CallExpression;
       const require = this.extractRequire(call);
       if (!require) return false;
+      // In JS transpilation, the only require('...') that needs to be
+      // rewritten is require('tslib'), which is emitted by TSC when loading
+      // ES6 polyfills.
```
evmar: Does this mean you want to feed plain require() into Closure Compiler?  Does that work?  I feel like I'm missing some context on the motivation for this change.

drjanitor:
this is js code that we just transpiled into ES5. in particular, this
fellow here would crash and burn:

https://github.com/google/closure-library/blob/master/closure/goog/bootstrap/nodejs.js

src/es5processor.ts
```diff
@@ -112,17 +114,24 @@ class ES5Processor extends Rewriter {
     this.emitFileComment();
 
     const moduleId = this.host.fileNameToModuleId(this.file.fileName);
-    const moduleName = this.host.pathToModuleName('', this.file.fileName);
-    // NB: No linebreak after module call so sourcemaps are not offset.
-    this.emit(`goog.module('${moduleName}');`);
+    // The goog.module() call should not be emitted if the file is the result
+    // of transpiling JS sources.
```
mprobst: maybe ", because emitting..."? otherwise this is a bit odd to read.

PR #788, opened 2018-04-18
Port CommonJS -> goog.module to a TS transformer

mprobst:


src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+
+import * as path from 'path';
+
+import {ModulesManifest} from './modules_manifest';
+import {visitNodeWithSynthesizedComments} from './transformer_util';
+import * as ts from './typescript';
+
+export interface GoogModuleProcessorHost {
+  /**
+   * Takes a context (the current file) and the path of the file to import
```
evmar: The current file... path?  Relative to what?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+   */
+  pathToModuleName(context: string, importPath: string): string;
+  /**
+   * If we do googmodule processing, we polyfill module.id, since that's
+   * part of ES6 modules.  This function determines what the module.id will be
+   * for each file.
+   */
+  fileNameToModuleId(fileName: string): string;
+  /** Whether to convert CommonJS module syntax to `goog.module` Closure imports. */
+  googmodule?: boolean;
```
evmar: Why does the GoogModuleProcessor have a googmodule boolean?  What does it mean for this to be false?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+  /** Whether the emit targets ES5 or ES6+. */
+  es5Mode?: boolean;
+  /** expand "import 'foo';" to "import 'foo/index';" if it points to an index file. */
+  convertIndexImportShorthand?: boolean;
+
+  options: ts.CompilerOptions;
+  host: ts.ModuleResolutionHost;
+}
```
evmar: I realize now that this is just a moved file, hmm.  Maybe it's not worth re-reviewing old code?  On the other hand maybe it's worth thinking about critically now that it's fresh?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+
+/** Returns true if expr is "module.exports = ...;". */
+function isModuleExportsAssignment(expr: ts.ExpressionStatement): boolean {
+  if (!ts.isBinaryExpression(expr.expression)) return false;
+  if (expr.expression.operatorToken.kind !== ts.SyntaxKind.EqualsToken) return false;
+  return isPropertyAccess(expr.expression.left, 'module', 'exports');
+}
+
+/** Returns true if expr is "module.exports = ...;". */
+function isExportsAssignment(expr: ts.ExpressionStatement): boolean {
```
evmar: Is this the same as the function above it?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+ *   Object.defineProperty(exports, "__esModule", { value: true });
+ * This matches that code snippet.
+ */
+function isEsModuleProperty(stmt: ts.ExpressionStatement): boolean {
+  // We're matching the explicit source text generated by the TS compiler.
+  // Object.defineProperty(exports, "__esModule", { value: true });
+  const expr = stmt.expression;
+  if (!ts.isCallExpression(expr)) return false;
+  if (!isPropertyAccess(expr.expression, 'Object', 'defineProperty')) return false;
+  // return true;
```
evmar: XXX

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+  if (ident.escapedText !== 'require') return null;
+
+  // Verify the call takes a single string argument and grab it.
+  if (call.arguments.length !== 1) return null;
+  const arg = call.arguments[0];
+  if (arg.kind !== ts.SyntaxKind.StringLiteral) return null;
+  return (arg as ts.StringLiteral).text;
+}
+
+function createGoogCall(methodName: string, literal: ts.StringLiteral): ts.CallExpression {
```
evmar: Docs

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+  return ts.createCall(
+      ts.createPropertyAccess(ts.createIdentifier('goog'), methodName), undefined, [literal]);
+}
+
+/**
+ * Extracts the namespace part of a goog: import, or returns null if the given
+ * import is not a goog: import.
```
evmar: Define "import" here, I think you mean the quoted string passed in an import statement and not the import statement itself.

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+  if (resolvedModule.indexOf('node_modules') === -1 &&
+      requestedModule.substr(requestedModule.lastIndexOf('/')) !==
+          resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
+    imported = './' + path.relative(path.dirname(fileName), resolvedModule).replace(path.sep, '/');
+  }
+  return imported;
+}
+
+/**
+ * importPathToGoogNamespace converts a TS/ES module './import/path' into a goog.module compatible
```
evmar: Maybe mention that it also handles "goog:" imports, otherwise it looks redundant with pathToModuleName.

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+    diagnostics: ts.Diagnostic[]): (context: ts.TransformationContext) =>
+    ts.Transformer<ts.SourceFile> {
+  return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
+    // TS' CommonJS processing uses onSubstituteNode to, at the very end of processing, substitute
+    // "modulename.default" for default imports. We intercept the substitution here, check if it's a
+    // .default access, then check if the original node (and thus original import) was from a goog:
+    // module, and if so, replace with just the module name.
+    const previousOnSubstituteNode = context.onSubstituteNode;
+    context.enableSubstitution(ts.SyntaxKind.PropertyAccessExpression);
+    context.onSubstituteNode = (hint, node: ts.Node): ts.Node => {
```
evmar: The containing function here is pretty big, maybe you could split this out?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+      let moduleVarCounter = 1;
+      function nextModuleVar() {
+        return `tsickle_module_${moduleVarCounter++}_`;
+      }
+
+      /**
+       * Replace "module.exports = ..." with just "exports = ...". Returns null if `expr` is not an
+       * exports assignment.
+       */
+      function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
```
evmar: Same here for splitting.

src/tsickle.ts
```diff
@@ -2025,12 +2025,16 @@ export function emitWithTsickle(
     }));
     tsickleSourceTransformers.push(
         classDecoratorDownlevelTransformer(typeChecker, tsickleDiagnostics));
+  } else {
+    // In either case, run a transformation that synthesizes all comments for consistency.
```
evmar: "In either case" here makes me think we do this if either of the above are true, but really only happens if they are both false.  Maybe you mean "in any case"?

test_files/abstract/abstract.js
```diff
@@ -2,7 +2,9 @@
  * @fileoverview added by tsickle
  * @suppress {checkTypes} checked by tsc
  */
-goog.module('test_files.abstract.abstract');var module = module || {id: 'test_files/abstract/abstract.js'};/**
+goog.module('test_files.abstract.abstract');
+var module = module || { id: 'test_files/abstract/abstract.ts' };
```
evmar: Ah this is the test change.  That is good, it's been bugging me.  (Usual Evan complaint: would be easier to review if you made "add whitespace" a separate change from the one you're making here.)

tsconfig.json
```diff
@@ -6,6 +6,7 @@
     "jsx": "react",
     "types": ["node", "jasmine"],
     "downlevelIteration": true,
+    "target": "es2015",
```
evmar: Why is this one necessary?

src/googmodule.ts
mprobst: Done.

src/googmodule.ts
mprobst: Actually, it turns out with this code change we catch whether to do the `googmodule` transformation on a higher level, so this can indeed go away.

src/googmodule.ts
mprobst: Yes, we should re-think what's needed here.

src/googmodule.ts
mprobst: Done.

src/tsickle.ts
mprobst: Reworded.

test_files/abstract/abstract.js
mprobst: I cannot not make the whitespace change together with this. I'm emitting using the TS API, which then later on indents as it likes. The advantage is that it also fixes up the source map, so we no longer need to worry about that.

tsconfig.json
mprobst: We removed these when alex moved them into the BUILD level config. But then you get lots of complaints in your IDE about features not being supported, so I'm adding back in what's needed.

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+            // later on though, so appears as a plain "__exportStar" on the top level here.
+            const isExportStar =
+                (ts.isIdentifier(expr.expression) && expr.expression.text === '__exportStar') ||
+                (ts.isIdentifier(expr.expression) && expr.expression.text === '__export');
+            if (isExportStar) callExpr = expr.arguments[0] as ts.CallExpression;
+            const ident = ts.createIdentifier(nextModuleVar());
+            const require = maybeCreateGoogRequire(exprStmt, callExpr, ident);
+            if (!require) break;
+            statements.push(require);
+            // If this is an export star
```
evmar: Seems redundant with the following line, maybe commit got snipped?

src/googmodule.ts
```diff
@@ -0,0 +1,388 @@
...
+            if (isExportStar) {
+              const args: ts.Expression[] = [ident];
+              if (expr.arguments.length > 1) args.push(expr.arguments[1]);
+              statements.push(ts.createStatement(ts.createCall(expr.expression, undefined, args)));
+            }
+            return;
```
evmar: The flow through this is a bit weird.  It seems like it's a series of
if (special case) handleIt && return;
and then "break" means "do the normal processing".  Here, I would expect the isExportStar to return and then it to break after.

If that's not the case, maybe some comments about what is expected to fall out at the end here would help?

tsconfig.json
evmar: IIRC this breaks something else (when I last tried it), but maybe I did something else wrong.  No problem.

src/googmodule.ts
mprobst: Completed the comment.

src/googmodule.ts
mprobst: I've added a comment to the function.

The invariant is that the function will add a node to statements in any case, even if not handling. We need it like that because there is one case where we have to add two nodes (`export *`).

We return once we're done with the node, otherwise break to the default handling. Does that make sense?

src/googmodule.ts
```diff
@@ -0,0 +1,416 @@
...
+            // later on though, so appears as a plain "__exportStar" on the top level here.
+            const isExportStar =
+                (ts.isIdentifier(expr.expression) && expr.expression.text === '__exportStar') ||
+                (ts.isIdentifier(expr.expression) && expr.expression.text === '__export');
+            if (isExportStar) callExpr = expr.arguments[0] as ts.CallExpression;
+            const ident = ts.createIdentifier(nextModuleVar());
+            const require = maybeCreateGoogRequire(exprStmt, callExpr, ident);
+            if (!require) break;
+            statements.push(require);
+            // If this is an export star. Split it up into the import (created by the maybe call
```
evmar: Maybe comma after "star", reads pretty weird otherwise.

tsconfig.json
mprobst: Seems to work so far, but we get a warning from the bazel build. I think that's better than a broken IDE...

src/googmodule.ts
```diff
@@ -0,0 +1,416 @@
...
+            }
+            // Check for:
+            //   "require('foo');" (a require for its side effects)
+            const expr = exprStmt.expression;
+            if (!ts.isCallExpression(expr)) break;
+            let callExpr = expr;
+            // Handle export * in ES5 mode (in ES6 mode, export * is dereferenced already).
+            // export * creates either a pure top-level '__exportStar(require(...))' or the imported
+            // version, 'tslib.__exportStar(require(...))'. The imported version is only substituted
+            // later on though, so appears as a plain "__exportStar" on the top level here.
```
evmar: Why does the below code look for __export as well?

src/googmodule.ts
mprobst: Done, that was a typo.

src/googmodule.ts
mprobst: It's `__export()` for the non-tslib version, and `__exportStar` for the tslib case. Fixed the comment.

PR #793, opened 2018-05-08
Emit file comments when only running googmodule. 

mprobst:
Previously, when tsickle ran only the googmodule transformer, e.g. when
both type annotation and decorator conversion is false, file level
comments would not be emitted, causing problems with some load bearing
comments. Always running at least the synthesize comments pass avoids
the issue.


src/tsickle.ts
```diff
@@ -2055,6 +2055,9 @@ export function emitWithTsickle(
     ]
   };
   if (host.googmodule) {
```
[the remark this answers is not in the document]
mprobst: No, that doesn't work. The googmodule processor must run for all files, even if `skipTransformForSourceFileIfNeeded` is true for them, so it cannot go into the regular `after` bucket. It also shouldn't be in the `afterTs` bucket, as any additional `after` ops would then happen after it, but the `googmodule` converter really should be last.

On the bright side, I've debugged this, and was able to get rid of using `synthesizeCommentsTransformer` here (pending TAP, but I've also been able to reproduce the issue locally). So this is less messy now.

test_files/file_comment/comment_before_elided_import.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * @fileoverview This is a comment before an import that will be elided.
```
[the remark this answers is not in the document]
mprobst: Clarified. The `import` is what's being elided, not the comment, which should indeed be kept (that's the point of the test).

test_files/file_comment.puretransform/comment_before_var.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @fileoverview This comment must not be emitted twice.
+ * @mods {google3.java.com.google.javascript.typescript.examples.boqui.boqui}
+ * @modName {foobar}
```
LucasSloan: Is mods and modName load bearing here?  Either explain, rename test, or generate 2 tests - one for `@mods` and one for comment before a constant.

src/tsickle.ts
```diff
@@ -19,7 +19,7 @@ import {ModulesManifest} from './modules_manifest';
 import {getEntityNameText, getIdentifierText, Rewriter, unescapeName} from './rewriter';
 import {containsInlineSourceMap, extractInlineSourceMap, parseSourceMap, removeInlineSourceMap, setInlineSourceMap, SourceMapper, SourcePosition} from './source_map_utils';
 import {createTransformerFromSourceMap} from './transformer_sourcemap';
-import {createCustomTransformers, noOpTransformer} from './transformer_util';
+import {createCustomTransformers, synthesizeCommentsTransformer} from './transformer_util';
```
LucasSloan: I'm not seeing a diff corresponding to this import's name being changed - unused import?

src/googmodule.ts
```diff
@@ -166,13 +166,11 @@ export function resolveIndexShorthand(
 function importPathToGoogNamespace(
     host: GoogModuleProcessorHost, file: ts.SourceFile, tsImport: string): ts.StringLiteral {
   let modName: string;
-  let isGoogImport = false;
```
LucasSloan: Unused?

src/tsickle.ts
mprobst: Good catch, done.

src/googmodule.ts
mprobst: Yes, this was unused, and is now removed.

test_files/file_comment.puretransform/comment_before_var.ts
mprobst: This is a 1:1 copy of the previously existing test (`file_comment/comment_before_var.ts`). The original test makes sure that merging a file overview comment before an exported var works. It must contain markers that it is a fileoverview comment. It does not quite matter what those are (fileoverview or mods etc), but this repros the exact problem we had in google3.

In the puretransform case, the tags in the comment are insignificant, but we generally try to keep the files identical between multiple variants (read: copies) of the same test case for different situations.

I could either change both or keep both this way I think. Given that changing them doesn't fix anything and is unrelated to this PR, I think I'd rather keep them the way they are.

PR #796, opened 2018-05-10
Emit exported typedefs for use in the local scope.

mprobst:
Previously, tsickle would emit an `exports.Foo` typedef, but then
references to this typedef in the local scope would use just `Foo`,
which breaks compilation.

Closure however does not support referencing `exports.Foo` within a
module. Additionally, because typedefs do not create values, they cannot
be aliased and re-exported (`exports.Foo = Foo;`). Thus the only
recourse is to emit the same typedef twice, once for export, once for
local use.

While not particularly elegant, given that typedefs in Closure have no
identity and are pure aliases, this should not create problems.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1417,21 +1417,22 @@ class Annotator extends ClosureRewriter {
     if (sym.flags & ts.SymbolFlags.Value) return;
 
     // Write a Closure typedef, which involves an unused "var" declaration.
-    // Note: in the case of an export, we cannot emit a literal "var" because
-    // TypeScript drops exports that are never assigned to (and Closure
-    // requires us to not assign to typedef exports).  Instead, emit the
-    // "exports.foo;" line directly in that case.
     this.newTypeTranslator(node).blacklistTypeParameters(
         this.symbolsToAliasedNames, node.typeParameters);
 
     const typeStr = this.typeToClosure(node, undefined, true /* resolveAlias */);
+    const typeName = node.name.getText();
     this.emit(`\n/** @typedef {${typeStr}} */\n`);
+    this.emit(`var ${typeName};\n`);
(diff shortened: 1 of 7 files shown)


src/tsickle.ts
```diff
@@ -1417,21 +1417,22 @@ class Annotator extends ClosureRewriter {
...
 
     const typeStr = this.typeToClosure(node, undefined, true /* resolveAlias */);
+    const typeName = node.name.getText();
     this.emit(`\n/** @typedef {${typeStr}} */\n`);
+    this.emit(`var ${typeName};\n`);
+    // In the case of an export, we cannot emit a `export var foo;` because TypeScript drops exports
+    // that are never assigned to (and Closure requires us to not assign to typedef exports).
+    // Instead, emit an additional "exports.foo;" line that defines the same typedef.
+    // tsickle must still emit the `var` line above so that the type can be used within the local
+    // module scope (Closure does not allow refering to "exports.Foo" within a module).
```
rkirov: In my mind the most idiomatic Closure is 

```
/** @typedef {} */
var Foo;
exports.Foo = Foo;
```

And kinda surprised that doesn't work. If they are serious about treating type-time objects as module scoped entities there should be a bug for that. Also add that to the comment of things that would be better than the current solution but don't work :/ (you have it in the cl description but not the comment here)

mprobst: I was reading the original comment here: "and Closure requires us to not assign to typedef exports" as meaning Closure requires us to never assign anything into the export. I think what it actually meant was we cannot assign a value (e.g. `{}`) into a `typedef` in Closure. I've confirmed that in a minimal example including use from a different site. Yay. It's still weird to assign an undefined value around, but you're right in that it is more idiomatic.

PR #798, opened 2018-05-10
Support property access on imported types.

mprobst:
TypeScript uses forward declares to refer to imported modules in type
annotations to match TS' loading semantics:

    import {SomeImport} from 'some-module';
    const foo: SomeImport = ...;

tsickle will introduce a forward declare for the imported module:

    var tsickle_fwd_decl_1 = goog.forwardDeclare('some-module')'
    const /** @type {!tsickle_fwd_decl_1.SomeImport} */ foo = ...;

TypeScript code can use types nested in an imported symbol:

    import {SomeImport} from 'some-module';
    const bar: SomeImport.SomeNestedType = ...;

Previously, tsickle would only use an alias for a top level symbol
directly (`SomeImport`), but not for a symbol accessed on an aliased
symbol. This change has tsickle consult the alias map for all symbols
encountered in a type expression.

    const bar: SomeImport.SomeNestedType = ...;
    // gets emitted as:
    const /** @type {!tsickle_fwd_decl_1.SomeImport.SomeNestedType} */ foo = ...;

This also required a change to make sure tsickle does not introduce
aliases for exported declared namespaces, as those do not have an
exported symbol that can be aliased but should rather fall back to the
global symbol emitted for them (covered by existing tests).

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1216,6 +1216,15 @@ class Annotator extends ClosureRewriter {
       if (sym.flags & ts.SymbolFlags.Alias) {
         sym = this.typeChecker.getAliasedSymbol(sym);
       }
+      // tsickle does not emit exports for ambient namespace declarations:
+      //    "export declare namespace {...}"
+      // So tsickle must not introduce aliases for them that point to the imported module, as those
+      // then don't resolve in Closure Compiler.
+      if (!sym.declarations ||
+          !sym.declarations.some(
+              d => d.kind !== ts.SyntaxKind.ModuleDeclaration || !isAmbient(d))) {
+        continue;
+      }
       // goog: imports don't actually use the .default property that TS thinks they have.
       const qualifiedName = nsImport && isDefaultImport ? forwardDeclarePrefix :
                                                           forwardDeclarePrefix + '.' + sym.name;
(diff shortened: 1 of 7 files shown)


test_files/type_propaccess.no_externs/clutzed_namespace.d.ts
```diff
@@ -0,0 +1,11 @@
+declare namespace ಠ_ಠ.clutz.type_propaccess.nested.clazz {
```
rkirov: basic question - why is tsickle consuming clutz .d.ts file? Those are supposed to have a source .js, closure should be consuming those and not the clutz->tsickle abomination.

src/tsickle.ts
```diff
@@ -1216,6 +1216,15 @@ class Annotator extends ClosureRewriter {
       if (sym.flags & ts.SymbolFlags.Alias) {
         sym = this.typeChecker.getAliasedSymbol(sym);
       }
+      // tsickle does not emit exports for ambient namespace declarations:
+      //    "export declare namespace {...}"
```
rkirov: Sorry, this keeps coming up again, but I need to get my terminology straight - you are saying this namespace is ambient because it has a 'declare'? I previously thought 'ambient' means not in a module, which cannot be true here because you have an 'export' keyword.

Also in the clutzed.d.ts there is no export

test_files/type_propaccess.no_externs/clutzed_namespace.d.ts
mprobst: This simulates code referencing clutz processed TS, but the clutz code does not go through tickle.

src/tsickle.ts
mprobst: declare means ambient. export means it's in a proper module as oppose to a namespace. I'd call a pure namespace that's not nested in a module "global".

This fix is actually covered by a different pre existing test that was tickled by the other change, the .d.ts covers a different case.

PR #803, opened 2018-05-14
Dereference aliases when checking for exported values.

mprobst:
tsickle emits a "forced load" when importing a module that only contains
types. The check whether an exported symbols is a value previously
happened *before* dereferencing aliases, which causes false negatives:

    const X = 1; // the original symbol X, which has the Value flag set
    export {X};  // creates a new symbol X, which is an alias of X above
                 // but does not inherit the Value flag.

tsickle would think this file only exports types, because the alias
symbol does not have the value flag set. This would lead tsickle to
emit duplicate goog.require statements:

    goog.module('importer');
    var exporter_1 = goog.require('exporter');
    const tsickle_forward_declare_1 = goog.forwardDeclare("exporter");
    goog.require("exporter"); // force type-only module to be loaded
    console.log(exporter_1.FOO);

With certain compiler flags, Closure Compiler would complain about the
duplicated import in the `// force type-only` line, causing builds to
break.


mprobst:
You might be curious:

tsickle emits a "forced load" when importing a module that only contains
types. This avoids a problem where TypeScript does not emit imports for
modules that are only imported for their types, which then means files
that only export types never get loaded. This is correct from TS'
perspective, but Closure Compiler in some run modes drops files that are
never imported from the set of inputs (prunes), and then later complains that
symbols defined in these files are never defined. The force load avoids
that problem. It's likely not to change program semantics either - it might affect load order for these files, but loading a module that only exports types is unlikely to change
runtime semantics.

test_files/import_alias/importer.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * @fileoverview added by tsickle
+ * @suppress {checkTypes} checked by tsc
+ */
+goog.module('test_files.import_alias.importer');
+var module = module || { id: 'test_files/import_alias/importer.ts' };
+var exporter_1 = goog.require('test_files.import_alias.exporter');
+const tsickle_forward_declare_1 = goog.forwardDeclare("test_files.import_alias.exporter");
```
rkirov: Odd that the forwardDeclare is emitted even when not used. Am I right that it doesn't affect the closure check that you are fixing here?

mprobst: It is maybe somewhat odd, but does simplify of code a bit – we just unconditionally forward declare any import, and then always use the forward declare (not the regular import) in all type declarations. That's consistent, and does not conflict with the require, and also doesn't trigger any of the Closure linter warnings that often cause issues with requires.

PR #804, opened 2018-05-14
Fix comment emit inside for arrow functions.

### LINKED ISSUE #802: Placing a comment in a single statement arrow function triggers ASI
tsickle produces incorrect JavaScript when it encounters a single statement arrow function with a line comment between the arrow and the statement. This bug prevents the body of the arrow function from executing. For example, this:
```
const foo = () =>
  // Print "bar"
  console.log("bar");
```
produces:
```
/**
 */
goog.module('_..index');var module = module || {id: 'index.js'};var /** @type {function(): void} */ foo = function () {
    // Print "bar"
    return 
    // Print "bar"
    console.log("bar");
};
```
Note that the comment was duplicated, and the duplication was placed on a line between `return` and the actual body of the function, triggering JavaScript ASI and exiting the function prematurely.

The following `tsconfig.json` was used with no additional `tsickle` options:
```
{
  "compilerOptions": {
    "target": "es5",
    "module": "commonjs",
    "experimentalDecorators": true
  }
}
```

May be related to [this issue](https://github.com/angular/tsickle/issues/57).

mprobst:
Previously, tsickle would synthesize a comment for expression bodies of
arrow functions:

    var x = () =>
        // comment
        1;

This would cause the comment to be emitted in between the return
statement and the expression after TS transform:

    var x = function() {
        // comment
        return
        // comment
        1;
    }

The first comment is TS correctly transforming the arrow function, the
second comment is tsickle re-emitting the synthesized comment, causing
Automatic Semicolon Insertion and thus breaking the code.

The fix is to just not do that - TypeScript emits comments in the
expression correctly, and tsickle already avoids modifying the comment
exactly because emitting comments in arrow functions is dangerous.

Fixes #802.


mprobst:
This is a workaround for https://github.com/Microsoft/TypeScript/issues/24096.

src/transformer_util.ts
```diff
@@ -286,6 +286,16 @@ export function visitNodeWithSynthesizedComments<T extends ts.Node>(
         context, sourceFile, node, sourceFile.statements,
         (node, stmts) => visitor(updateSourceFileNode(sourceFile, stmts) as ts.Node as T));
   } else {
+    // In arrow functions with expression bodies (`(x) => expr`), do not synthesize comment nodes
+    // that precede the body expression. When downleveling to ES5, TypeScript inserts a return
+    // statement and moves the comment in front of it, but then still emits any syntesized comment
+    // we create here. That can cause a line comment to be emitted after the return, which causes
+    // Automatic Semicolon Insertion, which then breaks the code. See arrow_fn_es5.ts for an
+    // example.
+    if (node.parent && (node as ts.Node).kind !== ts.SyntaxKind.Block &&
```
rkirov: OOC, why is the 'as ts.Node' cast needed. Looking at  `<T extends ts.Node>` and `node: T` that should be superfluous.

mprobst: I don't know, it's odd indeed. Without it, TS complains that `.kind` can only be [long list of `SyntaxKind` union type], but never `ts.SyntaxKind.Block`. It is though, and it can – TS must be incorrectly inferring a partial subtype here.

mprobst:
> Reading the function documentation "prevent the comment text ranges to be emitted, to allow changing these comments.", means that because that trick doesnt work for arrow functions we just reuse the TS mechanism at the cost of not being able to change the comment if we needed to. Is this correct?

Yes, that's correct. We don't emit comments for arrow functions like this anyway, because they don't quite work and can cause issues with semicolon insertion (e.g. when returning an arrow function we'd also introduce an incorrect wrap). So this doesn't regress anything that worked previously, but might be something we should revisit once the issue on the TS side is fixed (TS 2.9.1).

PR #808, opened 2018-05-16
Add web animations typings to closure_externs.js.

LucasSloan:
The typings for web animations were just added to the typescript standard library.  They partially replace the DefinitelyTyped typings for web animations.  The closure platform externs don't include web animations, which was previously fine, since tsickle produced the externs from the DefinitelyTyped typings.  With 2.8, many of the typings are in the standard library, so closure_externs.js needs to contain the corresponding externs.


src/closure_externs.js
```diff
@@ -78,3 +78,168 @@ var PromiseConstructorLike;
 /** @typedef {function(new:symbol)} */
 var SymbolConstructor;
```
rkirov: Add some comment on what belongs here - "Types that are part of lib???.d.ts, but not in ???.externs". Fill out ??? exactly so that we can decide in the future whether we should be adding or removing types from there. Both TS and closure platform typings change, so we would need to update correspondingly here.

test_files/web_animation.untyped/web_animations.d.ts
```diff
@@ -0,0 +1,72 @@
+// Type definitions for web-animations-js 2.2
```
rkirov: Does this test have to have all web-animations? Can you prune it to a more minimal part that is tested against closure_externs.js

rkirov: Also this needs a comment on what is it testing.

src/closure_externs.js
```diff
@@ -78,3 +78,168 @@ var PromiseConstructorLike;
...
 
+/**
+ * @constructor
+ * @param {(undefined|!AnimationEffectReadOnly)=} effect
+ * @param {(undefined|!AnimationTimeline)=} timeline
+ */
+var Animation = function(effect, timeline) {}
```
mprobst: I'm not sure this belongs into tsickle. The other externs we have here are mostly for fundamental platform builtins like `Symbol` and friends (and HTMLTableCell etc maybe shouldn't be here).

Web animations seems pretty orthogonal, and also just a ton of code. Can you consider what it'd take to find a better home for this, e.g. in a different internal location that you'd add to the default set of included externs?

mprobst:
This is no longer needed, right @LucasSloan?

PR #809, opened 2018-05-16
 fix(externs): emit only the local file name in a comment.

mprobst:
Otherwise, depending on the build system, this might include a long and non-stable path, causing excessive recompiles due to non-deterministic emits.


src/tsickle.ts
```diff
@@ -1745,7 +1745,8 @@ class ExternsWriter extends ClosureRewriter {
         this.writeExternsTypeAlias(node as ts.TypeAliasDeclaration, namespace);
         break;
       default:
-        this.emit(`\n/* TODO: ${ts.SyntaxKind[node.kind]} in ${namespace.join('.')} */\n`);
+        const locationStr = namespace.join('.') || path.basename(node.getSourceFile().fileName);
```
rkirov: Relativizing the path to google3 dir would be more informative than basename, but it might be hard to do.

PR #812, opened 2018-05-17
Convert all decorator downleveling to transformers.

mprobst:
Previously, we migrated the class level decorator conversion to transformers. This finishes the migration, and moves all decorator downleveling into the (now renamed) `decorator_downlevel_transformer`.


src/closure_externs.js
```diff
@@ -75,6 +75,6 @@ var PromiseConstructor;
 /** @typedef {function(new:Promise, function(function(*=), function(*=)))} */
 var PromiseConstructorLike;
 
-/** @typedef {function(new:symbol)} */
+/** @typedef {?} */
```
vikerman: Why this change?

PR #815, opened 2018-05-24
rel: 0.29.0

mprobst:


LucasSloan:
Our versioning scheme is 0.[TypescriptVersion].[number of releases since last typescript upgrade].  So this should bump the version to 0.28.1.

mprobst:
Was that really intentional? I didn't know that, and it's not documented in `README.md#Releasing`. The version is also already published I'm afraid.

Regardless of that, I'm also not sure that's a good idea. In this version, I've made substantial changes to tsickle internals and its emit since the last version, including some API breaking changes. So chances are that this will break some users, and they should not silently be upgraded to it if they have a patch level floating version dependency.

In general, I don't think we can or should even try to guarantee that tsickle will stay compatible during a TS minor version within the meaning of patch level bumps in semver.

LucasSloan:
@alexeagle could you weigh in here?

alexeagle:
Martin is right, if there are breaking changes, you update the second number when starting with 0. (the second number becomes the major)

0.28.1 should not have breaking changes from 0.28.0

I think this is the first time we had to make breaking changes not on the release schedule of typescript. If we had thought about it more, we could have factored this into the decision for when to make those changes, vs. when to add TS 2.9 support.

The numbering scheme was a convenience for us to keep track of what version of TS we support. I think we should just add TS 2.9 support now that it's released today, and cut 0.29.1, and let people figure it out. Note that our peer dep is the "official" way of saying the range we support https://github.com/angular/tsickle/blob/master/package.json#L21 so it's not necessary for the version number to do it too - just convenient.

PR #817, opened 2018-05-26
Handle the re-export edge case where typedef is swallowed.

### LINKED ISSUE #816: TypeDefs are not re-exported properly
See [index.js](https://gist.github.com/ppham27/b851f0f84ffd322298bb248d9a395a7a#file-index-js) outputted by tsickle.

The correct export is
```
exports.Foo = tsickle_forward_declare_1.Foo;
```

ppham27:
Allows one to use these types in Closure from where they are re-exported. For example,

If we have the file `export.ts` and some interface `Foo` in `export_helper_3.ts`:
```
import {Foo} from './export_helper_3';
// Do stuff with Foo.
export {Foo};
```

A type definition for Foo should be preserved. Currently, its stripped because
there's no trailing newline written when the node `export {Foo};\n` is written.
Thus, the type isn't available in Closure.

Fixes #816

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1119,7 +1119,8 @@ class Annotator extends ClosureRewriter {
               (exp.sym.flags & ts.SymbolFlags.Value) === 0;
       if (!isTypeAlias) continue;
       const typeName = this.symbolsToAliasedNames.get(exp.sym) || exp.sym.name;
-      this.emit(`/** @typedef {${typeName}} */\nexports.${exp.name}; // re-export typedef\n`);
+      // Leading newline prevents the typedef from being swallowed.
+      this.emit(`\n/** @typedef {${typeName}} */\nexports.${exp.name}; // re-export typedef\n`);
     }
   }
 
--- a/test_files/export/export.js
+++ b/test_files/export/export.js
@@ -40,3 +40,14 @@ const tsickle_forward_declare_4 = goog.forwardDeclare("test_files.export.export_
 var type_and_value_1 = goog.require('test_files.export.type_and_value');
 exports.TypeAndValue = type_and_value_1.TypeAndValue;
 const tsickle_forward_declare_5 = goog.forwardDeclare("test_files.export.type_and_value");
+const tsickle_forward_declare_6 = goog.forwardDeclare("test_files.export.export_helper_3");
+goog.require("test_files.export.export_helper_3"); // force type-only module to be loaded
+/**
+ * @return {!tsickle_forward_declare_6.Foo}
+ */
+function createFoo() {
+    return { fooStr: 'fooStr' };
+}
+exports.createFoo = createFoo;
+/** @typedef {tsickle_forward_declare_6.Foo} */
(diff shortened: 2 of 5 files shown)


src/tsickle.ts
```diff
@@ -1120,7 +1120,12 @@ class Annotator extends ClosureRewriter {
               (exp.sym.flags & ts.SymbolFlags.Value) === 0;
       if (!isTypeAlias) continue;
```
mprobst: This is actually on purpose, we might need to think a bit about how the right solution should look.

In TypeScript, types are not backed by values. That is, there is no runtime value for the `Foo` in `export type Foo = number;`. When we translate to Closure, some types become values (e.g. interfaces), but some do not (e.g. typedefs).

This is orthogonal to "Ambient". Ambient is something that in the current compilation unit is declared, but not defined - but if there's a `declare class Foo {}`, then there must still be an actual runtime symbol for `Foo` around.

The previous code was attempting to export both type aliases (typedefs in Closure) *and* interfaces using the same syntax, by exporting a typedef that's an alias to the underlying type. In the code below, `Bar` could be either an interface or a typedef:

```
/** @typedef {tsickle_forward_declare_1.Bar} */	+/** @const */
exports.Bar; // re-export typedef
```

Closure (AFAIR) does not support assigning values into `typedef`s, they are supposed to go undefined. So when you're accessing `tsickle_forward_declare_1.Bar` to assign into `Bar`, if there's no value backing it, we might be accessing an undefined value, but worse, create a symbol that's of unknown type to Closure.

Do I understand you correctly that the exported interface does not work for you? In the example that you give, the code contains:

```
// ...
exports.Foo; // re-export typedef
```

Is the problem maybe rather that the `/** @typedef {!tsickle_forward_declare_1.Foo} */` comment that's supposed to go before it is swallowed for some reason?

ppham27: Hi Martin,

Thanks for the reply. Yes if the `typedef` was preserved, I'd be happy, too. It wasn't easy to figure out why that `typedef` is swallowed, though, so just exporting the value seemed like an easier solution.

Any idea on why the `typedef` is swallowed? I'd be happy to submit that fix if you could point me in the right direction.

ppham27:
Hi @mprobst, I've reworked this change to make it so that the `@typedef` annotation isn't swallowed.

PTAL

src/tsickle.ts
```diff
@@ -1119,7 +1119,8 @@ class Annotator extends ClosureRewriter {
...
       const typeName = this.symbolsToAliasedNames.get(exp.sym) || exp.sym.name;
-      this.emit(`/** @typedef {${typeName}} */\nexports.${exp.name}; // re-export typedef\n`);
+      // Leading newline prevents the typedef from being swallowed.
```
mprobst: Load bearing leading whitespace. Sigh. But this is the right fix, until we port all of this to transformers.

mprobst:
I'm running this through the global test infrastructure against our internal repo. Will merge once that comes back (hopefully successfully).

PR #820, opened 2018-05-31
Update to rules_typescript 0.15.0 and tagged rules_webtesting 0.2.0

gregmagolan:


gregmagolan:
FYI: rules_typescript 0.14.0 was broken by a change in the yarn registry that broke an internal yarn.lock file. Good to update the 0.15.0

mprobst:
Thanks Greg!

PR #823, opened 2018-06-04
drop use of the google-closure-compiler npm module

evmar:
This module depends on a bunch of other npm modules (like vinyl
and so on) that we don't use, which makes it difficult to run
inside of Google.

Instead, rewrite just the tiny bit of the compiler API that we want.

Note: this keeps the google-closure-compiler npm dep in package.json
just for access to the compiler .jar file.  Users can override which
jar to use by setting compiler.jarPath.

### DIFF
--- a/test/closure.ts
+++ b/test/closure.ts
@@ -1,0 +1,85 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * @fileoverview Wrapper around shelling out to the Java Closure Compiler.
+ *
+ * This is the same functionality as the google-closure-compiler npm module,
+ * but to make things easy to use in Google's internal environment where
+ * the dependencies of that module are unavailable, we don't use any of that
+ * module's JavaScript code but only for compiler.jar.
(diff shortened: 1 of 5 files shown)


evmar:
(I believe this change will make it possible for us to hack things up to run the e2e test against google3)

test/closure.ts
```diff
@@ -0,0 +1,59 @@
+import * as child_process from 'child_process';
```
mprobst: Add a file overview comment explaining what this does, and why we're using it in place of the NPM module?

test/closure.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+/** The type of compilation results, containing exit code and output. */
+export interface Result {
+  exitCode: number;
+  stdout: string;
+  stderr: string;
+}
+
+export class Compiler {
```
mprobst: does this class pull its own weight, or should it just be a function?

test/closure.ts
evmar: Done.

test/closure.ts
evmar: I was originally trying to match to original API, then I diverged, so I may as well go all the way.  Done.

evmar:
Ping!

mprobst:
@evmar still LGTMs.

PR #824, opened 2018-06-11
use bazel build rules from npm

evmar:
This lets tsetse use its own version of tsutils, insulating
it from the tsutils pulled in by e.g. tslint.

I also regenerated yarn.lock.


evmar:
This doesn't work on Travis I guess.

PR #827, opened 2018-06-12
drop tsickle.formatDiagnostics

evmar:
TypeScript now exposes its own formatDiagnostics function, so we no
longer need to export one.  In particular, to correctly format
diagnostics you need a FormatDiagnosticsHost to resolve paths against.

Changing this required rebaselining some tests.


evmar:
So... much... yak... hair...

src/tsickle.ts
```diff
@@ -84,21 +84,30 @@ export let closureExternsBlacklist: string[] = [
   'WorkerGlobalScope',
 ];
 
+/**
+ * @deprecated Use TypeScript's ts.formatDiagnostics() or
+ *    formatDiagnosticsWithColorAndContext() directly instead.
+ */
 export function formatDiagnostics(diags: ReadonlyArray<ts.Diagnostic>): string {
```
mprobst: Consider deleting?

PR #828, opened 2018-06-12
Bump the TypeScript version to 2.9

alippai:


alippai:
Tried to upgrade as https://github.com/Microsoft/TypeScript/wiki/API-Breaking-Changes didn't show any changes, however it seems there are both in types and behavior. I'm not familiar with the TS Compiler API, so I can't proceed :/

evmar:
That has been our typical experience with upgrading TypeScript versions.  Unfortunately the only answer is to debug and fix it.  We'll probably take on upgrading to 2.9 soon.

alippai:
Updating to TS version 2.9.2 didn't help the regression

alippai:
I think @mprobst added the upgrade today :)

mprobst:
I did, and forgot to reference the right issue here, sorry and thanks for
noticing!

PR #830, opened 2018-06-14
clarify handling of index importing

evmar:
Minor code change: make the 'is in node_modules' check resilient
to situations where the entire tsickle source tree is contained
within a node_modules directory.

To make that change, I refactored the function to add a lot more
comments, because I had trouble following the existing logic.

### DIFF
--- a/src/googmodule.ts
+++ b/src/googmodule.ts
@@ -134,18 +134,51 @@ const TS_EXTENSIONS = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
  * Workaround for https://github.com/Microsoft/TypeScript/issues/12597
  */
 export function resolveIndexShorthand(
-    host: {options: ts.CompilerOptions, host: ts.ModuleResolutionHost}, fileName: string,
-    imported: string): string {
-  const resolved = ts.resolveModuleName(imported, fileName, host.options, host.host);
+    {options, host}: {options: ts.CompilerOptions, host: ts.ModuleResolutionHost},
+    pathOfImportingFile: string, imported: string): string {
+  // The strategy taken here is to use ts.resolveModuleName() to resolve the import to
+  // a specific path, and then if that path is different than the path that was
+  // asked for, we assume it was an index import and construct a new import statement.
+  //
+  // We need to be careful about paths -- `pathOfImportingFile` is the absolute path to the
+  // source file, while `imported` may be a path relative to that.
+
+  const resolved = ts.resolveModuleName(imported, pathOfImportingFile, options, host);
   if (!resolved || !resolved.resolvedModule) return imported;
   const requestedModule = imported.replace(TS_EXTENSIONS, '');
   const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(TS_EXTENSIONS, '');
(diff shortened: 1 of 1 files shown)


src/googmodule.ts
```diff
@@ -134,18 +134,50 @@ const TS_EXTENSIONS = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
  * Workaround for https://github.com/Microsoft/TypeScript/issues/12597
  */
 export function resolveIndexShorthand(
-    host: {options: ts.CompilerOptions, host: ts.ModuleResolutionHost}, fileName: string,
-    imported: string): string {
-  const resolved = ts.resolveModuleName(imported, fileName, host.options, host.host);
+    {options, host}: {options: ts.CompilerOptions, host: ts.ModuleResolutionHost},
+    pathOfImportingFile: string, imported: string): string {
+  // The strategy taken here is to use ts.resolveModuleName() to resolve the import to
```
rkirov: why not make this jsdoc?

src/googmodule.ts
```diff
@@ -134,18 +134,50 @@ const TS_EXTENSIONS = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
...
   if (!resolved || !resolved.resolvedModule) return imported;
   const requestedModule = imported.replace(TS_EXTENSIONS, '');
   const resolvedModule = resolved.resolvedModule.resolvedFileName.replace(TS_EXTENSIONS, '');
-  if (resolvedModule.indexOf('node_modules') === -1 &&
-      requestedModule.substr(requestedModule.lastIndexOf('/')) !==
-          resolvedModule.substr(resolvedModule.lastIndexOf('/'))) {
-    imported = './' + path.relative(path.dirname(fileName), resolvedModule).replace(path.sep, '/');
+
+  // Check if the resolution went into node_modules.
+  // Note that `resolved` has an attribute isExternalLibraryImport that is documented
```
rkirov: I was confused what `resolved` meant, but then I saw its the name of the local var. Would it make sense to make it the name of the interface that is returned from ts.resolveModuleName?

src/googmodule.ts
```diff
@@ -134,18 +134,50 @@ const TS_EXTENSIONS = /(\.ts|\.d\.ts|\.js|\.jsx|\.tsx)$/;
...
+    // a rewrite anyway.
+    return imported;
+  }
+
+  // If we get here, it seems that the import resolved to somewhere else.
+  // Construct a new import path.  Note that it must be relative to the original
+  // filename, and also that path.relative() randomly uses the cwd if any of
+  // its arguments aren't themselves absolute.
+  return './' +
+      path.relative(path.dirname(pathOfImportingFile), resolvedModule).replace(path.sep, '/');
```
rkirov: path.sep replacement is odd here, because above you do lastIndexOf('/') calculations.

src/googmodule.ts
evmar: Good idea, done.

src/googmodule.ts
evmar: Yeah.  :(  Preserving existing behavior, I bet this doesn't work on Windows anyway.

PR #833, opened 2018-06-14
clean up use of TypeScript helpers

evmar:
TypeScript used to have 'helpers' for builtins like __decorate.
They eventually moved to the 'tslib' library.

Clean up tsickle to match the new tslib world.
We accidentally had two copies of the helpers; delete them both,
and check in the current version of our tslib modification.


evmar:
Note to self: this fails in our tests for older TypeScript versions, because the directory layout is like:

```
/
  /tsconfig.json
  /node_modules
  /test
  /test_ts24
  /test_ts24/tsconfig.json
  /test_ts24/node_modules
```

And somehow when the `//test_ts24:unit_test` test runs (which sources files out of `/test`) it doesn't have a copy of `node_modules` in the right place in the runfiles, causing tsc to be unable to find `tslib`.

evmar:
This only changes files in the tsickle test suite, so I think it's safe to sync to g3.

PR #834, opened 2018-06-15
Only emit forwardDeclares for exported modules.

mprobst:
Previously, tsickle would emit a forwardDeclare even if it did not emit
an export statement for the referenced module. That'd cause a module
to be forward declared that was never imported, which then breaks the
build on the Closure side, as the forward declare is never satisfied.

### DIFF
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -305,6 +305,7 @@ export function addDiffMatchers() {
   // tslint:disable:no-require-imports
   jasmine.addMatchers(require('jasmine-diff')(jasmine, {
     colors: true,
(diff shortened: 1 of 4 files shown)


test/test_support.ts
```diff
@@ -305,6 +305,7 @@ export function addDiffMatchers() {
   // tslint:disable:no-require-imports
   jasmine.addMatchers(require('jasmine-diff')(jasmine, {
     colors: true,
+    inline: true,
```
evmar: I notice that the diff output of this library is less satisfying than the old one, sigh.

PR #836, opened 2018-06-19
Port tsickle to TypeScript 2.9

mprobst:
Please see the individual commits, which should be reviewable in isolation.


src/tsickle.ts
```diff
@@ -1967,7 +1967,15 @@ export function emitWithTsickle(
           content = combineSourceMaps(program, fileName, content);
         }
         if (host.addDtsClutzAliases && isDtsFileName(fileName) && sourceFiles) {
-          content = addClutzAliases(fileName, content, sourceFiles, typeChecker, host);
+          // Only bundle emits pass more than one source file for .d.ts writes. Bundle emits however
+          // are not supported by tsickle, as we cannot annotate them for Closure in any meaningful
+          // way anyway.
+          if (!sourceFiles || sourceFiles.length > 1) {
```
evmar: Are upstream d.ts files (including lib.d.ts) not source files here?

src/tsickle.ts
```diff
@@ -2013,189 +2021,75 @@ export function emitWithTsickle(
...
+  });
+  if (!localExports.length) return dtsFileContent;
+
+  const moduleName = host.pathToModuleName('', sourceFile.fileName);
+  const clutzModuleName = moduleName.replace(/\./g, '$');
+
+  // Clutz might refer to the name in two different forms (stemming from goog.provide and
+  // goog.module respectively).
+  // 1) global in clutz:   ಠ_ಠ.clutz.module$contents$path$to$module_Symbol...
+  // 2) local in a module: ಠ_ಠ.clutz.module$exports$path$to$module.Symbol ..
```
evmar: Is the module_Symbol vs module.Symbol distinction here important?  I don't see the underscores in the Clutz test suite.

src/tsickle.ts
```diff
@@ -2013,189 +2021,75 @@ export function emitWithTsickle(
...
-    }
-    constrainedParamList.push(constrainedParam);
-    paramList.push(param.name.getText());
-  }
-
-  const params = `<${paramList.join(',')}>`;
-  const paramsWithContraint = `<${constrainedParamList.join(',')}>`;
-
-  return {params, paramsWithContraint};
-}
```
evmar: Overall this is a lot of deleted code (good!) but it probably would be useful to have the author of this code (Lucas? Rado?) take a look too.

test_files/import_empty/import_empty.js
```diff
@@ -1,7 +1,9 @@
 /**
- * @fileoverview added by tsickle
- * @suppress {checkTypes} checked by tsc
+ * @fileoverview Make sure tsickle does not crash on empty imports.
+ * @suppress {checkTypes,extraRequire} checked by tsc
```
evmar: BTW, I have a pending CL to make the suppressions match on all transformation codepaths.

src/tsickle.ts
```diff
@@ -283,7 +283,10 @@ abstract class ClosureRewriter extends Rewriter {
       }
       // Merge the parameters into a single list of merged names and list of types
       const sig = typeChecker.getSignatureFromDeclaration(fnDecl);
-      if (!sig) throw new Error(`invalid signature ${fnDecl.name}`);
+      if (!sig || !sig.declaration) throw new Error(`invalid signature ${fnDecl.name}`);
+      if (sig.declaration.kind === ts.SyntaxKind.JSDocSignature) {
+        throw new Error(`JSDoc signature ${fnDecl.name}`);
+      }
```
evmar: Should we be using this for all our JSDoc parsing now?  I am confused about the circumstances where TS obeys JSDoc in a TS file.

src/tsickle.ts
mprobst: @LucasSloan PTAL.

src/tsickle.ts
mprobst: I find stuff like:

`src/test/java/com/google/javascript/clutz/partial/override.d.ts:  class module$contents$override_Base extends module$contents$override_Base_Instance {`

I think this is an artifact of Closure's `goog.provide` vs `goog.module` stuff, but tbh I don't quite know which is which. I'm aping the previous emit, and it seems to work in google3.

src/tsickle.ts
mprobst: they are not, this is just the single file that the emit happens for.

src/tsickle.ts
mprobst: Not sure. The `JSDocSignature` happens AFAIU when TypeScript resolves an import from TypeScript to a function in JS that has JSDoc declared. We'll probably deal with that more in the future, can revisit then.

PR #838, opened 2018-06-19
drop test for importHelpers=false

evmar:
To reduce our configuration proliferation, don't even bother
testing for importHelpers=false.  We only support importHelpers=true.


evmar:
I'm not sure why we have this test, but it doesn't seem useful to me.

mprobst:
Probably because it was the status quo before `importHelpers` existed. Yes, let's drop it. Bonus points for adding a check in `tsickle.ts` that `importHelpers` is set, and fail if it isn't.

PR #840, opened 2018-06-20
rearrange @fileoverview handling

evmar:
The fileoverview logic had three different cases where it added to
the suppress list, and extraRequire was only included in one of the
three.  I want to add another entry to our suppression list, so I'll
start by refactoring this so that suppressions are handled in one place.

This rebaselines ~every test because we now consistently suppress
extraRequire.

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -29,6 +29,43 @@ export function isClosureFileoverviewComment(text: string) {
 }
 
 /**
+ * Given a parsed @fileoverview comment, ensures it has all the attributes we need.
+ * This function can be called to modify an existing comment or to make a new one.
+ *
+ * @param tags Comment as parsed list of tags; modified in-place.
+ */
+function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
+  // Ensure we start with a @fileoverview.
+  if (!tags.find(t => t.tagName === 'fileoverview')) {
+    tags.splice(0, 0, {tagName: 'fileoverview', text: 'added by tsickle'});
+  }
+
+  // Find or create a @suppress tag.
+  // Closure compiler barfs if there's a duplicated @suppress tag in a file, so the tag must
+  // only appear once and be merged.
+  let suppressTag = tags.find(t => t.tagName === 'suppress');
+  let suppressions: Set<string>;
+  if (suppressTag) {
+    suppressions = new Set((suppressTag.type || '').split(',').map(s => s.trim()));
+  } else {
+    suppressTag = {tagName: 'suppress', text: 'checked by tsc'};
+    tags.push(suppressTag);
+    suppressions = new Set();
+  }
+
+  // Ensure our suppressions are included in the @suppress tag:
+  // 1) Suppress checkTypes.  We believe the code has already been type-checked by TypeScript,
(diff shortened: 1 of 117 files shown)


evmar:
This is a pre-refactoring for my 'if (false)' change for class members.

src/fileoverview_comment_transformer.ts
```diff
@@ -28,6 +28,43 @@ export function isClosureFileoverviewComment(text: string) {
   return current !== null && current.tags.some(t => FILEOVERVIEW_COMMENT_MARKERS.has(t.tagName));
 }
 
+/**
+ * Given a parsed @fileoverview comment, ensures it has all the attributes we need.
+ * This function can be called to modify an existing comment or to make a new one.
+ *
+ * @param tags Comment as parsed list of tags; modified in-place.
+ */
+function modifyFileoverviewComments(tags: jsdoc.Tag[]) {
```
mprobst: `augmentFileoverviewComments`?

src/fileoverview_comment_transformer.ts
```diff
@@ -28,6 +28,43 @@ export function isClosureFileoverviewComment(text: string) {
...
+  // Ensure we start with a @fileoverview.
```
mprobst: FWIW the order I think doesn't matter, but this is cleaner.

src/fileoverview_comment_transformer.ts
```diff
@@ -28,6 +28,43 @@ export function isClosureFileoverviewComment(text: string) {
...
+  }
+
+  // Ensure our suppressions are included in the @suppress tag:
+  // 1) Suppress checkTypes.  We believe the code has already been type-checked by TypeScript,
+  // and we cannot model all the TypeScript type decisions in Closure syntax.
+  suppressions.add('checkTypes');
+  // 2) Suppress extraRequire.  We remove extra requires at the TypeScript level, so any require
+  // that gets to the JS level is a load-bearing require.
+  suppressions.add('extraRequire');
+  suppressTag.type = Array.from(suppressions.values()).sort().join(',');
```
mprobst: FYI `Set` in JavaScript is a linked hash set, `.values()` iterates in insertion order, so you don't have to sort here (unless you generally want these tags to be emitted in order).

src/fileoverview_comment_transformer.ts
evmar: This also kept the test churn down (it would put the @fileoverview second on every test otherwise).

src/fileoverview_comment_transformer.ts
evmar: The set could have already contain extraRequire (due to an existing suppress tag) before our add, so it would come out first in that case.  I guess it'd still be stable in that case.  Hm, maybe it's better to have our additions to the list always come last?

mprobst: I think it really doesn't matter in either way.

src/fileoverview_comment_transformer.ts
evmar: Done.

mprobst:
LGTM.

evmar:
[finally got a green build in google3 with this]

mprobst:
Woohoo!

PR #841, opened 2018-06-21
conditionally support an explicit path to tslib

evmar:
To run this test suite inside Google, we want to be able to pass
in the path to tslib.d.ts explicitly.  Add support for this that
is effectively disabled in the public tsickle test suite.

### DIFF
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -18,6 +18,9 @@ import {BasicSourceMapConsumer, sourceMapTextToConsumer} from '../src/source_map
 import * as tsickle from '../src/tsickle';
 import {getCommonParentDirectory} from '../src/util';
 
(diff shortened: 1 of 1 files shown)


test/test_support.ts
```diff
@@ -32,6 +35,8 @@ export const baseCompilerOptions: ts.CompilerOptions = {
   allowJs: false,
   importHelpers: true,
   noEmitHelpers: true,
+  baseUrl: '.',
+  paths: tslibPath ? {'tslib': [tslibPath]} : undefined,
```
rkirov: add a comment on the lookup mechanism for 'tslib'. It was a bit magic for me.

evmar: Done.

PR #845, opened 2018-06-26
sort symbols before emitting them

evmar:
TypeScript 2.8 and 2.9 differ on the order of symbols returned
by getExportsOfModule().  To make the tests stable across these
versions, sort the symbols before iterating them.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -2020,6 +2020,13 @@ export function emitWithTsickle(
   };
 }
 
+/** Compares two strings and returns a number suitable for use in sort(). */
+function stringCompare(a: string, b: string): number {
+  if (a < b) return -1;
+  if (a > b) return 1;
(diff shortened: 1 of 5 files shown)


src/tsickle.ts
```diff
@@ -2021,6 +2021,13 @@ export function emitWithTsickle(
   };
 }
 
+/** Compares two strings and returns a number suitable for use in sort(). */
+function stringCompare(a: string, b: string): number {
```
mprobst: our locale should always be C, so `a.localeCompare(b)` should do it here, right? But this is probably safer.

evmar: I considered using that, and kinda puzzled over whether locales could matter, and eventually decided just to not worry about it.  You might reasonably wonder why there is no ordinary string compare function, sigh.

PR #852, opened 2018-07-02
Only emit @extends for classes without `extends`.

mprobst:
tsickle emits an `@extends` for classes implementing classes, to help
Closure track the type hierarchy. However for classes that have an
actual `class X extends Y` clause, this can cause Closure to get
confused about the type hierarchy, leading to incorrect optimization.

This change prevents emitting both `@extends` and `extends`, avoiding
the problem.

### DIFF
--- a/test_files/extend_and_implement/extend_and_implement.js
+++ b/test_files/extend_and_implement/extend_and_implement.js
@@ -1,0 +1,30 @@
+/**
+ *
+ * @fileoverview Reproduces a problem where tsickle would emit "\\@extends
+ * {ClassInImplements}", conflicting the ES6 extends syntax, leading to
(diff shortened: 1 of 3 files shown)


test_files/extend_and_implement/extend_and_implement.js
```diff
@@ -0,0 +1,30 @@
+/**
+ *
```
evmar: I wonder why we get this extra whitespace here.

mprobst: Huh. Maybe we parse @extends incorrectly? I now also notice we emit `\\@extends`, not `\@extends`. I'll take a look later.

PR #849, opened 2018-06-29
use es6 form when generating export statement

evmar:
This produces equivalent output when targeting goog.module
(because it becomes an exports.foo = foo line) so no tests are
affected.

But when tsickle is emitting ES6, this stays as an export {}
statement, which is the proper output for that.

Note: apparently tsickle has no tests covering the ES6 output
mode(!), or this would've been caught earlier.

Fixes #848.

PR #853, opened 2018-07-03
Avoid typedef name conflicts.

mprobst:
Introducing a new name in a scope can cause several different problems,
such as incorrect references or accidentally duplicated exports.

This change introduces a name mangling scheme for type aliases, so that
the newly introduced value symbols do not (likely) cause conflicts.

This problem in tsickle was exposed, but not caused, by #849. The issue
always existed, it was just suppressed by users and went unnoticed
(partially due to untyped mode). The exports change caused an
unsuppressible error that this change should fix, together with the
previous ones.


src/tsickle.ts
```diff
@@ -1452,29 +1453,46 @@ class Annotator extends ClosureRewriter {
...
     // In the case of an export, we cannot emit a `export var foo;` because TypeScript drops exports
     // that are never assigned values (and Closure requires us to not assign values to typedef
     // exports).
     // tsickle must also still emit the `var` line above so that the type can be used within the
     // local module scope (Closure does not allow refering to "exports.Foo" within a module).
-    // With that, emit an additional "export {foo};" to export the type.
+    // With that, emit an additional "export {foo_mangled as foo};" to export the type.
     if (hasModifierFlag(node, ts.ModifierFlags.Export)) {
-      this.emit(`export {${typeName}};\n`);
+      this.emit(`export {${mangledTypeName} as ${typeName}};\n`);
```
evmar: Do we need to avoid this export in the case where the value has already been exported?

mprobst:
Actually I'm concerned about the behaviour of this with typedefs that are used before emit, e.g.:

```ts
function useFoo(f: Foo) {}
type Foo = number;
```

This actually doesn't cause a problem in a naive test, because `Foo` is emitted as `number` in `useFoo`s signature, but in the general case, this will cause problems as we might emit the type before we come up with the mangled name.

We could fix with two-pass (ugh) or determining the mangled name at the use site (also ugh, in particular with calls going from type-translator back into tsickle.ts).

I think I prefer #854, just entirely skip introducing a local variable, which should neatly side-step most of the problem.

mprobst:
Closing this now that #854 has landed.

PR #854, opened 2018-07-03
 Do not emit local type aliases.

mprobst:
This is an alternative fix for the problem also tackled in PR #853.


src/tsickle.ts
```diff
@@ -50,6 +50,11 @@ export interface AnnotatorHost {
    * If true, do not modify quotes around property accessors.
    */
   disableAutoQuoting?: boolean;
+  /**
+   * Annotator has to emit "exports.Foo" style statements if and only if also doing the googmodule
+   * emit transformation.
```
evmar: Why?  Is this just avoiding the bug, or is there some fundamental reason?

(Also maybe this comment should open with something like "true if we want googmodule output" or some such)

src/tsickle.ts
```diff
@@ -1452,30 +1457,30 @@ class Annotator extends ClosureRewriter {
...
     // In the case of an export, we cannot emit a `export var foo;` because TypeScript drops exports
     // that are never assigned values (and Closure requires us to not assign values to typedef
     // exports).
     // tsickle must also still emit the `var` line above so that the type can be used within the
     // local module scope (Closure does not allow refering to "exports.Foo" within a module).
-    // With that, emit an additional "export {foo};" to export the type.
-    if (hasModifierFlag(node, ts.ModifierFlags.Export)) {
-      this.emit(`export {${typeName}};\n`);
-    }
+    // With that, emit an additional "export {foo_mangled as foo};" to export the type.
```
evmar: This comment ... is left from the previous change I think?

test_files/promisectorlike/promisectorlike.js
```diff
@@ -5,7 +5,7 @@
 goog.module('test_files.promisectorlike.promisectorlike');
 var module = module || { id: 'test_files/promisectorlike/promisectorlike.ts' };
 /**
- * @param {PromiseConstructorLike} ctorLike
+ * @param {function(new: (!PromiseLike<?>), function(function((undefined|?|!PromiseLike<?>)=): void, function(?=): void): void): ?} ctorLike
```
evmar: Can't this alias expansion potentially cause us to reference types that aren't in scope?

src/tsickle.ts
mprobst: You mean as opposed to just emitting an `export {X};`?

The problem with just emitting `export {X};` is if `X` is a type, not a value, TypeScript will not convert it to an `exports.X` statement. If we make `X` a value, we're causing all kinds of problems with newly introduced type/value conflicts, and also the problem with the import/local var/export conflicts leading to a duplicate import.

Basically what we're doing here is instead of introducing a new local value symbol and having TS create the export for it, we're doing the `exports.` emit ourselves to avoid having to create the value symbol.

The long term for this code btw is to move to `googmodule`, but for that we'll have to figure out how to transport the `symbolToAliasMap` all the way over there.

test_files/promisectorlike/promisectorlike.js
mprobst: No, because we have `ensureSymbolDeclared` that will emit a `forwardDeclare` if necessary. We currently have both code paths, the ensure declared that introduces forward declares, and then sometimes when TS creates a symbol with an aliasSymbol on it, we use that - but we cannot predict or force that, so it's pretty random.

src/tsickle.ts
evmar: I think I don't understand the comment, it's saying one of
a) has to emit exports.Foo rather than nothing?
b) has to emit exports.Foo rather than export {Foo}?

and also I don't understand why it's conditioned on googmodule

src/tsickle.ts
mprobst: Yeah, fixed.

src/tsickle.ts
mprobst: You're right. Essentially we cannot just stay with an ES6 export, because TS drops those. We can fix that by introducing a local variable (which would get exported), but we cannot do that because of the problems I'm fixing here.

So what remains is emitting an export "manually". I've keyed that on `googmodule`, but I think the better thing to do is keying it on whether we'll emit `CommonJS`. If we don't do that, we don't know how to emit the faux export, so we should not do it - but also, if we're not doing `CommonJS`, the downstream user cannot use `googmodule`, which means they cannot import the type from Closure anyway, so there's no point in doing this at all.

src/tsickle.ts
```diff
@@ -1452,30 +1452,40 @@ class Annotator extends ClosureRewriter {
     this.emit(`${namespace.join('.')};\n`);
   }
 
+  private typedefCounter = 1;
```
evmar: Doc this.

src/tsickle.ts
```diff
@@ -1452,30 +1452,40 @@ class Annotator extends ClosureRewriter {
...
     // namespaces, the two emits would conflict if tsickle emitted both.
     const sym = this.mustGetSymbolAtLocation(node.name);
     if (sym.flags & ts.SymbolFlags.Value) return;
+    // Type aliases are always emitted as the resolved underlying type, so there is no need to emit
+    // anything, except for exported types.
+    if (!hasModifierFlag(node, ts.ModifierFlags.Export)) return;
+    // A pure ES6 export (`export {Foo}`) is insufficient, as TS does not emit those for pure types,
+    // so tsickle has to pick a module format. We're using CommonJS to emit googmodule, and code not
+    // using googmodule doesn't care about the Closure annotations anyway, so just skip emitting if
+    // the module target isn't commonjs.
```
evmar: This is good.

So much subtlety!

src/tsickle.ts
mprobst: Ugh, good catch, it's unused now.

PR #855, opened 2018-07-06
drop gulp in favor of our own check_format script

evmar:
The only reason we have gulp involved is for checking clang-format.
    I'd like to also check that README.md is formatted.  To do that
    we'll need more formatter logic.  Rather than add more gulp, I figured
    I'd go the other way and remove the gulp in this change first.
    
    PS: I wrote the checker as a JS program rather than a shell script to
    have some hope of this working on Windows.


evmar:
CC @alexeagle

evmar:
I have a followup ready that adds prettier to this for markdown files.

check_format.js
```diff
@@ -0,0 +1,41 @@
...
+  }
+  return true;
+}
+
+function main(args) {
+  const fix = args[0] === '--fix';
+  const sourceFiles = glob.sync('{*.js,src/**/*.[jt]s,test/**/*.[jt]s}');
+  for (const path of sourceFiles) {
+    if (!checkFile(path, fix)) {
+      console.error(`${path} not formatted; run 'node check_format.js --fix'`);
```
LucasSloan: The old version of this only printed one command if you needed to fix it.  Could you accumulated failing paths and print once?

check_format.js
```diff
@@ -0,0 +1,41 @@
...
+    fs.writeFileSync(path, formatted);
+  }
+  return true;
+}
+
+function main(args) {
+  const fix = args[0] === '--fix';
+  const sourceFiles = glob.sync('{*.js,src/**/*.[jt]s,test/**/*.[jt]s}');
+  for (const path of sourceFiles) {
+    if (!checkFile(path, fix)) {
```
LucasSloan: It's not clear that checkFile will overwrite the file if fix is true here.  I'd prefer if there was clear fix codepath and another clear check codepath.

check_format.js
evmar: The 'return' statement right after means this only prints once.  Did you mean you want the error message to list out which files are wrong?

check_format.js
evmar: It's a bit complicated by how we need to read the text of the file in to fix a file, so the check and fix codepaths are very close to one another (see the 'fix' flag in checkFile).  Maybe I can rename something to address this?

evmar:
Ping.

check_format.js
LucasSloan: I'd be happy with a checkFile() and a fixFile() function that both called checkAndFixFile() with a different boolean if you're concerned about duplication.

check_format.js
LucasSloan: I wasn't seeing the return, so I guess this is ok.  I just didn't want users to see 5 'to format, run COMMAND' messages in a row.

check_format.js
evmar: OK, PTAL

evmar:
Also rewrote the description based on your question.

check_format.js
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+const childProcess = require('child_process');
+const fs = require('fs');
+const glob = require('glob');
+
+const CLANG_FORMAT = 'node_modules/.bin/clang-format';
```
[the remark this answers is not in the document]
evmar: I looked into it and it exposes a function that involves callbacks rather than running sync (while amusingly also using a .existsSync within it, so not really async) which makes it hard to use.  I'll leave that for future work.

check_format.js
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Checks that a path is formatted.
+ * @return new text of file if it needs formatting or null otherwise.
+ */
+function checkFile(path) {
+  const source = fs.readFileSync(path, 'utf8');
+  const formatted = childProcess.execFileSync(CLANG_FORMAT, [path], {encoding: 'utf8'});
```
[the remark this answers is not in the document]
evmar: It doesn't.  It would've made all this easier.

alexeagle:
Note, I've tried to make it convenient to run the incremental formatter in a pre-commit hook, did you see https://github.com/angular/clang-format/blob/master/bin/check-clang-format.js ?

evmar:
I did!  But it seems to be a git pre-commit hook, and not usable as part of our build process (aka travis), right?

PR #859, opened 2018-07-09
use 'const' for module imports in es6 output mode

evmar:
Emit
  const x = goog.require(...)
instead of
  var x = goog.require(...)

Using 'const' here is semantically more correct (you shouldn't reassign
these anyway) and also forwards type aliases in some circumstances.


evmar:
Ahh, good catch!

In my experimentation with the original bug it also works to emit a `@const`, like

```
/** @const */
var mod_2 = mod_1;
```

maybe I should just always do that instead?

evmar:
The ES5<->ES6 thing is now fixed.  This includes rebaselines of only the subset of affected tests (the ES6 ones).

dsmilkov:
Any updates on this? Thanks!

evmar:
Oops, I somehow totally dropped this. I gotta test it again against google3 but then I will merge.

evmar:
This works in google3 (with some minor test adjustments), so I'm merging it.

PR #860, opened 2018-07-10
Use diff-match-patch for a basic diff matcher.

mprobst:
jasmine-diff is a much more general solution that (tries to) implement
object and array diffing, which tsickle does not use in its tests.

It's string-diff support turns out to be very limited. The diff
algorithm either does very basic unified diff, or a much too over-eager
inline diff. At the same time, jasmine-diff compares strings *after*
abbreviating them, causing incorrect diffs, or even error messages not
mentioning a diff at all.

This change adds a basic diff matcher based on diff-match-patch. It
still has a number of limitations, like not quite lining up line breaks
for removed and added lines. Even so, it's a substantial improvement
over jasmine-diff, and gets us back to actually readable test output.

from d0c1fcc

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,16 +20,17 @@
     "typescript": ">=2.4.2 <2.9"
   },
   "devDependencies": {
+    "@types/diff-match-patch": "^1.0.32",
     "@types/glob": "5.0.35",
     "@types/jasmine": "2.8.6",
     "@types/minimist": "1.2.0",
     "@types/mkdirp": "0.5.2",
     "@types/node": "^10.3.3",
     "clang-format": "1.2.2",
+    "diff-match-patch": "^1.0.1",
     "glob": "7.1.2",
     "google-closure-compiler": "20161024.3.0",
     "jasmine": "3.1.0",
-    "jasmine-diff": "^0.1.3",
     "tslint": "5.9.1",
     "typescript": "2.9.2"
   },
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -7,6 +7,7 @@
  */
 
 import * as assert from 'assert';
+import {DIFF_DELETE, DIFF_EQUAL, DIFF_INSERT, diff_match_patch as DiffMatchPatch} from 'diff-match-patch';
 import * as fs from 'fs';
 import * as glob from 'glob';
 import * as path from 'path';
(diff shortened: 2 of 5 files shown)


mprobst:
See attached screenshot for example output. `jasmine-diff` would often create a smorgasbord of diff chunks even for very obvious diffs.

screenshot from 2018-07-10 10-11-46

test/test_support.ts
```diff
@@ -299,17 +300,71 @@ export function getSourceMapWithName(
...
+  }
+  const dmp = new DiffMatchPatch();
+  dmp.Match_Distance = 0;
+  dmp.Match_Threshold = 0;
+  const diff = dmp.diff_main(expected, actual);
+  dmp.diff_cleanupSemantic(diff);
+  if (!diff.length) return {pass: true};
+  let message = '\x1B[0;0mstrings differ:\n';
+  for (const d of diff) {
+    switch (d[0]) {
```
evmar: Is this maybe a tuple?

for (const [diffType, text] of diff) {

test/test_support.ts
```diff
@@ -299,17 +300,71 @@ export function getSourceMapWithName(
...
+        message += '\x1B[90;42m' + d[1] + '\x1B[0;0m';
+        break;
+      default:
+        throw new Error('unexpected diff result: ' + d);
+    }
+  }
+  return {pass: false, message};
+}
+
+// Augment the global "jasmine.Matchers" type with out toEqualWithDiff function.
```
evmar: s/out/our/

package.json
```diff
@@ -20,12 +20,14 @@
...
     "@types/glob": "5.0.35",
     "@types/jasmine": "2.8.6",
     "@types/minimist": "1.2.0",
     "@types/mkdirp": "0.5.2",
     "@types/node": "^10.3.3",
     "clang-format": "1.2.2",
+    "diff-match-patch": "^1.0.1",
     "glob": "7.1.2",
     "google-closure-compiler": "20161024.3.0",
     "jasmine": "3.1.0",
```
evmar: delete jasmine-diff dep here?

test/test_support.ts
```diff
@@ -299,17 +300,71 @@ export function getSourceMapWithName(
...
+  if (!diff.length) return {pass: true};
+  let message = '\x1B[0;0mstrings differ:\n';
+  for (const d of diff) {
+    switch (d[0]) {
+      case DIFF_EQUAL:
+        message += d[1];
+        break;
+      case DIFF_DELETE:
+        // light gray on red.
+        message += '\x1B[37;41m' + d[1] + '\x1B[0;0m';
```
evmar: Super minor, but:

`\x1b[0;0m` can just be written `\x1b[0m`, the 0 resets everything. Here and below.

test/test_support.ts
mprobst: Done.

package.json
mprobst: Done.

test/test_support.ts
mprobst: Good point, done.

test/test_support.ts
mprobst: Done.

PR #862, opened 2018-07-11
Fully convert tsickle to TypeScript's transformer API.

mprobst:


mprobst:
This is not quite ready for review insofar as it doesn't delete the old code yet. The new code could be reviewed.

src/BUILD
```diff
@@ -1,29 +1,34 @@
...
 # We don't want to copy a file that's licensed under an external license into
 # our source tree, and this mechanism allows the :src library to import it.
 genrule(
     name = "typescript-2.4",
-    outs = ["typescript-2.4.d.ts"],
     srcs = ["//test_ts24:node_modules/typescript/lib/typescript.d.ts"],
+    outs = ["typescript-2.4.d.ts"],
     # see https://docs.bazel.build/versions/master/be/general.html#genrule_args
-    cmd = "cp $< $@"
+    cmd = "cp $< $@",
```
evmar: It's probably worth (for minimizing diff purposes) sumbitting buildifiler CLs separately first.

src/fileoverview_comment_transformer.ts
```diff
@@ -75,20 +75,48 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
...
     }
 
-    // Closure Compiler considers the *last* comment with @fileoverview (or @externs or @nocompile)
-    // that has not been attached to some other tree node to be the file overview comment, and
-    // only applies @suppress tags from it.
-    // AJD considers *any* comment mentioning @fileoverview.
+    // Closure Compiler considers the *last* comment with @fileoverview (or @externs or
+    // @nocompile) that has not been attached to some other tree node to be the file overview
+    // comment, and only applies @suppress tags from it. AJD considers *any* comment mentioning
+    // @fileoverview.
```
evmar: Looks like maybe you're rewrapping to 80.  Maybe it's worth doing that in a pre-pass?

src/jsdoc.ts
```diff
@@ -281,11 +369,12 @@ function serialize(
   if (tags.length === 0) return '';
   if (tags.length === 1) {
     const tag = tags[0];
-    if ((tag.tagName === 'type' || tag.tagName === 'nocollapse') &&
+    if ((tag.tagName === 'type' || tag.tagName === 'typedef' || tag.tagName === 'nocollapse') &&
```
evmar: Can submit this one separately.

src/rewriter.ts
```diff
@@ -228,9 +228,9 @@ export abstract class Rewriter {
 
 /** Returns the string contents of a ts.Identifier. */
 export function getIdentifierText(identifier: ts.Identifier): string {
-  // NOTE: the 'text' property on an Identifier may be escaped if it starts
-  // with '__', so just use getText().
-  return identifier.getText();
+  // NOTE: 'escapedText' on an Identifier may be escaped if it starts with '__'. The alternative,
+  // getText(), cannot be used on synthesized nodes, so unescape the identifier below.
+  return unescapeName(identifier.escapedText as string);
```
evmar: Separate

src/rewriter.ts
```diff
@@ -247,6 +247,6 @@ export function getEntityNameText(name: ts.EntityName): string {
  */
 export function unescapeName(name: string): string {
   // See the private function unescapeIdentifier in TypeScript's utilities.ts.
-  if (name.match(/^___/)) return name.substr(1);
+  if (name.startsWith('___')) return name.substring(1);
```
evmar: Separate

test_files/interface/interface_extends.js
```diff
@@ -8,20 +8,26 @@ var module = module || { id: 'test_files/interface/interface_extends.ts' };
  * @record
  */
 function ParentInterface() { }
-/** @type {number} */
-ParentInterface.prototype.x;
+if (false) {
```
evmar: Seems like something went bad in a merge here (?)

src/BUILD
mprobst: I'm heavily editing this file (adding new transformers, removing obsolete code) - I think it's not worth picking that apart for the 10 lines diff.

src/fileoverview_comment_transformer.ts
```diff
@@ -75,10 +64,38 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
...
+    const originalComments = ts.getLeadingCommentRanges(text, 0) || [];
+    if (!firstStatement) {
+      // In an empty source file, all comments are file-level comments.
+      fileComments = synthesizeCommentRanges(sf, originalComments);
+    } else {
+      // Search for the first comment split from the file with a \n\n. All comments before that are
+      // considered fileoverview comments, all comments after that belong to the next statement(s).
+      // If none found, comments remains empty, and the code below will insert a new fileoverview
+      // comment.
+      for (let i = 0; i < originalComments.length; i++) {
```
Kompwu: Please prefer `const` over `let`.

package.json
```diff
@@ -29,11 +29,12 @@
     "clang-format": "1.2.2",
     "diff-match-patch": "^1.0.1",
     "glob": "7.1.2",
-    "google-closure-compiler": "20161024.3.0",
+    "google-closure-compiler": "20180610.0.2",
```
evmar: Why do we need to change this?

src/externs.ts
```diff
@@ -0,0 +1,439 @@
...
+  // simpler to just blacklist it.
+  'ErrorConstructor',
+  'Symbol',
+  'WorkerGlobalScope',
+];
+
+/**
+ * generateExterns generates extern definitions for all ambient declarations in the given source
+ * file. It returns a string representation of the Closure JavaScript, not including the externs
+ * header.
```
evmar: What does 'externs header' mean here?

src/externs.ts
```diff
@@ -0,0 +1,439 @@
...
+   * we only want to emit the "\@record" for Foo on the first one.
+   */
+  function isFirstDeclaration(decl: ts.DeclarationStatement): boolean {
+    if (!decl.name) return true;
+    const sym = typeChecker.getSymbolAtLocation(decl.name)!;
+    if (!sym.declarations || sym.declarations.length < 2) return true;
+    return decl === sym.declarations[0];
+  }
+
+  function writeVariable(name: string, namespace: ReadonlyArray<string>, value?: string) {
```
evmar: Docs on this and the one below it.  I think it might be better name 'writeVariableStatement' since it incldues the 'var' and the semi.

evmar: Hm, I looked into this more and I guess this is just moved from the old code.  Still might be worth doing some minor fixes.

src/externs.ts
```diff
@@ -0,0 +1,439 @@
...
+  // interface into properties on Closure's Error object, but for now it's
+  // simpler to just blacklist it.
+  'ErrorConstructor',
+  'Symbol',
+  'WorkerGlobalScope',
+];
+
+/**
+ * generateExterns generates extern definitions for all ambient declarations in the given source
+ * file. It returns a string representation of the Closure JavaScript, not including the externs
```
evmar: Not for this CL, but do you think it's eventually worth using transformers here too?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+import {addSyntheticTrailingComment} from 'typescript';
+
+import {hasExportingDecorator} from './decorators';
+import * as googmodule from './googmodule';
+import * as jsdoc from './jsdoc';
+import {ModuleTypeTranslator} from './module_type_translator';
+import * as transformerUtil from './transformer_util';
+import * as ts from './typescript';
+
+export interface AnnotatorHost {
```
evmar: Doc. I think here 'Annotator' is the tsickle term for "thing that adds closure comments", right?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+      let tagName = decl.kind === ts.SyntaxKind.InterfaceDeclaration ? 'extends' : 'implements';
+
+      const sym = mtt.typeChecker.getSymbolAtLocation(impl.expression);
+      if (!sym) {
+        // It's possible for a class declaration to extend an expression that
+        // does not have have a symbol, for example when a mixin function is
+        // used to build a base class, as in `declare MyClass extends
+        // MyMixin(MyBaseClass)`.
+        //
+        // Handling this correctly is tricky. Closure throws on this
```
evmar: I was just poking in this in kythe too: https://github.com/google/kythe/commit/8881f354095d52fd900c503825b0b84d2a7276e8

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+            member as ts.MethodDeclaration | ts.GetAccessorDeclaration | ts.SetAccessorDeclaration);
+      }
+      // Non-abstract methods only exist on classes, and are handled in regular emit.
+    } else {
+      unhandled.push(member);
+    }
+  }
+
+  if (ctors.length > 0) {
+    const ctor = ctors[0];
```
evmar: Is this choice arbitrary?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+
+/** Flags that declare a field of the same name if set on a ctor parameter. */
+const FIELD_DECLARATION_MODIFIERS: ts.ModifierFlags = ts.ModifierFlags.Private |
+    ts.ModifierFlags.Protected | ts.ModifierFlags.Public | ts.ModifierFlags.Readonly;
+
+/**
+ * emitMemberTypes emits the type annotations for members of a class. It's necessary in the case
+ * where TypeScript syntax specifies there are additional properties on the class, because to
+ * declare these in Closure you must declare these separately from the class.
+ */
```
evmar: Doc the return type here, because it's a bit funny?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+    }
+  }
+}
+
+/** Flags that declare a field of the same name if set on a ctor parameter. */
+const FIELD_DECLARATION_MODIFIERS: ts.ModifierFlags = ts.ModifierFlags.Private |
+    ts.ModifierFlags.Protected | ts.ModifierFlags.Public | ts.ModifierFlags.Readonly;
+
+/**
+ * emitMemberTypes emits the type annotations for members of a class. It's necessary in the case
```
evmar: No longer 'emit'

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+  // on interfaces.
+  const isInterface = ts.isInterfaceDeclaration(typeDecl);
+  const propertyDecls = [
+    ...staticProps.map(
+        p => createClosurePropertyDeclaration(
+            mtt, staticPropAccess, p, isInterface && !!p.questionToken)),
+    ...nonStaticProps.map(
+        (p) => createClosurePropertyDeclaration(
+            mtt, instancePropAccess, p, isInterface && !!p.questionToken)),
+    ...paramProps.map(
```
evmar: Does it work to do
[...staticProps, ...nonStaticProps, ...paramProps].map((p) => ...

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+
+/** Removes comment metacharacters from a string, to make it safe to embed in a comment. */
+export function escapeForComment(str: string): string {
+  return str.replace(/\/\*/g, '__').replace(/\*\//g, '__');
+}
+
+function createClosurePropertyDeclaration(
+    mtt: ModuleTypeTranslator, expr: ts.Expression,
+    prop: ts.PropertyDeclaration|ts.PropertySignature|ts.ParameterDeclaration,
+    optional: boolean): ts.Statement|null {
```
evmar: Can this return null?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+       *
+       * This is triggered when TS code writes "class X extends (Foo as Bar) { ... }", commonly done
+       * to support mixins. For extends clauses in classes, the code below drops the cast and any
+       * parentheticals, leaving just the original expression.
+       *
+       * This is an incomplete workaround, as Closure will still bail on other super expressions,
+       * but retains compatibility with the previous emit that (accidentally) dropped the cast
+       * expression.
+       *
+       * TODO(martinprobst): remove this once the Closure side issue has been resolved.
```
evmar: Is there a bug on this?

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+        mjsdoc.updateComment();
+      }
+
+      /** Returns true if a value export should be emitted for the given symbol in export *. */
+      function shouldEmitValueExportForSymbol(sym: ts.Symbol): boolean {
+        if (sym.flags & ts.SymbolFlags.Alias) {
+          sym = typeChecker.getAliasedSymbol(sym);
+        }
+        if ((sym.flags & ts.SymbolFlags.Value) === 0) {
+          // Note: We create explicit reexports via closure at another place in
```
evmar: trailed off

package.json
mprobst: It's in the commit message, CC from 2016 would fail on the newer output that included more accurate cast expressions.

mprobst:
Re buildifier, I sent you https://github.com/angular/tsickle/pull/870, will rebase on that once it's in.

src/enum_transformer.ts
```diff
@@ -145,11 +144,13 @@ export function enumTransformer(typeChecker: ts.TypeChecker, diagnostics: ts.Dia
       }
 
       const varDecl = ts.createVariableStatement(
-          [ts.createToken(ts.SyntaxKind.ConstKeyword)],
-          [ts.createVariableDeclaration(
-              name, undefined,
-              ts.createObjectLiteral(
-                  ts.setTextRange(ts.createNodeArray(values, true), node.members), true))]);
+          undefined,
```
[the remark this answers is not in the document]
mprobst: It's `modifiers`, but surprisingly (and this took me a while) *not* `const` vs `let` vs `var`. Those are in the `NodeFlags` on `VariableDeclarationList`. `modifiers` on the statement is stuff like `export`. Added comments on both.

src/externs.ts
```diff
@@ -0,0 +1,439 @@
...
+ * For example, a
+ *   declare interface Foo { bar: string; }
+ * Would generate a
+ *   /.. \@externs ./
+ *   /.. \@record ./
+ *   var Foo = function() {};
+ *   /.. \@type {string} ./
+ *   Foo.prototype.bar;
+ *
+ * The generated externs indicate to Closure Compiler that symbols are extern to the optimization
```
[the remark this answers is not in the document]
mprobst: Done.

src/externs.ts
```diff
@@ -0,0 +1,439 @@
...
+            continue;
+          }
+          // TODO: For now property names other than Identifiers are not handled; e.g.
+          //    interface Foo { "123bar": number }
+          break;
+        case ts.SyntaxKind.MethodSignature:
+        case ts.SyntaxKind.MethodDeclaration:
+          const method = member as ts.MethodDeclaration;
+          const isStatic = hasModifierFlag(method, ts.ModifierFlags.Static);
+          const methodSignature = `${method.name.getText()}$$$${isStatic ? 'static' : 'instance'}`;
```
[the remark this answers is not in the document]
mprobst: It's from here: https://github.com/angular/tsickle/blob/master/src/tsickle.ts#L1718

But yeah, I agree, that's a problem with this mega code review. In practice, I've tried to not change any code, except for the bits where it was really needed (ha, I know). That is, all code handling the API based transformation of code is new, all other code should just be a move. I had to restructure some of the book keeping due to the different structure (largely moving to function scopes vs previously having a class/instance based model).

As a rule of thumb, all code was originally in tsickle.ts. If it deals with AST construction, it's new. If it deals with information collection, it's just moved.

I hope that helps a bit? Maybe the best indicator for you is function/method names, I kept those the same with maybe one or two exceptions.

src/fileoverview_comment_transformer.ts
```diff
@@ -75,10 +68,58 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
...
+        // an @fileoverview etc, and reporting an error if so.
+        sf = updateSourceFileNode(
+            sf, ts.createNodeArray([notEmitted, firstStatement, ...sf.statements.slice(1)]));
+        break;
+      }
+
+      // Now walk every top level statement and escape/drop any @fileoverview comments found.
+      // Closure ignores all @fileoverview comments but the last, so tsickle must make sure not to
+      // emit duplicated ones.
+      // TODO(martinprobst): make this an error instead of silently dropping the comment.
```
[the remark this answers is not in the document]
mprobst: It's easy to find the failures, it's just a global presubmit. I just wanted to avoid front-loading more global changes into this PR, and keep the difference to the current behaviour of tsickle small.

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
```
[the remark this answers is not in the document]
mprobst: Done.

src/jsdoc_transformer.ts
mprobst: Done (though only moved, in my defense!)

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+      isClass && decl.heritageClauses.some(hc => hc.token === ts.SyntaxKind.ExtendsKeyword);
+  for (const heritage of decl.heritageClauses!) {
+    if (!heritage.types) continue;
+    if (isClass && heritage.token !== ts.SyntaxKind.ImplementsKeyword && !isAmbient(decl)) {
+      // If a class has "extends Foo", that is preserved in the ES6 output
+      // and we don't need to do anything.  But if it has "implements Foo",
+      // that is a TS-specific thing and we need to translate it to the
+      // the Closure "@implements {Foo}".
+      // However for ambient declarations, we only emit externs, and in those we do need to
+      // add "@extends {Foo}" as they use ES5 syntax.
```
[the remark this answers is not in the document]
mprobst: Still shared with externs.ts (that's why the fn is exported).

src/jsdoc_transformer.ts
mprobst: You tell me, I think you wrote this code :-)

But actually, you can only have param props on the last ctor. Fixed the code and added a test.

src/jsdoc_transformer.ts
mprobst: Done.

src/jsdoc_transformer.ts
mprobst: Done.

src/jsdoc_transformer.ts
mprobst: Nice, done.

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+            ts.createBlock([]),
+            )));
+    ts.setSyntheticLeadingComments(abstractFnDecl, [jsdoc.toSynthesizedComment(tags)]);
+    propertyDecls.push(ts.setSourceMapRange(abstractFnDecl, fnDecl));
+  }
+
+  // See test_files/fields/fields.ts:BaseThatThrows for a note on this wrapper.
+  return ts.createIf(
+      ts.createLiteral(false),
+      ts.createBlock(propertyDecls.filter((d): d is ts.Statement => d !== null), true));
```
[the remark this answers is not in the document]
mprobst: Good catch, together with the below this is not necessary anymore. `createClosurePropertyDeclaration` first returned null for undeclarable props, but now does return a not emitted stmt.

src/jsdoc_transformer.ts
mprobst: See above.

src/jsdoc_transformer.ts
```diff
@@ -0,0 +1,937 @@
...
+ * TODO(martinprobst): remove once the Angular issue is fixed.
+ */
+export function removeTypeAssertions(): ts.TransformerFactory<ts.SourceFile> {
+  return (context: ts.TransformationContext) => {
+    return (sourceFile: ts.SourceFile) => {
+      function visitor(node: ts.Node): ts.Node {
+        switch (node.kind) {
+          case ts.SyntaxKind.TypeAssertionExpression:
+          case ts.SyntaxKind.AsExpression:
+            return ts.visitNode((node as ts.AssertionExpression).expression, visitor);
```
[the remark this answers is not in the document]
mprobst: Yes!

src/jsdoc_transformer.ts
mprobst: There's an internal one (b/111487374); I'm never quite sure where to file bugs with Closure.

src/jsdoc_transformer.ts
mprobst: Done.

src/main.ts
```diff
@@ -87,6 +86,27 @@ function loadSettingsFromArgs(args: string[]): {settings: Settings, tscArgs: str
   return {settings, tscArgs};
 }
 
+/**
+ * Determine the lowest-level common parent directory of the given list of files.
+ */
+export function getCommonParentDirectory(fileNames: string[]): string {
```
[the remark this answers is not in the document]
mprobst: Yes, just trying to kill `utils.ts` (which, surprise, was attracting poorly thought out code like honey...)

test_files/class/class.js
```diff
@@ -49,8 +52,12 @@ if (false) {
  * @extends {Interface}
  */
 function InterfaceExtendsInterface() { }
-/** @type {function(): void} */
-InterfaceExtendsInterface.prototype.interfaceFunc2;
+if (false) {
```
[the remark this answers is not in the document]
mprobst: I've unified the emit for interfaces and classes in an effort to reduce duplication; both now go through `createMemberTypeDeclaration`. My reasoning is that it doesn't hurt (well, I hope), and if we need to add more logic on filtering members or handling special conditions, we have just one place to fix.

src/jsdoc.ts
mprobst: I had to add this because we now use this code path to emit typedefs, and wanted to keep the text change small.

src/module_type_translator.ts
```diff
@@ -0,0 +1,541 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+import * as googmodule from './googmodule';
```
evmar: Some sort of @fileoverview on this would be helpful.

src/module_type_translator.ts
```diff
@@ -0,0 +1,541 @@
...
+  symbolsToAliasedNames = new Map<ts.Symbol, string>();
+
+  /**
+   * The set of module symbols forward declared in the local namespace (with goog.forwarDeclare).
+   *
+   * Symbols not imported must be declared, which is done by adding forward declares to
+   * `extraImports` below.
+   */
+  private forwardDeclaredModules = new Set<ts.Symbol>();
+  private forwardDeclares: ts.Statement[] = [];
```
evmar: Docs on these members.

src/transformer_util.ts
```diff
@@ -697,3 +157,66 @@ export function createSingleQuoteStringLiteral(text: string): ts.StringLiteral {
...
+}
+
+/**
+ * debugWarn logs a debug warning.
+ *
+ * These should only be used for cases where tsickle is making a questionable judgement about what
+ * to do. By default, tsickle does not report any warnings to the caller, and warnings are hidden
+ * behind a debug flag, as warnings are only for tsickle to debug itself.
+ */
+export function debugWarn(
```
evmar: This function does not seem to be related to transformer_util

test/e2e_closure_test.ts
```diff
@@ -49,7 +49,6 @@ describe('golden file tests', () => {
         'checkRegExp',
         'checkTypes',
         'checkVars',
-        'commonJsModuleLoad',
```
evmar: One reason I keep harping about separate CLs is that otherwise we have no context on why changes like this were made, except that "Martin noticed it while changing 3000 other lines of code".  And you could explain each of these by "it was necessary", but it'd be nice to have a log of why.  For this particular one, given that we're trying to preserve existing outputs, why is it needed?

test/test_support.ts
```diff
@@ -405,7 +366,7 @@ export function compileWithTransfromer(
   const program = ts.createProgram(fileNames, compilerOptions, tsHost);
   expectDiagnosticsEmpty(ts.getPreEmitDiagnostics(program));
 
-  const rootModulePath = rootPath ? rootPath : getCommonParentDirectory(fileNames);
+  const rootModulePath = rootPath ? rootPath : path.dirname(fileNames[0]);
```
evmar: Why is this needed?

test_files/cast_extends/cast_extends.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @fileoverview Reproduces an issue where tsickle would emit a cast for the "extends" claus, and
+ * Closure would report an error due to the extends expression not resolving to a plain identifier.
```
evmar: How do transformers trigger this?

test_files/enum/enum.ts
```diff
@@ -75,5 +75,10 @@ enum StringKeyEnum {
 enum MixedEnum {
   STR = 'abc',
   NUM = 3,
-  'string key' = 4 
+  'string key' = 4
+}
+
+/** JSDoc in here. */
```
evmar: Can you add comments near here about what this is trying to test?

test_files/export/export_star_imported.js
```diff
@@ -12,18 +12,17 @@ exports.export1 = export_helper_1.export1;
 exports.export3 = export_helper_1.export3;
 exports.export5 = export_helper_1.export5;
 exports.export4 = export_helper_1.export4;
-const tsickle_forward_declare_2 = goog.forwardDeclare("test_files.export.export_helper");
-/** @typedef {!tsickle_forward_declare_2.Bar} */
+/** @typedef {!tsickle_forward_declare_1.Bar} */
 exports.Bar; // re-export typedef
-/** @typedef {!tsickle_forward_declare_2.TypeDef} */
+/** @typedef {!tsickle_forward_declare_1.TypeDef} */
```
evmar: Why are these tests changing?

test_files/jsdoc/jsdoc.js
```diff
@@ -137,21 +135,24 @@ class RedundantJSDocShouldBeStripped {
...
-/** *
+/**
  * Don't emit type comments for Polymer behaviors
  * if they are declared via the Polymer function.
-  @type {?} */
+ * @type {?}
+ */
 let Polymer;
-Polymer({ behaviors: ['test'] });
+Polymer({ behaviors: [(/** @type {?} */ ('test'))] });
```
evmar: Will this break PolymerPass?  Or is fixing that the CL I saw go by...

test_files/jsdoc_types.untyped/module2.ts
```diff
@@ -3,8 +3,7 @@ export class ClassTwo {}
 export interface Interface { x: number }
 export class ClassWithParams<T> {}
 
-// TODO(evanm):
-// export type TypeAlias = number;
-// export type TypeAliasWithParam<T> = T[];
```
evmar: Did you fix these?  How is this related to transformers?

test/test_support.ts
mprobst: Because I want to run this test in google3, where we don't compile `main.ts` at all, plus this method is poorly thought out as you point out above, so we're better off not using it. Again, this is in a separate commit.

test_files/cast_extends/cast_extends.ts
mprobst: Transformers are much better at retaining comments and parens - previously we'd just silently drop the whole comment sequence that replaced the cast.

test_files/export/export_star_imported.js
mprobst: Previously, the code eagerly emitted a duplicated `forwardDeclare` for `export_helper`, because emitting a forwardDeclare for an `export *` did not check whether the module symbol already had a forward declare. I've harmonized that code during the rewrite.

(I think that is, it's two weeks ago that I wrote this by now).

test_files/jsdoc/jsdoc.js
mprobst: Yes. It turns out that Tobias added a crazy hack to tsickle to work around five failing files in google3. I went ahead and fixed the five files, and now aomarks also fixed PolymerPass, so this is no longer an issue.

Once JSCompiler releases with the fix, we should probably just delete this code.

test_files/jsdoc_types.untyped/module2.ts
mprobst: I didn't fix these in this change, they've been fixed a long time ago. I just came across the commented out lines here and cleaned it up. I don't think it's worth extracting a separate commit for this.

PR #864, opened 2018-07-11
pass through the is-es5 flag into golden tests

evmar:
We have some tests that are marked for testing es5 emit, but
we failed to pass that flag through to the googmodule processor.
Fix this bug and rebaseline the tests.

### DIFF
--- a/test_files/abstract/abstract.js
+++ b/test_files/abstract/abstract.js
@@ -4,6 +4,8 @@
  */
 goog.module('test_files.abstract.abstract');
 var module = module || { id: 'test_files/abstract/abstract.ts' };
+module = module;
+exports = {};
 /**
  * @abstract
  */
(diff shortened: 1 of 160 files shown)


test_files/abstract/abstract.js
```diff
@@ -4,6 +4,8 @@
  */
 goog.module('test_files.abstract.abstract');
 var module = module || { id: 'test_files/abstract/abstract.ts' };
+module = module;
```
rkirov: :/ for my own education what is this? any why is it esX dependent?

evmar: I found this code:

        // The module=module assignment suppresses an unused variable warning which may trigger
        // depending on the project's compilation flags.

https://github.com/angular/tsickle/blob/1f28293d3c1aa2ccf1d163bf19f91e01f487c635/src/googmodule.ts#L461

I bet we could probably turn it off now that we suppress uselessCode.

Also it's weird how we use 'es6' to mean 'closure output' but so it goes.

PR #866, opened 2018-07-13
Handle conditional and substitution types.

### LINKED ISSUE #865: some conditional types crash tsickle
The following makes tsickle throw:

```typescript
export type Filter<T, U> = T extends U ? T : never;
export function ofType(): Filter<string, string> {
  return null as any;
};

```

believe it or not this is real-world code from https://github.com/ngrx/platform/pull/1183/files#diff-37851f2da2623baafce63a5976e8d86b

mprobst:
These cannot be represented in Closure Compiler easily (at least without
emitting Type Transformation Language, which does not work on typedefs),
so just emit `?` for the time being, instead of crashing.

Fixes #865.

### DIFF
--- a/test_files/conditional_type/conditional_type.ts
+++ b/test_files/conditional_type/conditional_type.ts
@@ -1,0 +1,1 @@
+export type Filter<T, U> = T extends U ? T : never;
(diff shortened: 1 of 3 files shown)


test_files/conditional_type/conditional_type.ts
```diff
@@ -0,0 +1 @@
+export type Filter<T, U> = T extends U ? T : never;
```
rkirov: I was going to say add an example of substitution type, but if I read this correctly - https://github.com/Microsoft/TypeScript/blob/master/src/compiler/types.ts#L4051

they can only occur internally to conditional types, thus can't be tested.

PR #871, opened 2018-07-18
wrap interface declarations in if(false)

evmar:
In our upcoming world of transformers, it's easier if we consistently
always wrap interface member declarations in an if(false) block.

This partially rolls back some of 249273f6417ea3696934c5b71abbbf4b202ce612,
which dropped function wrapper in favor of either using if(false) or
no wrapper at all for interfaces, so now we consistently use if(false).

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -1326,10 +1326,14 @@ class Annotator extends ClosureRewriter {
     const name = getIdentifierText(iface.name);
     this.emit(`function ${name}() {}\n`);
 
-    const memberNamespace = [name, 'prototype'];
-    for (const elem of iface.members) {
-      const isOptional = elem.questionToken != null;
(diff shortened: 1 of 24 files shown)


evmar:
I think this won't ruin your branch because you just need to UPDATE_GOLDENS=y after merging this.

src/tsickle.ts
```diff
@@ -1326,11 +1326,13 @@ class Annotator extends ClosureRewriter {
     const name = getIdentifierText(iface.name);
     this.emit(`function ${name}() {}\n`);
 
+    this.emit(`\nif (false) {\n`);
```
mprobst: can you only emit this if there are any members?

evmar: Sure.  I think it's pretty uncommon to declare an interface with no members, but this code's dying anyway so it can't hurt.

PR #873, opened 2018-07-25
Deduplicate JSDoc tags in function declarations.

mprobst:
Previously, we'd emit tags like `@export` multiple times. causing
Closure complaints down the line.


src/module_type_translator.ts
```diff
@@ -410,7 +410,15 @@ export class ModuleTypeTranslator {
   getFunctionTypeJSDoc(fnDecls: ts.SignatureDeclaration[], extraTags: jsdoc.Tag[] = []):
       [jsdoc.Tag[], string[]] {
     const typeChecker = this.typeChecker;
-    const newDoc = extraTags;
+
+    // De-duplicate tags and docs found for the fnDecls.
```
evmar: Will this deduplicate @params, which we do want multiple of?  (E..g if the function has @param x and @param y)

mprobst: No, tags called `@param` or `@return` are special cased below. We also have plenty of tests for this, I know because I broke them at first :-)

PR #875, opened 2018-07-26
 Handle UniqueESSymbol.

mprobst:
TypeScript distinguishes internally between the primitive type `symbol`,
and the type of one specific unique symbol. This allows TS to
distinguish symbols that are defined on a particular type or not.
Closure has no such facility, so translate both as just plain `symbol`.


src/type_translator.ts
```diff
@@ -336,8 +336,7 @@ export class TypeTranslator {
         }
         return this.symbolToString(type.symbol, true);
       case ts.TypeFlags.ESSymbol:
-        // NOTE: currently this is just a typedef for {?}, shrug.
-        // https://github.com/google/closure-compiler/blob/55cf43ee31e80d89d7087af65b5542aa63987874/externs/es3.js#L34
+      case ts.TypeFlags.UniqueESSymbol:
```
evmar: Maybe put a comment in here about how these are different?

PR #837, opened 2018-06-19
Repro for node_modules not taking effect.

mprobst:

PR #877, opened 2018-07-27
WIP: Update to rules_nodejs 0.11.2

gregmagolan:
DO NOT MERGE YET

Includes transitive skylib dep now required by rules_nodejs

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -2,9 +2,16 @@ workspace(name="tsickle")
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    url = "https://github.com/bazelbuild/rules_nodejs/archive/0.9.1.zip",
-    strip_prefix = "rules_nodejs-0.9.1",
-    sha256 = "6139762b62b37c1fd171d7f22aa39566cb7dc2916f0f801d505a9aaf118c117f",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.11.2.zip"],
+    strip_prefix = "rules_nodejs-0.11.2",
+    sha256 = "c00d5381adeefb56e0ef959a7b168cae628535dab933cfad1c2cd1870cd7c9de",
+)
+
+http_archive(
+    name = "bazel_skylib",
+    urls = ["https://github.com/bazelbuild/bazel-skylib/archive/0.3.1.zip"],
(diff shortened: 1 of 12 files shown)


gregmagolan:
This is to test the latest rules_nodejs release downstream. Its failing on something unexpected:
```
rc/jsdoc_transformer.ts:87:68 - error TS4078: Parameter 'decl' of exported function has or is using private name 'HasTypeParameters'.
87 export function maybeAddTemplateClause(docTags: jsdoc.Tag[], decl: HasTypeParameters) {
                                                                      ~~~~~~~~~~~~~~~~~
```
Any ideas? Doesn't seem related to updating rules_nodejs. Looks more like a typescript version issue but maybe the latest rules_nodejs brought that out since it prevents typescript 2.7.2 from getting resolved by ts_library and uses the user typescript version instead which is `"typescript": "git://github.com/mprobst/TypeScript.git#5456479a4a"` in this case.

gregmagolan:
@mprobst I'm getting failures on the tests for past typescript versions with the updates of rules_nodejs & rules_typescript. Could you take a quick look a let me know what you think? The updates for the rules fix some issues with non-hermetic resolves so its possible that the tests were not being run correctly before. Is tsickle expected to work with all of these typescript versions?

evmar:
https://github.com/angular/tsickle/pull/837 has a repro with some links to bugs I think.

evmar:
(I don't understand that bug or what's gone wrong here, sorry if it's not relevant.)

mprobst:
@gregmagolan thanks for the PR. That the tests are failing actually indicates things are working - we had no test coverage for old TS versions for a while now, so of course things diverged.

@evmar @rkirov @alexeagle given that we've been in this broken state for a year, I strongly suspect there's no value in us to maintain this backwards-compat test suite. The g3 users are on a fixed version anyway, and Angular is no longer using substantial parts of tsickle AFAIU (and in the process of migrating away entirely). Should we just drop `test_ts*` and only run with the one TS version?

@gregmagolan this might sound a bit as if this work wasn't worth it, but I find it really reassuring that we now have a hermetic version of the node_modules that does what we expect it to do, so nothing went to waste here.

mprobst:
Obsolete.

PR #879, opened 2018-07-31
Support exporting local type aliases.

mprobst:
The error reporting mechanism here was plain wrong, it is totally fine
to export symbols that do not have an alias, because they are defined
locally or globally.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -795,8 +795,6 @@ export function jsdocTransformer(
         if (host.untyped) return exportDecl;
 
         const result: ts.Node[] = [exportDecl];
-        const typeTranslator =
-            moduleTypeTranslator.newTypeTranslator(ts.getOriginalNode(exportDecl));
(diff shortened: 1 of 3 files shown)


src/jsdoc_transformer.ts
```diff
@@ -805,12 +803,8 @@ export function jsdocTransformer(
...
           if (!isTypeAlias) continue;
-          const typeName = moduleTypeTranslator.symbolsToAliasedNames.get(aliasedSymbol);
-          if (!typeName) {
-            moduleTypeTranslator.error(
-                ts.getOriginalNode(exportDecl), `cannot find alias for re-exported symbol`);
-            continue;
-          }
+          const typeName =
+              moduleTypeTranslator.symbolsToAliasedNames.get(aliasedSymbol) || aliasedSymbol.name;
           // Leading newline prevents the typedef from being swallowed.
```
evmar: May as well delete this comment while you're in here.

PR #880, opened 2018-07-31
Update to TS 3.0

alippai:


mprobst:
Thanks for the contribution!

I'd expect this PR to fail, because tsickle depends on a deprecated TS API that was removed in 3.0 (SymbolBuilder). Fixing that will be a bit tricky from the way it looks, so it might be a while until we can move forward here.

alippai:
Oh, I see now. I'll stop assuming https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#typescript-30 is detailed enough :) :P

alippai:
https://github.com/Microsoft/TypeScript/wiki/API-Breaking-Changes this is the correct link (with the correct change notice)

evmar:
Yep, I started exploring our options in https://github.com/angular/tsickle/pull/878 but it sounds like it might not be easy.

Issue #863, opened 2018-07-11
build error when a method has documentation with annotations above it ie. @param #964


aykutyararbas:
Please also refer to issue at: 
https://github.com/dherges/ng-packagr/issues/964#issuecomment-403762925

Type of Issue
Build error

[X] Bug Report
[ ] Feature Request
Description
Adding doc with method signature annotations causing packager to error.

LIKE
BUILD ERROR
src/app/common/featureflags.service.ts(59,3): warning TS0: the type annotation on @param is redundant with its TypeScript

Example service method

/**

Returns the service label for a given service
Returns the id , if the service is not available
@param {string} id
@returns {string}
*/
public getServiceLabel(id: string): string {
if(this.SERVICE_REGISTRY[id]) {
return this.SERVICE_REGISTRY[id].label;
} else {
return id;
}
}
How To Reproduce
Steps
1 - Add a documentation to one of your service
Like the one above start with / * * and add @param
End your documentation

2 - Run ng-packagr -p ng-package.json

Can you reproduce the error in the integration tests in ng-packagr?
If possible, take a look at the integration/samples and try to break one of >>these builds!

Is the error you faced in an application importing the library
Try to break the Angular CLI app in integration/consumers/ng-cli!

Expected Behaviour
A bug: please describe what behavior or result you expected
Build without erroring out.

Version Information
$ node_modules/.bin/ng-packagr --version
ng-packagr: 5.6.0
@angular/*: 5.2.10
typescript: ~2.6.2
"rxjs": "^5.5.11",
node: v8.11.1
npm: 5.6.0
Please include all version numbers that might be relevant, e.g. third-party libraries

Thanks and regards,

Aykut~

evmar:
Why not remove the {string} ?

aykutyararbas:
Well, it is the documentation portion of it, like in the Java Docs. 
When you remove that standard representation of the method's parameters you'll not be able to pass any type in the documentation, at the end any doc builders will not be able to use it.

The main issue seems to be 'context'.  When building/ compiling if you are in the documentation, you can easily ignore these lines.

jscharett:
This "feature" seems to make it impossible to compile large angular apps/libraries that have made use of the type info.  Granted, it may be irrelevant, but for a large library, its very painful to have to go and "correct" this.  And the kicker, its just a comment, not actual code!  It would be much better to send it out as a warning vs an error.

evmar:
As included in the original report, this is a warning, not an error:

[...]

PR #881, opened 2018-08-01
 Fail on duplicated and run in file comments. 

mprobst:
In TypeScript, file level comments are comments that are separated from
the file body with an empty line. In Closure, file level comments are
comments marked with a particular (often load bearing) JSDoc tag, e.g.
`@fileoverview` or `@pintomodule`.

If a TypeScript contains such a Closure file level comment in a comment
that is *not* a TypeScript file level comment, TypeScript will often not
emit the comment at all, e.g. because it was placed on an interface or
type import that will not be emitted. That violates user expectations,
as their (load bearing) comment is now silently gone.

This change has tsickle report hard errors whenever it encounters a file
comment that's either not properly separated from the file body, is
duplicated, or appears after the initial statement in the body of the
file, so that the situation can be discovered and fixed early by the
user.

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -7,7 +7,7 @@
  */
 
 import * as jsdoc from './jsdoc';
-import {createNotEmittedStatement, synthesizeCommentRanges, updateSourceFileNode} from './transformer_util';
+import {createNotEmittedStatement, reportDiagnostic, synthesizeCommentRanges, updateSourceFileNode} from './transformer_util';
 import * as ts from './typescript';
 
 /**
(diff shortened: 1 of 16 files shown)


src/fileoverview_comment_transformer.ts
```diff
@@ -61,89 +61,106 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
...
-        // TODO(martinprobst): consider checking here whether any of the trailing comments contains
-        // an @fileoverview etc, and reporting an error if so.
-        sf = updateSourceFileNode(
-            sf, ts.createNodeArray([notEmitted, firstStatement, ...sf.statements.slice(1)]));
-        break;
+export function transformFileoverviewCommentFactory(diagnostics: ts.Diagnostic[]) {
+  return (): (sourceFile: ts.SourceFile) => ts.SourceFile => {
+    function checkNoFileoverviewComments(
+        context: ts.Node, comments: jsdoc.SynthesizedCommentWithOriginal[], message: string) {
+      for (let j = 0; j < comments.length; j++) {
```
evmar: Looks like this could be a for-of, or at least use 'i' as index.

src/fileoverview_comment_transformer.ts
```diff
@@ -61,89 +61,106 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
...
-    let tags: jsdoc.Tag[] = [];
-    for (let i = fileComments.length - 1; i >= 0; i--) {
-      const parse = jsdoc.parseContents(fileComments[i].text);
-      if (parse !== null && parse.tags.some(t => FILEOVERVIEW_COMMENT_MARKERS.has(t.tagName))) {
-        fileoverviewIdx = i;
-        tags = parse.tags;
-        break;
+      // Closure Compiler considers the *last* comment with @fileoverview (or @externs or
+      // @nocompile) that has not been attached to some other tree node to be the file overview
+      // comment, and only applies @suppress tags from it. AJD considers *any* comment mentioning
```
evmar: Is AJD public?  Maybe just add "(a Google-internal tool)" if not.

src/transformer_util.ts
```diff
@@ -190,19 +189,51 @@ export function createMultiLineComment(original: ts.Node, text: string) {
...
-    file: context.getSourceFile(),
-    start: context.getStart(),
-    length: context.getEnd() - context.getStart(),
+  host.logWarning(createDiagnostic(node, ts.DiagnosticCategory.Warning, messageText));
+}
+
+/**
+ * Creates and reports a diagnostic by adding it to the given array.
+ *
+ * This is used for erros and warnings in tsickle's input. Emit errors (the default) if tsickle
```
evmar: Typo: errors

src/transformer_util.ts
```diff
@@ -190,19 +189,51 @@ export function createMultiLineComment(original: ts.Node, text: string) {
...
+ * cannot emit a correct result given the input. Emit warnings for questionable input if there's a
+ * good chance that the output will work.
+ *
+ * For typical tsickle users, errors are always reported and break the compilation operation,
```
evmar: typical = google3?  This user seemed to be experiencing warnings as errors, though I'm not sure what tool they were using. https://github.com/angular/tsickle/issues/863

src/transformer_util.ts
```diff
@@ -190,19 +189,51 @@ export function createMultiLineComment(original: ts.Node, text: string) {
...
+export function reportDiagnostic(
+    diagnostics: ts.Diagnostic[], node: ts.Node, messageText: string, textRange?: ts.TextRange,
+    category = ts.DiagnosticCategory.Error) {
+  diagnostics.push(createDiagnostic(node, category, messageText, textRange));
+}
+
+function createDiagnostic(
+    node: ts.Node, category: ts.DiagnosticCategory, messageText: string,
+    textRange?: ts.TextRange): ts.Diagnostic {
+  // Cannot use getStart as node might be synthesized.
```
evmar: It seems to use getStart just below (?)

src/transformer_util.ts
```diff
@@ -190,19 +189,51 @@ export function createMultiLineComment(original: ts.Node, text: string) {
...
+ * @param textRange pass to overrride the text range from the node with a more specific range.
+ */
+export function reportDiagnostic(
+    diagnostics: ts.Diagnostic[], node: ts.Node, messageText: string, textRange?: ts.TextRange,
+    category = ts.DiagnosticCategory.Error) {
+  diagnostics.push(createDiagnostic(node, category, messageText, textRange));
+}
+
+function createDiagnostic(
+    node: ts.Node, category: ts.DiagnosticCategory, messageText: string,
```
evmar: It's a little funny how this one takes its args in a different order the one above.

src/fileoverview_comment_transformer.ts

src/fileoverview_comment_transformer.ts

src/transformer_util.ts
mprobst: Yeah, typical = google3. I have no idea about that other thing.

src/transformer_util.ts
mprobst: Done.

src/transformer_util.ts
mprobst: Done.

PR #882, opened 2018-08-02
fix __@observable

evmar:
This fixes the rxjs sync.  See individual commits.


src/type_translator.ts
```diff
@@ -9,6 +9,25 @@
...
+ *     'complex name': string;
+ *   }
+ *   Foo.bar;  // ok
+ *   Foo['bar']  // ok
+ *   Foo['complex name']  // ok
+ *
+ * In Closure-land, we want identify that the legal name 'bar' can become an
+ * ordinary field, but we need to skip strings like 'complex name'.
+ */
+export function isValidClosurePropertyName(name: string): boolean {
```
evmar: This is a weird place for this function.  But its previous home was in a jsdoc file which is also a weird location.  I don't feel great about it.

test_files/unique_symbol/unique_symbol.ts
```diff
@@ -0,0 +1,17 @@
...
+// library.
+declare global {
+  interface SymbolConstructor {
+    readonly observable: symbol;
+  }
+}
+
+// This test is verifying the type of this expression, which ultimately
+// refers to some TypeScript internal __@observeable thing.
+export let x: {[Symbol.observable]: string}|null = null;
```
mprobst: can you give this one more property? That'd make it more obvious to me that it's only filtering the bad ones.

By the way, I think this also fixes another problem:

```
const x: {'quoted-thing': string}: {...};
```

Maybe add a test for that so we don't regress to just specially treating symbols? In that case, I think it'd be nice to call this test folder something like non identifier properties or so.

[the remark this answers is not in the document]
evmar: Done

evmar: Done

PR #884, opened 2018-08-05
 Upgrade all package dependencies.

mprobst:
... by running `yarn upgrade --latest` (took me a while to find the magical incantation).

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
   "dependencies": {
     "minimist": "^1.2.0",
     "mkdirp": "^0.5.1",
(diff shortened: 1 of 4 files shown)


mprobst:
@evmar yeah I think it's fine from that angle. We don't really have a strategy for keeping these in sync, except for "fix if anything breaks". That might be the path of least resistance here.

I'm mostly upgrading for Closure Compiler and formatting; the other packages had some reported DoS vulnerabilities. I don't think that's particularly worrying, as it's only dev dependencies, but still, hygiene.

PR #892, opened 2018-09-11
Anon class handling

evmar:
Two commits, review separately.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -521,7 +521,14 @@ export class TypeTranslator {
         this.warn('class has no symbol');
         return '?';
       }
-      return '!' + this.symbolToString(type.symbol, /* useFqn */ true);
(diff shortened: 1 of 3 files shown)


src/type_translator.ts
```diff
@@ -521,7 +521,13 @@ export class TypeTranslator {
         this.warn('class has no symbol');
         return '?';
       }
-      return '!' + this.symbolToString(type.symbol, /* useFqn */ true);
+      const name = this.symbolToString(type.symbol, /* useFqn */ true);
+      if (name === '(Anonymous class)') {
+        // Values that have anonymous class types produce this name, but the type
+        // appears otherwise identical to a named class.
```
rkirov: Care to comment on intent to handle this differently in the future? I.e. produce a type that closure understands vs '?'.

PR #898, opened 2018-09-20
suppress missingReturn closure check

evmar:
If the TS compiler's exhaustiveness analysis concludes that all
possible branches return a value, then we don't want Closure to
second-guess this decision.


mprobst:
I hadn't seen this come up in a TGP lately, but I guess you ran into it?

evmar:
One example is that there's a google3 copy of this file that doesn't have this final return statement:
https://github.com/angular/angular/blob/482e12c940cc4eb7e958467f88f4ab5de73d5bd4/packages/core/src/reflection/reflection_capabilities.ts#L143

Which think is actually a bug. :~(
But it's accepted by TS because it's in third_party I think, and then we warn about it.

evmar:
For third-party code we should turn off Closure warnings I think.  But there are other examples in first-party code that I can't link here, so I think the underlying idea stands.

PR #899, opened 2018-09-21
Handle ambient exported enums.

mprobst:
TypeScript just emits code accessing these, but no export statements.


src/enum_transformer.ts
```diff
@@ -102,6 +103,9 @@ export function enumTransformer(typeChecker: ts.TypeChecker, diagnostics: ts.Dia
       // this is left to fix for another day.
       if (isInNamespace(node)) return ts.visitEachChild(node, visitor, context);
 
+      // Ambient enums should not be materialized.
```
evmar: Can you put in a ", because [reason]" here?  It's not immediately obvious to me.

mprobst: Done - it's because TS doesn't produce values for an ambient enum either. Previously we'd accidentally generate values on the Closure side, which doesn't match vanilla TS then.

test_files/declare_export.untyped/declare_export.ts
```diff
@@ -4,3 +4,4 @@ export declare class ExportDeclaredClass { x: number; };
 export declare function exportedDeclaredFn(): string;
 export declare const multiExportedDeclaredVar1, multiExportedDeclaredVar2: number;
 export declare type X = string;
+export declare enum ExportDeclaredEnum { VALUE = 1 };
```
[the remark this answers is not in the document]
mprobst: Accidental, fixed.

PR #900, opened 2018-09-26
Handle `this` return types.

mprobst:
Closure compiler represents these by a method that has an `@template`d
`@this` type.

TypeScript does not expose the required `isThisType` field, which this
works around by casting through `any`.


src/module_type_translator.ts
```diff
@@ -472,29 +472,35 @@ export class ModuleTypeTranslator {
...
+        };
         const retType = typeChecker.getReturnTypeOfSignature(sig);
-        const retTypeString: string = this.typeToClosure(fnDecl, retType);
-        let returnDoc: string|undefined;
-        for (const {tagName, text} of tags) {
-          if (tagName === 'return') {
-            returnDoc = text;
-            break;
+        // tslint:disable-next-line:no-any accessing TS internal field.
+        if ((retType as any).isThisType) {
```
evmar: This is probably fine, but it also seems likely there's some way to get at the 'this' without private API, like getting flags from it.

test_files/return_this/return_this.ts
```diff
@@ -0,0 +1,21 @@
+class MethodsReturnThis {
+  returnsThis() {
+    return this;
+  }
+
+  explicitThis(): this {
+    return this;
+  }
+
+  templateAndThis<T>(t: T): this {
```
evmar: Might be worth having an overloaded function too, to check the overload merging case.

mprobst: Done.

src/module_type_translator.ts
mprobst: There isn't, I've looked quite a bit. The type then only looks like a parameterized type otherwise.

mprobst:
PTAL. I added a special case for access to this that I'm not super proud of. Also see internal discussion on b/116692047.

src/jsdoc_transformer.ts
```diff
@@ -429,7 +427,23 @@ export function jsdocTransformer(
        */
       const expandedStarImports = new Set<string>();
 
+      /**
+       * While Closure compiler supports parameterized types, including parameterized `this` on
+       * methods, it does not support constraints on them. That means that a `\@template`d `this` is
+       * always considered to be `unknown`.
```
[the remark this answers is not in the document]
mprobst: `this` is just `this` e.g. in a class body (`class Foo { bar() { this; } }`), that is, JSCompiler has hard-wired support for `this` inside classes (as you'd expect - maybe I'm misunderstanding your question?).

The problem this fixes is that we need to represent `@return {this}` (which does not work in JSCompiler) as an `@template {THIS} @return {THIS}`. That works in so far as it communicates to the caller that the return value is an instance of the given `this` context the method is invoked on:

```
class Parent {
  /** @template THIS @return {THIS} @this {THIS} */ takeThat() { return this; }
}
class Sub extends Parent {}

const sub = new Sub().takeThat();
// typeof sub is now `Sub`, not `Parent`!
```

However, JSCompiler has no support for `@template` type checking within methods in general, nor constraints on them. So `this` within `takeThat` is now `unknown` to JSCompiler, which causes warnings in some binaries (conformance checks).

To work around that, we're with this casting every access to `this` to the constraint known to us, i.e. that it must at least be a subtype of `Parent`.

I think we cannot easily solve this for the general case right now (arbitrary `@template` tags), as those types in the general case might be hard or impossible to express in JSCompiler (unbound `T` style types, or complex constraints). We could special case the basic `foo<T extends Bar>` to cast every access to `T` to a `Bar`, but I'm not sure if that's worth it, and it might have other problems (returning `Bar` not matching the specific instantiation of `T`, not sure if JSCompiler checks that).

mprobst:
PTAL, this grew a bit.

mprobst:
> I'm not clear on how visitEachChild knows when to stop recursion.

visitEachChild has the given visitor function walk the next AST level (each immediate child). We're calling into our main visitor function, which then dispatches back into our main `switch` statement. That in turn will dispatch back into `visitEachChild`. The end of recursion is reached when the entire AST has been walked.

This is the general pattern in TS transformers (and I think it's common in AST manipulation): you have a dispatch method (called `visitor` here) that visits every AST node. It picks out specific cases it knows how to handle (the `switch`), and for all others recurses using the `visitEachChild` handler.

We need to recurse through the full AST, because you could have another class with a method with `: this` return type at any level, e.g. nested in a method (there's a test for that).

We know what this nodes *not* to fix up by backing up and setting `contextThisType` in places that have a this context - `FunctionLikeDeclaration` and `ClassDeclaration`. So in your example, for the last two lines we'll back up `contextThisType` when we come across the (arrow) `function`. Within the function's AST tree, we'll then leave `this` alone.

src/jsdoc_transformer.ts
```diff
@@ -526,21 +541,38 @@ export function jsdocTransformer(
...
+        return result;
+      }
+
+      /**
+       * In methods with a templated this type, adds explicit casts to accesses on this.
+       *
+       * @see contextThisType
+       */
+      function visitThisExpression(node: ts.ThisExpression) {
+        if (!contextThisType) return ts.visitEachChild(node, visitor, context);
```
evmar: What are the children of a this expression?

src/module_type_translator.ts
```diff
@@ -377,7 +377,7 @@ export class ModuleTypeTranslator {
    *    function statement; for overloads, name will have been merged.
    */
   getFunctionTypeJSDoc(fnDecls: ts.SignatureDeclaration[], extraTags: jsdoc.Tag[] = []):
-      [jsdoc.Tag[], string[]] {
+      {tags: jsdoc.Tag[], parameterNames: string[], thisReturnType: ts.Type|null} {
```
evmar: Can you doc what thisReturnType is?  It's the actual return type if the function returns this, right?

src/jsdoc_transformer.ts
mprobst: None. But I think it's better to follow the general pattern instead of special casing this per syntax construct, isn't it?

src/module_type_translator.ts
mprobst: Good catch, I forgot that we don't annotate arrows. Done.

mprobst: Don.

mprobst:
You're right re arrow function, I forgot that we don't annotate those at all and though it'd be part of the function like. Fixed, with test.

evmar:
Ah, I was thinking arrow functions went through the functionlikedeclaration branch (should they?) in which case we'd need to undo the special this handling there.  If there's a test then I'm happy.

evmar:
(LGTM)

mprobst:
@evmar re should arrow functions be going through the function like/annotation branch. If I remember correctly, we initially annotated them, but ran into issues where it's hard to annotate an expression (as opposed to a declaration) that's somewhere inline in code.

PR #901, opened 2018-09-26
consistently use absolute paths

evmar:
Two commits, can be reviewed separately.

This is part of my yak shave for running tsickle tests in google3, but it also came up in an unrelated yak shave, so I may as well submit it.  If the approach looks good to you I'll need to test it against g3 and verify I didn't break any assumptions there as well.

### DIFF
--- a/src/externs.ts
+++ b/src/externs.ts
@@ -109,10 +109,11 @@ const EXTERNS_HEADER = `/**
  * Concatenate all generated externs definitions together into a string, including a file comment
  * header.
  */
-export function getGeneratedExterns(externs: {[fileName: string]: string}): string {
+export function getGeneratedExterns(
+    rootDir: string, externs: {[fileName: string]: string}): string {
   let allExterns = EXTERNS_HEADER;
   for (const fileName of Object.keys(externs)) {
-    allExterns += `// externs from ${fileName}:\n`;
+    allExterns += `// externs from ${path.relative(rootDir, fileName)}:\n`;
     allExterns += externs[fileName];
   }
   return allExterns;
--- a/src/externs.ts
+++ b/src/externs.ts
@@ -108,9 +108,11 @@ const EXTERNS_HEADER = `/**
 /**
  * Concatenate all generated externs definitions together into a string, including a file comment
  * header.
+ *
+ * @param rootDir Project root.  Emitted comments will reference paths relative to this root.
+ *    This param is effectively required, but made optional here until Angular is fixed.
  */
-export function getGeneratedExterns(
-    rootDir: string, externs: {[fileName: string]: string}): string {
+export function getGeneratedExterns(externs: {[fileName: string]: string}, rootDir = ''): string {
   let allExterns = EXTERNS_HEADER;
   for (const fileName of Object.keys(externs)) {
     allExterns += `// externs from ${path.relative(rootDir, fileName)}:\n`;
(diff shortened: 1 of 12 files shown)


src/externs.ts
```diff
@@ -124,14 +124,9 @@ export function getGeneratedExterns(
...
- *
- * @param contextFileName if given is used as the context file path to resolve the source file's
- *     name against.
  */
-export function moduleNameAsIdentifier(
-    host: AnnotatorHost, fileName: string, contextFileName?: string): string {
-  const resolved = resolveModuleName(host, contextFileName || '', fileName);
-  return host.pathToModuleName('', resolved).replace(/\./g, '$');
+export function moduleNameAsIdentifier(host: AnnotatorHost, fileName: string): string {
+  return host.pathToModuleName('', fileName).replace(/\./g, '$');
```
mprobst: reminder: we'll need to make sure the bazel rules implementation (in the host) has the same assertion here before we land this.

test/BUILD
```diff
@@ -30,6 +30,8 @@ ts_library(
 jasmine_node_test(
     name = "unit_test",
     srcs = [":unit_test_lib"],
+    # Unit tests need access to tslib.
```
mprobst: shouldn't that rather be a "deps" entry on unit_test_lib then? data is not how TS dependencies are supposed to work.

test/e2e_main_test.ts
```diff
@@ -24,25 +26,22 @@ describe('toClosureJS', () => {
     const result = toClosureJS(
         testSupport.compilerOptions, filePaths, {isTyped: true},
         (filePath: string, contents: string) => {
-          files.set(filePath, contents);
+          files.set(path.relative(testSupport.compilerOptions.rootDir!, filePath), contents);
```
mprobst: Add a comment that this post-processes to be relative again?

test/test_support.ts
```diff
@@ -19,8 +19,25 @@ import * as ts from 'typescript';
...
+ * the source tree set up under the hermetic testing environment.
+ */
+function rootDir(): string {
+  const runfiles = process.env['RUNFILES'];
+  assert(runfiles);
+  return path.join(runfiles!, 'tsickle');
+}
+
+/**
+ * Path to tslib.d.ts, relative to the project root.
```
mprobst: it's absolute, isn't it?

test/test_support.ts
```diff
@@ -19,8 +19,25 @@ import * as ts from 'typescript';
...
+ *
+ * The test suite runs the TS compiler, which wants to load tslib.
```
mprobst: I don't understand why we need tslib to begin with - do have test code that has an import of tslib? that does seem odd.

test/test_support.ts
```diff
@@ -37,22 +54,20 @@ export const baseCompilerOptions: ts.CompilerOptions = {
   importHelpers: true,
   noEmitHelpers: true,
   baseUrl: '.',
-  paths: tslibPath ? {
+  paths: {
     // The compiler builtin 'tslib' library is looked up by name,
     // so this entry controls which code is used for tslib.
     'tslib': [tslibPath]
-  } :
-                     undefined,
```
mprobst: yay for fixing formatting by API change! :-)

test/BUILD
evmar: The tests run the TS compiler, which needs tslib.d.ts as an input data file.  I clarified the comment.

The way this works is that if you have importHelpers on, and your code does e.g.
  x = {...y}
then TS wants to use its tslib shim for Object.assign().  To do so it attempts to module resolve 'tslib'.  If tslib isn't found (or if it doesn't have a __assign export) then it's a compile error.

So for the tests to be able properly run TypeScript compilation, they need to ensure that the compiler they run succeeds in resolving tslib.

src/externs.ts
evmar: Which assertion?  It appears this code works in google3 without modifications to its pathToModuleName.

evmar:
Made one final change: the new arg to getGeneratedExterns is now optional, because Angular doesn't pass it and I can't simultaneously change Angular.

evmar:
This appears to work in google3.

src/externs.ts
mprobst: Yeah it works, but I think we should have the "all paths are absolute" assert in the g3 `pathToModuleName` as well, for safety - right?

evmar: Ah I see what you mean, yes.  Right now I still cannot assert paths are absolute:
https://github.com/angular/tsickle/pull/901/files#diff-1310126e2af8f00760656adfc34dc400R38

Our test suite actually manages to sneak the string "goog:foo.bar" into this function, which ends up getting canonicalized into a module name like 'goog_foo.bar' (!).  I want to clean that up to but I figured this change was a good stopping point that marked the next steps clearly.

PR #902, opened 2018-09-29
translate the empty object type {} as *

evmar:
{} in TypeScript means "any value other than null or undefined".
We translated this as Closure {!Object}, but {*} better captures
this.

See the comments for more reasoning.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -715,9 +715,21 @@ export class TypeTranslator {
         }
         return `!Object<${keyType},${this.translate(valType)}>`;
       } else if (!callable && !indexable) {
-        // Special-case the empty object {} because Closure doesn't like it.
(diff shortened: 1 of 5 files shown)


src/type_translator.ts
```diff
@@ -715,9 +715,21 @@ export class TypeTranslator {
...
+        // {string} or {number}.  This would mean you cannot assign a
+        // number to a variable of TS type {}.
+        //
+        // We get closer with {*}, aka the ALL type.  This one better
+        // captures the typical use of the TS {}, which users use for
+        // "I don't care".
+        //
+        // {*} unfortunately does include null/undefined, so it's a closer
+        // match for TS 3.0's 'unknown'.
+        return '*';
```
mprobst: I wonder if we have more places where we return `Object` and mean `{}` or `*`.

PR #906, opened 2018-10-10
restore UPDATE_GOLDENS

evmar:
With my path refactoring, the UPDATE_GOLDENS feature of tsickle broke.
Fix it and also make the code a tiny bit less hacky.

### DIFF
--- a/test/test_support.ts
+++ b/test/test_support.ts
@@ -211,25 +211,23 @@ export class GoldenFileTest {
     return /\.shim\b/.test(this.name);
   }
 
-  /**
-   * Find the absolute path to the tsickle root directory by reading the
-   * symlink bazel puts into bazel-bin back into the test_files directory
-   * and chopping off the test_files portion.
-   */
-  getWorkspaceRoot(): string {
-    if (!this.tsFiles.length || !this.tsFiles[0]) {
-      throw new Error(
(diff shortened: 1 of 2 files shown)


test/test_support.ts
```diff
@@ -211,27 +211,25 @@ export class GoldenFileTest {
...
+ *
+ * To update test golden files (including deleting them when necessary), we need the real path to
+ * the source tree.
+ */
+export function getSourceRoot(): string {
+  // The trick used here is to dereference a symlink for a file that is known to be present in
+  // runfiles, then map the resulting path back the containing directory.
+  const pathInSourceTree = 'node_modules/tslib/tslib.d.ts';
+  const dereferencedPath = fs.readlinkSync(path.join(rootDir(), pathInSourceTree));
+  return dereferencedPath.substr(0, dereferencedPath.length - pathInSourceTree.length - 1);
```
mprobst: I personally try to steer clear of `substr` in JavaScript, as it's subtly different from `substring` but has the same name (yay API design...), and doesn't match with `slice`.

But more generally, isn't this better expressed (more understandable) as `path.relative`?

evmar: I want the reverse of path.relative() here -- I am trying to trim the suffix off and save the part without the relative bit.

PR #908, opened 2018-10-11
add flag for goog.requireType instead of goog.forwardDeclare

evmar:
goog.requireType more closely matches the semantics of a TypeScript
type import.

As part of this change, clean up a few of the comments in the area
and simplify the code a bit.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -711,24 +711,28 @@ export function jsdocTransformer(
       }
 
       function visitImportDeclaration(importDecl: ts.ImportDeclaration) {
-        // No need to forward declare side effect imports.
+        // For each import, insert a goog.requireType for the module, so that if TypeScript does not
+        // emit the module because it's only used in type positions, the JSDoc comments still
+        // reference a valid Closure level symbol.
+
+        // No need to requireType side effect imports.
         if (!importDecl.importClause) return importDecl;
-        // Introduce a goog.forwardDeclare for the module, so that if TypeScript does not emit the
-        // module because it's only used in type positions, the JSDoc comments still reference a
-        // valid Closure level symbol.
+
         const sym = typeChecker.getSymbolAtLocation(importDecl.moduleSpecifier);
         // Scripts do not have a symbol, and neither do unused modules. Scripts can still be
         // imported, either as side effect imports or with an empty import set ("{}"). TypeScript
         // does not emit a runtime load for an import with an empty list of symbols, but the import
         // forces any global declarations from the library to be visible, which is what users use
-        // this for. No symbols from the script need forward declaration, so just return.
+        // this for. No symbols from the script need requireType, so just return.
+        // TODO(evmar): revisit this.  If TS needs to see the module import, it's likely Closure
+        // does too.
         if (!sym) return importDecl;
 
         const importPath = googmodule.resolveModuleName(
             {options: tsOptions, moduleResolutionHost}, sourceFile.fileName,
             (importDecl.moduleSpecifier as ts.StringLiteral).text);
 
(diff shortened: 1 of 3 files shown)


evmar:
Note: this fails CI because we need a new Closure release to expose requireType.  I've already poked at that, so I will rebase this CL once that is available, but I figured I could get the CL ready anyway.

src/jsdoc_transformer.ts
```diff
@@ -772,7 +773,7 @@ export function jsdocTransformer(
         if (importedModuleSymbol) {
           // Forward declare all explicitly imported modules, so that symbols can be referenced and
```
mprobst: xxx forward declare

src/module_type_translator.ts
```diff
@@ -209,83 +208,56 @@ export class ModuleTypeTranslator {
...
     // Already imported? Do not emit a duplicate forward declare.
-    if (this.forwardDeclaredModules.has(moduleSymbol)) return;
+    if (this.requireTypeModules.has(moduleSymbol)) return;
     if (this.host.typeBlackListPaths && this.host.typeBlackListPaths.has(importPath)) {
-      return;  // Do not emit goog.forwardDeclare or goog.require for blacklisted paths.
+      return;  // Do not emit goog.requireType for blacklisted paths.
     }
     const nsImport = googmodule.extractGoogNamespaceImport(importPath);
-    const forwardDeclarePrefix = `tsickle_forward_declare_${++this.forwardDeclareCounter}`;
+    const requireTypePrefix = `tsickle_forward_declare_${this.additionalImports.length}`;
```
mprobst: any opinion on changing the variable name, now that it is no longer a `goog.forwardDeclare`? From a certain angle, it's still conceptually a forward declare. Doesn't matter much either way.

mprobst: Given that we don't care much about the var name, consider `length + 1` to minimize the test diff.

evmar: To minimize the test diff, I left the emitted variable name alone.

src/jsdoc_transformer.ts
```diff
@@ -710,24 +710,26 @@ export function jsdocTransformer(
...
-        // module because it's only used in type positions, the JSDoc comments still reference a
-        // valid Closure level symbol.
+
         const sym = typeChecker.getSymbolAtLocation(importDecl.moduleSpecifier);
         // Scripts do not have a symbol, and neither do unused modules. Scripts can still be
         // imported, either as side effect imports or with an empty import set ("{}"). TypeScript
         // does not emit a runtime load for an import with an empty list of symbols, but the import
         // forces any global declarations from the library to be visible, which is what users use
-        // this for. No symbols from the script need forward declaration, so just return.
+        // this for. No symbols from the script need requireType, so just return.
```
mprobst: we might want to revisit this part in the future. If users import code just for the global symbols it declares, emitting a `goog.requireType` that doesn't assign to a local variable might be the right thing to do; except I don't know if that's supported for `goog.requireType`.

Can you leave a TODO?

src/module_type_translator.ts
```diff
@@ -210,83 +207,56 @@ export class ModuleTypeTranslator {
   }
 
   /**
-   * Returns the `const x = goog.forwardDeclare...` text for an import of the given `importPath`.
-   * This also registers aliases for symbols from the module that map to this forward declare.
+   * Records that we we want a `const x = goog.requireType...` import of the given `importPath`,
+   * which will be inserted when we emit.
+   * This also registers aliases for symbols from the module that map to this requireType.
+   *
+   * @param isExplicitImport True if this comes from an underlying 'import' statement, false
```
mprobst: nice!

PR #910, opened 2018-10-11
consistently avoid emitting inexpressible heritage clauses

evmar:


evmar:
Argh I just want the second commit, but Travis is taking a lonnnng time.

src/jsdoc_transformer.ts
```diff
@@ -112,88 +112,107 @@ export function maybeAddHeritageClauses(
...
-          // do is fail to emit the @implements, which isn't so harmful.
-          continue;
-        }
-        alias = type.symbol;
-      }
-      if (alias.flags & ts.SymbolFlags.Alias) {
-        alias = mtt.typeChecker.getAliasedSymbol(alias);
+      docTags.push({
+        tagName: isExtends ? 'extends' : 'implements',
+        type: parentName || 'InexpressibleType',
```
mprobst: Can you add a bit more commentary on what `InexpressibleType` should achieve on the Closure side? Why do we prefer emitting that over just not emitting anything?

src/jsdoc_transformer.ts
```diff
@@ -112,88 +112,107 @@ export function maybeAddHeritageClauses(
...
-        continue;
+      if (!isExtends) {
+        // Closure can only @implements an interface, not a class.
+        mtt.debugWarn(decl, `omitting @implements of a class: ${expr.getText()}`);
+        return null;
       }
-      // typeToClosure includes nullability modifiers, so call symbolToString directly here.
-      docTags.push({tagName, type: typeTranslator.symbolToString(alias, true)});
+    } else if (alias.flags & ts.SymbolFlags.Value) {
+      // If it's something other than a classin the value namespace, then it will
```
mprobst: type `classin`

src/jsdoc_transformer.ts
```diff
@@ -112,88 +112,107 @@ export function maybeAddHeritageClauses(
...
+      // Handling this correctly is tricky. Closure throws on this
+      // `extends <expression>` syntax (see
+      // https://github.com/google/closure-compiler/issues/2182). We would
+      // probably need to generate an intermediate class declaration and
+      // extend that.
+      mtt.debugWarn(decl, `could not resolve supertype: ${expr.getText()}`);
+      return null;
+    }
+
+    let alias: ts.Symbol = sym;
```
mprobst: my bad, but in the code below `alias` is kind of odd to read. Maybe you can find a way to write this such that below it's just `supertypeSymbol` or so?

src/jsdoc_transformer.ts
```diff
@@ -112,88 +112,107 @@ export function maybeAddHeritageClauses(
...
-      // typeToClosure includes nullability modifiers, so call symbolToString directly here.
-      docTags.push({tagName, type: typeTranslator.symbolToString(alias, true)});
+    } else if (alias.flags & ts.SymbolFlags.Value) {
+      // If it's something other than a classin the value namespace, then it will
+      // not be a type in the Closure output (because Closure collapses
+      // the type and value namespaces).
+      mtt.debugWarn(
+          decl, `omitting heritage reference to a type/value conflict: ${expr.getText()}`);
+      return null;
+    } else if (alias.flags & ts.SymbolFlags.TypeLiteral) {
```
mprobst: I wonder if more generally, we should be whitelisting here instead of blacklisting known bad situations. Not sure if it's worth it.

src/closure_externs.js
```diff
@@ -77,3 +77,10 @@ var PromiseConstructorLike;
 
 /** @typedef {?} */
 var SymbolConstructor;
+
+/**
+ * This type is used as a marker in locations where TypeScript knows the type
+ * but we cannot express that type in the Closure system.
+ * @record
```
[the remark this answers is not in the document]
evmar: Added more comment here.

src/jsdoc_transformer.ts
evmar: I think the earlier comment now covers it.

src/jsdoc_transformer.ts
evmar: PTAL, I rearranged a bit (with no semantic changes so your presubmit should still be good)

test_files/class/class.js
```diff
@@ -151,7 +159,7 @@ class AbstractClassImplementsClass {
 }
 /**
  * @abstract
- * @extends {AbstractClass}
+ * @implements {InexpressibleType}
```
mprobst: Is this change kosher?

test_files/interface/implement_import.js
```diff
@@ -28,7 +29,7 @@ if (false) {
     MyPoint.prototype.y;
 }
 /**
- * @extends {tsickle_forward_declare_1.User}
+ * @implements {InexpressibleType}
```
mprobst: Is this correct?

test_files/class/class.js
evmar: This is changing

```
  class C implements OtherClass
```

where before we did this weird hack around how implements is vaguely the same as 'extends'.  So the diff is showing that I removed the weird hack.  I think it's better to be predictable than fancy here.

PR #911, opened 2018-10-12
Add polymer to JSDOC_TAGS_WHITELIST

aomarks:
https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/Annotation.java#L133


evmar:
CI failure requires you to run a formatter command locally.

PR #914, opened 2018-10-17
Upgrade to TypeScript 3.1.3.

mprobst:
The commits in this PR should be individually reviewable.


mprobst:
Thanks to @weswigham for help with the new `symbolToEntityName` API! :-)

src/type_translator.ts
```diff
@@ -245,12 +245,27 @@ export class TypeTranslator {
       this.ensureSymbolDeclared(sym);
     }
 
-    // This follows getSingleLineStringWriter in the TypeScript compiler.
+    const name = this.typeChecker.symbolToEntityName(
```
evmar: The "other approaches" comments around L233 are probably worth dropping given that they've probably rotted, and the impl comment below here better describes what's going on anyway.

src/type_translator.ts
```diff
@@ -530,10 +522,9 @@ export class TypeTranslator {
         return '?';
       }
       const name = this.symbolToString(type.symbol, /* useFqn */ true);
-      if (name === '(Anonymous class)') {
-        // Values that have anonymous class types produce this name, but the type
-        // appears otherwise identical to a named class.  Given that the type is
-        // anonymous here, there's not really a useful name we can emit.
+      if (name === '?') {
+        // Values that have anonymous class types produce '?'. Make sure not to emit '!?', which is
```
evmar: I'm a little surprised by this change.  I can see that by not using some user-facing 'display builder' we'd no longer get the '(Anonymous class)' string, but I can't see why the above code ever produces the '?' string

src/type_translator.ts

src/type_translator.ts
mprobst: see the `if (!name) return '?';` line just after calling `symbolToEntityName`. That API returns `undefined` for things that cannot be properly named, which we turn into `?`.

mprobst:
See #916 for a minor cleanup followup.

PR #915, opened 2018-10-17
DO NOT REVIEW ts3.1.3 with heritage fix

mprobst:


d0vi:
Why are you guys not using the `^version` format?

mprobst:
@d0vi `tsickle` has TypeScript as a peer dependency, so downstream code can use a TS version it likes together with it. For our project though, we want reproducible builds, which means we want a pinned version. Otherwise each build could pick up a different TS version, and we'd never know if something is actually broken or working.

d0vi:
Oh, thank you for the explanation, @mprobst. I was just asking because the pinned 3.1.3 is giving me warnings when I am working with Angular 7 and TS 3.1.6. I guess I should be using 3.1.3 whenever I have tsickle in my project too.

PR #916, opened 2018-10-18
Refactor `symbolToString`.

mprobst:
- drop the now unused `fqn` parameter.
- push logic to use `?` to usages of `symbolToString`. Whether to emit
  a `?` depends on the use site (e.g. heritage clauses, `!?`), so the
  method should return `string|undefined` and have users choose what to
  do.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -228,12 +228,11 @@ export class TypeTranslator {
   }
 
   /**
-   * Converts a ts.Symbol to a string.
-   * @param useFqn whether to scope the name using its fully qualified name. Closure's template
-   *     arguments are always scoped to the class containing them, where TypeScript's template args
-   *     would be fully qualified. I.e. this flag is false for generic types.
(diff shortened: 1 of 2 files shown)


src/type_translator.ts
```diff
@@ -517,8 +516,8 @@ export class TypeTranslator {
         this.warn('class has no symbol');
         return '?';
       }
-      const name = this.symbolToString(type.symbol, /* useFqn */ true);
-      if (name === '?') {
+      const name = this.symbolToString(type.symbol);
+      if (!name) {
         // Values that have anonymous class types produce '?'. Make sure not to emit '!?', which is
```
evmar: This comment needs a fix.

src/type_translator.ts
```diff
@@ -545,7 +544,7 @@ export class TypeTranslator {
           return '?';
         }
       }
-      return '!' + this.symbolToString(type.symbol, /* useFqn */ true);
+      return '!' + this.symbolToString(type.symbol);
```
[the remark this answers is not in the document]
mprobst: See #917.

PR #918, opened 2018-10-19
Revert "consistently avoid emitting inexpressible heritage clauses (#…

mprobst:
…910)"

This reverts commit e466dc5f2dbabdbc78ff1295e66407cad4f7a68c.


mprobst:
Reason for rollback: this breaks Closure type based optimization because it drops information about certain subtype relationships (`abstract class Child implements AbstractClass {}`), where we used to emit something that made the pattern work.

src/closure_externs.js
```diff
@@ -77,13 +77,3 @@ var PromiseConstructorLike;
...
-
-/**
- * This type is used as a marker in locations where TypeScript knows the type
- * but we cannot express that type in the Closure system.
- *
- * It's used in 'implements' clauses so that a person debugging the emitted JS
- * can see that tsickle saw the type but ignored it.
- * @record
- */
-class InexpressibleType {}
```
evmar: I think one of your CLs used InexpressibleType in its emit, you might need to retain this part?  (If 'git grep InexpressibleType' in your tree after this patch has no matches, then ignore this.)

evmar:
I'm going to try to fix this forward with https://github.com/angular/tsickle/pull/921 , and I will close this if that succeeds.  For now, do not submit.

evmar:
OK, I think this is obsolete.

PR #921, opened 2018-10-20
restore 'implements class' hack

evmar:
tsickle previously converted
  class X implements OtherClass
into
  class X extends OtherClass
in some situations.

In e466dc5f I removed this behavior, but it appears we have code that
actually relies on this.  I now think this behavior was a mistake but for
now let's preserve our behavior so that we can get the other benefits of
this change.


evmar:
You can see this as revert some of the tests here:
https://github.com/angular/tsickle/commit/e466dc5f2dbabdbc78ff1295e66407cad4f7a68c#diff-07f3eb1bb726bfcf47018dd7c18f30ff

PR #922, opened 2018-10-21
fix: don't add comments in json files

alan-agius4:
TypeScript 2.9 introduced the `resolveJsonModule` feature, however `tsickle` will parse json files add the overview comment will will make them invalid json files.

Example:
```
/**
 * @suppress {checkTypes,extraRequire,uselessCode} checked by tsc
 */
{
    "fo": true
}
```

This change will make sure that `.json` files are not parsed.

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -84,6 +84,10 @@ export function transformFileoverviewCommentFactory(diagnostics: ts.Diagnostic[]
     }
 
     return (sourceFile: ts.SourceFile) => {
+      if (sourceFile.fileName.toLowerCase().endsWith('.json')) {
+        return sourceFile;
+      }
+
       const text = sourceFile.getFullText();
 
       let fileComments: ts.SynthesizedComment[] = [];
--- a/test/tsickle_test.ts
+++ b/test/tsickle_test.ts
@@ -21,6 +21,7 @@ describe('emitWithTsickle', () => {
       customTransformers?: tsickle.EmitTransformers): {[fileName: string]: string} {
     const tsCompilerOptions:
         ts.CompilerOptions = {...testSupport.compilerOptions, ...tsConfigOverride};
+    tsCompilerOptions.outDir = path.join(tsCompilerOptions.rootDir!, 'out');
 
     const sources = new Map<string, string>();
     for (const fileName of Object.keys(tsSources)) {
@@ -55,7 +56,7 @@ describe('emitWithTsickle', () => {
     tsickle.emitWithTsickle(
         program, tsickleHost, tsHost, tsCompilerOptions, /* sourceFile */ undefined,
         (fileName: string, data: string) => {
-          jsSources[path.relative(tsCompilerOptions.rootDir!, fileName)] = data;
+          jsSources[path.relative(tsCompilerOptions.outDir!, fileName)] = data;
         },
         /* cancellationToken */ undefined, /* emitOnlyDtsFiles */ undefined, customTransformers);
     return jsSources;
@@ -102,6 +103,29 @@ describe('emitWithTsickle', () => {
         {es5Mode: false, googmodule: false});
 
     expect(jsSources['b.js']).toContain(`export { Foo } from './a';`);
+  });
+
+  it('should emit valid JSON when resolveJsonModule is true', () => {
+    const tsSources = {
+      'a.ts': `
+        import x from './b.json';
+        export { x };
+      `,
+      'b.json': '{"foo": "bar"}',
+    };
+
+    const jsSources = emitWithTsickle(
+        tsSources, {
+          moduleResolution: ts.ModuleResolutionKind.NodeJs,
+          module: ts.ModuleKind.ES2015,
+          resolveJsonModule: true,
+          esModuleInterop: true,
+          allowSyntheticDefaultImports: true,
+        },
+        {es5Mode: false, googmodule: false});
+
+    const jsonContent = JSON.parse(jsSources['b.json']);
+    expect(jsonContent.foo).toBe('bar');
   });
 
   describe('regressions', () => {


test/tsickle_test.ts
```diff
@@ -104,6 +104,32 @@ describe('emitWithTsickle', () => {
...
+  it('should not add comments when using reolveJsonModule in json files', () => {
+    const tsSources = {
+      'a.ts': `
+        import x from './b.json';
+        export { x };
+      `,
+      'b.json': '{"foo": "bar"}',
+    };
+
+    const tmp = path.join(process.cwd(), '.tmp');
```
evmar: Why is this needed?  The other tests don't need a temp directory.

test/tsickle_test.ts
```diff
@@ -104,6 +104,32 @@ describe('emitWithTsickle', () => {
     expect(jsSources['b.js']).toContain(`export { Foo } from './a';`);
   });
 
+  it('should not add comments when using reolveJsonModule in json files', () => {
```
evmar: Typo: resolve

src/fileoverview_comment_transformer.ts
```diff
@@ -80,6 +80,10 @@ export function transformFileoverviewCommentFactory(diagnostics: ts.Diagnostic[]
     }
 
     return (sourceFile: ts.SourceFile) => {
+      if (sourceFile.fileName.endsWith('.json')) {
```
evmar: Do you understand why only this transformer needs changes?  We register a bunch here:
https://github.com/angular/tsickle/blob/master/src/tsickle.ts#L103

test/tsickle_test.ts
alan-agius4: Cause otherwise there will be an error
that I am trying to override the source file. Since after transformation json remains json, while the Ts files will become js files thus they don’t emit this error.

src/fileoverview_comment_transformer.ts
alan-agius4: I think the other transforms do not have an effect on object literals, however this one was adding a comment at top of file which ended up causing an invalid json. 

Not sure what do you think, if we should disable all transformers for json or just this one.

evmar: I think we should disable all transformers for json.

test/tsickle_test.ts
evmar: Can you clarify what the error is?  (Depending on the environment where the tsickle tests run, I think this process.cwd() won't work, so we should figure out some other way of making this test work.)

src/fileoverview_comment_transformer.ts
alan-agius4: So pretty much add this logic to all transformers?

test/tsickle_test.ts
alan-agius4: The error is `TS5055: Cannot write file .../test/unit_test.runfiles/tsickle/b.json' because it would overwrite input file.`

src/fileoverview_comment_transformer.ts
evmar: Or maybe in the place where the transformers are registered?  I'm not sure how this all fits together, you tell me :)

evmar: I see.  I think what you had at first (just special-casing it in the one transfomer that breaks things) is fine then.

test/tsickle_test.ts
alan-agius4: I pushed a fixed. The problem was that no outDir is specified it will emit the transpiled code in the same folder of the source.

That is not a problem when having `ts` to `js` as when transpiling this will not override the source file as it will generate another file. However when having `json` files in the pipeline it will mean that the emitted file will override the source file as they are both named the same with the same extention.

test/tsickle_test.ts
```diff
@@ -21,6 +21,7 @@ describe('emitWithTsickle', () => {
       customTransformers?: tsickle.EmitTransformers): {[fileName: string]: string} {
     const tsCompilerOptions:
         ts.CompilerOptions = {...testSupport.compilerOptions, ...tsConfigOverride};
+    tsCompilerOptions.outDir = path.join(tsCompilerOptions.rootDir!, 'tmp');
```
evmar: I'm confused again.  Does this mean, even without tsickle involved, you cannot use resolveJsonModule without setting an outDir because the compiler wants to overwrite the input files?

alan-agius4: Yes.

evmar: I see, wow!

Can you name this 'out' instead of 'tmp', just in case it shows up in an error message somewhere?

alan-agius4: Sure, done.

tgv1975:
Can someone please review and merge this? It's breaking Angular library compilation, too, if it imports JSON files:

image

The comment on top of the file breaks rollup-plugin-json. Comments in JSON files aren't allowed to begin with, this tool actually adds comments to JSON, instantly invalidating the format.

Please prioritize this. Thank you!

mprobst:
@vikerman is taking a look at why these dependencies keep cropping up.

tgv1975:
Any progress on this? The behavior seems to have remained the same.

mprobst:
I'm afraid nobody is currently working on this. Solving this problem also
doesn't make sense - there's no way to import `.json` files like this in
Closure Compiler, and the sole purpose of `tsickle` is to annotate files
for Closure Compiler.

So I think the Real Bug (tm) is that you somehow depend on tsickle, which
you should not. That's either an Angular problem (@alexeagle?) or a problem
in your local setup though.

alan-agius4:
@mprobst, TypeScript allows to import json as modules; the problem is most likely due that users are using this TSC feature with tsickle in ngpackagr, since in ng-packagr tsickle is used to solve the TDZ of Angular’s forwardRef.

mprobst:
ngpackagr should not use tsickle unless the code it emits is intended to go
through Closure Compiler. There are many, many limitations in tsickle that
are purely there to deal with Closure Compiler. You really, really don't
want to use this tool unless you're specifically targeting Closure.

alan-agius4:
I am getting more keen into adding the transformer in ngpackagr similar to what was done in the CLI

tgv1975:
@mprobst Thanks for your reply. My setup is simply what Angular offers for library compilation and packaging via ngpackagr, as @alan-agius4 also mentioned.

Regardless of the use case, couldn't tsickle simply be prevented to annotate JSON files at all, in no circumstance, since that is against the JSON standard anyway?

mprobst:
You could do that, but it won't fix your problem. Your problem is that
you're using tsickle in a place where you really, really should not. It
brings a host of other issues as well that you might just not have noticed
yet.

tgv1975:
Thing is, I'm not explicitly using tsickle, it is apparently used by Angular to compile and package libraries. However, not allowing tsickle to add comments to JSON (and breaking the JSON spec) would actually solve my problem, since the process fails when invalid JSON comes down the pipe (namely in the Rollup JSON plugin). I have manually edited said plugin to cleanup that tsickle comment, and the compilation passes like a charm.

I am certain you are right and there may be other, hidden, implications of this, but for now I strongly believe that simply not allowing tsickle to corrupt valid JSON with comments, under any circumstances, would go a long way.

tgv1975:
The PR from @alan-agius4 is addressing just that (avoid adding comments to any JSON file) in a straightforward way, as far as I can tell, it simply needs someone to review and merge.

evmar:
What does Closure compiler do when you tell it to import a json file?  I think it will fail, won't it?

tgv1975:
No news on this? Basically one cannot build an Angular library if it happens to import JSON files because of this issue, as simple as that. Would not allowing tsickle corruption of JSON files with comments under any circumstance be too far fetched?

evmar:
Rereading the above comments: I think the conclusion is that if you get into this state, something has gone wrong with your app configuration.  Another way of saying this is that the correct fix for your problem is to not use tsickle.

alexeagle:
re-read what @alan-agius4 said above - ng-packagr is stuck using tsickle because it has a needed feature (decorator downleveling that doesn't introduce a temporal dead zone) but it should not - we ought to have implemented decorator downleveling in a different place (a typescript CustomTransformer as well as something to fix up JS files in node_modules) - @clydin did a hack to port the feature into the Angular webpack loader so that the tsickle dependency was removed from Angular apps, but @alan-agius4 proposes that we also pull it into ng-packagr in order to remove the tsickle dependency from Angular libraries.

tgv1975:
@evmar My configuration is the default, generated with the ng CLI, I did not do any customization, I did not chose to use tsickle. As I said before, at no point did I explicitly involve tsickle. There's nothing I can do there, as far as I can tell.

@alexeagle Thanks for your reply. Does this mean said "hack" will at some point fix ng-packagr in this respect, by making it not use tsickle anymore?

alexeagle:
I referenced https://github.com/ng-packagr/ng-packagr/issues/1400 - that's the actual bug

I'll leave it up to Martin and Evan if they want to close this PR

tgv1975:
@alexeagle Thanks for that, hopefully it will sort out this issue. Until then I'm stuck with manually editing the Rollup JSON plugin to remove the JSON comment every time yarn overwrites it with the original.

evmar:
The fix doesn't seem to bad so if doing the wrong thing here is the best place to fix all the other pieces, we may as well do it I guess.  But now that I think about it: if we do fix this in tsickle, it would only show up in a newer Angular release, right?  And in that case you might as well be using ivy which doesn't involve this(?).

Also I vaguely recall @alexeagle making the tsickle dep toggleable at runtime or something.  Is it possible for @tgv1975 to hand-edit a config file somewhere to turn it off?

(Sorry in advance for my really vague comments here, I don't know much about Angular.)

alexeagle:
Nope, in this case it's not user controlled, the ng-packagr tool always does this.

alan-agius4:
Actually it is under the user controller.

Tsickle can be disable by setting the `annotateForClosureCompiler` compiler
option to false

alan-agius4:
FYI: https://github.com/ng-packagr/ng-packagr/pull/1401

tgv1975:
Great stuff, guys, thank you! @alan-agius4 I found the `annotateForClosureCompiler` switch in the tsconfig.json of the individual libraries, under the `angularCompilerOptions` I guess that's where I turn it of.

@evmar I suppose no tool, ever, should corrupt JSON with comments, regardless of the context, so maybe having this fix (not parsing JSON files at all) is not ill-advised after all, and may prevent unforeseen such incidents down the road.

PR #924, opened 2018-10-22
emit @private/@protected where needed

evmar:
These not only matter for downstream JS code, but also appear to
matter for some optimizations.

### DIFF
--- a/test_files/protected/protected.ts
+++ b/test_files/protected/protected.ts
@@ -1,0 +1,13 @@
+/** This test checks that we emit \@private/\@protected where necessary. */
+
+export {};
+
+class Protected {
+  private privateMember: string;
+  protected protectedMember: string;
+
+  constructor(
+      private anotherPrivate: string,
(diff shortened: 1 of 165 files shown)


evmar:
Chances are low that this will work in google3...

test_files/protected/protected.ts
```diff
@@ -0,0 +1,13 @@
+/** This test checks that we emit @private/@protected where necessary. */
```
mprobst: nit: should you escape the `\@private`? I never know how escaping even works in JSDoc though...

evmar:
I escaped the @private just in case, and also had to adjust the test runner to not fail on unused private fields.

evmar:
PTAL, I had to add a new @suppress.

evmar:
Interestingly, the failing e2e test tipped me off to this -- adding more flag changes to our e2e test is usually wrong, because we need our emit to be accepted by whatever flags users have set.

Issue #907, opened 2018-10-11
JSDoc annotations are prepended with backslash in externs output


Feiyang1:
I have some JSDoc annotations n the source code, e.g. @link, @example. In the externs output, they are escaped by backslash. 
Input:
```javascript 
/**
   * {@link http://example.com} and
   * {@link
   *   https://example.com} for detailed documentation.
   *
   * @example
   * console.log('hello');
  * 
   * @example
   * console.log('hello again');
   * 
*/
export function hello(): void;
```
Output:, notice that all annotations are prepended with backslash and the second @example is missing.
```javascript
/**
 * @externs
 * @suppress {duplicate,checkTypes}
 */
// NOTE: generated by tsickle, do not edit.
// externs from :
/** @const */
var _$$index$d = {};

/**
 * {\@link http://example.com} and
 * {\@link
 *   https://example.com} for detailed documentation.
 * 
 * \@example 
 * console.log('hello');
 *  / 
 * console.log('hello again');
 * 
 * @return {void}
 */
_$$index$d.hello = function() {};
```

Is there a way to keep the comment section as is in the externs?

mprobst:
This is working as intended - Closure Compiler would complain about `@example`, because it is not a standard JSDoc tag. Can you expand on your use case here?

evmar:
Another way to put this is that tsickle is generating compiler syntax (which happens to be written in a thing that looks like jsdoc) but it isn't generating jsdoc.

I think regexing the output might be a hacky way to solve your problem.  I'm not sure about why the @example was deleted, that seems like the sort of bug we might want to fix.  (On the other hand under the reasoning of the previous paragraph I'm not sure why we preserve any comment text that isn't a load-bearing compiler statement.)

mprobst:
I suspect we might have collated the different @example tag content blocks
together? Not sure.

Feiyang1:
Thank you for the insight! Is preserving the comments something you would consider? If not, would you consider fixing the deletion of the second @example ?

I haven't looked into tsdoc yet, but it could be a good option. I'm trying to improve the documentation process incrementally, it would be great if tsickle can help to bridge the gap without changing too much as the first step.

evmar:
I'm not sure I can fix this right now, but we'd take a patch if you wrote one.

Edit this file to include your example, run with the the UPDATE_GOLDENs flag in the README, then tinker with src/jsdoc.ts until it does what you want.
[...]

PR #932, opened 2018-10-26
do not merge jsdoc tags with the same name with flag --noTagMerge

Feiyang1:
Currently JSDoc tags with the same names are merged in the output externs. 
For example, 

```javascript
/**
 *  @example
 *   example 1
 *  @example
 *   example 2
*/
function test() {}
```
becomes
```javascript
/**
 * \@example
 *   example 1 /
 *   example 2
 * @return {void}
 */
function test() { }
```
Notice that @example tags are merged.

This PR adds a new flag --noTagMerge. If set to true, JSDoc tags with the same name will not be merged. This is to address an issue I opened earlier: https://github.com/angular/tsickle/issues/907
With this change, and command  `tsickle --noTagMerge --externs=output.js`
 the output will look like:
```javascript
/**
 * \@example 
 *   example 1
 * \@example 
 *   example 2
 * @return {void}
 */
function test() {}
```

I wanted to add a test case, but I couldn't find any place for module_type_translator.ts. Please let me know how I can do it properly, and I will add the test case.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -61,6 +61,10 @@ export interface AnnotatorHost {
    */
   enableAutoQuoting?: boolean;
   /**
+   * If true, JSDoc tags with the same name are not merged.
+   */
+  noTagMerge?: boolean;
+  /**
    * Whether tsickle should insert goog.provide() calls into the externs generated for `.d.ts` files
    * that are external modules.
    */
--- a/src/main.ts
+++ b/src/main.ts
@@ -31,6 +31,8 @@ export interface Settings {
 
   /** If true, log internal debug warnings to the console. */
   verbose?: boolean;
+
+  noTagMerge?: boolean;
 }
 
 function usage() {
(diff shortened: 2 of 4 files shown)


mprobst:
Didn't look at the code yet, but:

* tests are in subdirectories of `test_files/...`, create a new folder and
a .ts file, then run `UPDATE_GOLDENS=y bazel run //test:golden_test`, then
iterate from there using `bazel test`
* make sure this doesn't break merging of e.g. an `@param` tag when there
are multiple function overloads
* I'm generally hesitant to add more flags. Could we decide on a set of
tags where we don't want the merging, and just always do the right thing?

Feiyang1:
I added a test case in test/tsickle_test.ts. Is it the right place ?
I also tried to add a test file under test_files, but it always creates a js file without using the flag.
How do I tell the test suite to compile the new file with the flag I added?

mprobst:
See my reply by email - I'm not sure we're following the right path here.

evmar:
You cannot control this by a flag, because some tags *must* be merged, such as @return.  We should instead control this in tsickle -- there's no good reason for tsickle to ever merge @example tags, so we should be able to construct a whitelist of tags that should or shouldn't be merged.

evmar:
Should I close this?  Or do you intend to follow up on it?

Feiyang1:
Please close it. I agree it is the wrong direction to go.

PR #933, opened 2018-10-31
emit @private/@protected on abstract methods

evmar:
For various reasons, abstract methods (which have no body)
go down a different code path than the code that handles non-abstract
methods, so my previous change to handle @private/@protected missed
these.

### DIFF
--- a/src/module_type_translator.ts
+++ b/src/module_type_translator.ts
@@ -420,9 +420,16 @@ export class ModuleTypeTranslator {
         addTag(tag);
       }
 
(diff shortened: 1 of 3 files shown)


evmar:
This fails in google3 with

 ERROR - Parse error. extra visibility tag
     * @protected
       ^

Needs more investigation.

mprobst:
Maybe we have existing JSDoc that contains `@protected` tags and we're not properly filtering/deduping?

PR #935, opened 2018-11-13
clean up use of hosts

evmar:
I noticed some code was passing a CompilerHost when it didn't need to.  These separate CLs clean it up a bit.


mprobst:
LGTM.

PR #936, opened 2018-11-13
don't dereference aliases in heritage clauses

evmar:
If A implements B, we shouldn't refer to the type of the B symbol
directly, rather than attempting to resolve B to some other symbol,
because that other symbol may refer to a module that hasn't been
imported.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -114,8 +114,8 @@ export function maybeAddHeritageClauses(
       isExtends: boolean, hasExtends: boolean,
       expr: ts.ExpressionWithTypeArguments): {tagName: string, parentName: string}|null {
     let tagName = isExtends ? 'extends' : 'implements';
-    let sym = mtt.typeChecker.getSymbolAtLocation(expr.expression);
-    if (!sym) {
+    const localSym = mtt.typeChecker.getSymbolAtLocation(expr.expression);
+    if (!localSym) {
       // It's possible for a class declaration to extend an expression that
       // does not have have a symbol, for example when a mixin function is
       // used to build a base class, as in `declare MyClass extends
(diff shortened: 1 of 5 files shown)


evmar:
Gonna try this in g3 to see what happens before I ask for review.

evmar:
@mprobst , I have done some more thinking about how we ought to treat symbol aliases, and I am using this CL as a demo of the general pattern.  Lemme know what you think about the approach.

src/jsdoc_transformer.ts
```diff
@@ -130,29 +130,30 @@ export function maybeAddHeritageClauses(
...
       // Don't emit references to blacklisted types.
       return null;
     }
 
-    if (sym.flags & ts.SymbolFlags.Class) {
+    if (trueSym.flags & ts.SymbolFlags.TypeAlias) {
+      // It's implementing a type alias.  This is illegal in Closure.
+      // We cannot just dereference the type alias, because it might refer
+      // to a symbol that hasn't been imported.
+      mtt.debugWarn(decl, `omitting heritage of type alias: ${expr.getText()}`);
```
mprobst: in another change, we'll need to adapt this code to collect all the heritage clauses we were unable to emit, and then emit the "incorrect assignment" pattern for all of them, right?

src/jsdoc_transformer.ts
```diff
@@ -130,29 +130,30 @@ export function maybeAddHeritageClauses(
...
-      }
-      sym = type.symbol;
-    }
-    if (sym.flags & ts.SymbolFlags.Alias) {
-      sym = mtt.typeChecker.getAliasedSymbol(sym);
+    // localSym can be an alias (e.g. an imported symbol); dealias it before
+    // examining flags, like whether it's a type or a value.  But keep the
+    // original symbol around, because we want to only reference the local
+    // name in any emitted reference.
+    let trueSym = localSym;
```
mprobst: maybe `dereferencedSym`? The other isn't untrue, it's just the local one. Or maybe `declarationSym`, given it's the symbol at the declaration location?

src/jsdoc_transformer.ts
```diff
@@ -173,22 +174,22 @@ export function maybeAddHeritageClauses(
...
+    } else if (trueSym.flags & ts.SymbolFlags.TypeLiteral) {
       // A type literal is a type like `{foo: string}`.
       // These can come up as the output of a mapped type.
       mtt.debugWarn(decl, `omitting heritage reference to a type literal: ${expr.getText()}`);
       return null;
     }
 
     // typeToClosure includes nullability modifiers, so call symbolToString directly here.
-    const parentName = typeTranslator.symbolToString(sym);
+    const parentName = typeTranslator.symbolToString(localSym);
```
mprobst: I'm still wondering about the pattern though where when we `forwardDeclare` (or soon: `requireType`) a module, we're building a mapping from the symbols as declared by the module to unique names. With this here, we're looking up by local symbol to find the name alias. Given tests look green, maybe I'm misunderstanding what symbol the import is pointing to, and all is fine? This might need attaching a debugger and checking the object identity of the symbols involved...

PR #940, opened 2018-11-16
Allow patch level TypeScript peer-dependencies.

mprobst:
Previously, tsickle required a precise peer-dependency version of
`3.1.3`. With this patch, tsickle allows any patch level upgrade to
TypeScript, e.g. 3.1.6.


mprobst:
Dropping this in favor of #941.

PR #945, opened 2018-11-20
use an invalid assignment to communicate type relations

evmar:
drops InexpressibleType in favor of a helper function.  see individual commits.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -60,13 +60,17 @@ export function maybeAddTemplateClause(docTags: jsdoc.Tag[], decl: HasTypeParame
 /**
  * Adds heritage clauses (\@extends, \@implements) to the given docTags for decl. Used by
  * jsdoc_transformer and externs generation.
+ * @return If there are any supertypes that we cannot construct an appropriate Closure
+ *      annotation for, their Closure names are returned.  The caller must emit appropriate
+ *      code to inform the compiler that the types are related in some way.  See the
+ *      various calls to debugWarn() below for cases where this can occur.
  */
 export function maybeAddHeritageClauses(
     docTags: jsdoc.Tag[], mtt: ModuleTypeTranslator,
-    decl: ts.ClassLikeDeclaration|ts.InterfaceDeclaration) {
-  if (!decl.heritageClauses) return;
+    decl: ts.ClassLikeDeclaration|ts.InterfaceDeclaration): string[] {
+  const relatedSuperTypes: string[] = [];
+  if (!decl.heritageClauses) return relatedSuperTypes;
   const isClass = decl.kind === ts.SyntaxKind.ClassDeclaration;
-  const hasExtends = decl.heritageClauses.some(c => c.token === ts.SyntaxKind.ExtendsKeyword);
   for (const heritage of decl.heritageClauses) {
     const isExtends = heritage.token === ts.SyntaxKind.ExtendsKeyword;
     if (isClass && isExtends) {
@@ -83,6 +87,7 @@ export function maybeAddHeritageClauses(
       addHeritage(isExtends, expr);
     }
   }
+  return relatedSuperTypes;
 
   /**
    * Computes the Closure name of an expression occurring in a heritage clause,
@@ -94,7 +99,6 @@ export function maybeAddHeritageClauses(
    * @param isExtends True if we're in an 'extends', false in an 'implements'.
    */
   function addHeritage(isExtends: boolean, expr: ts.ExpressionWithTypeArguments): void {
-    let tagName = isExtends ? 'extends' : 'implements';
     let sym = mtt.typeChecker.getSymbolAtLocation(expr.expression);
     if (!sym) {
       // It's possible for a class declaration to extend an expression that
@@ -134,6 +138,10 @@ export function maybeAddHeritageClauses(
       return;
     }
 
+    // typeToClosure includes nullability modifiers, so call symbolToString directly here.
+    const parentName = typeTranslator.symbolToString(sym);
+    if (!parentName) return;
+
     if (sym.flags & ts.SymbolFlags.Class) {
       if (!isClass) {
         // Closure interfaces cannot extend or implements classes.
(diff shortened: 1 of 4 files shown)


evmar:
Let me know if this seems reasonable to you, I still need to test it.

src/jsdoc_transformer.ts
```diff
@@ -138,59 +123,55 @@ export function maybeAddHeritageClauses(
...
+    // typeToClosure includes nullability modifiers, so call symbolToString directly here.
+    const parentName = typeTranslator.symbolToString(sym);
+    if (!parentName) return;
+
     if (sym.flags & ts.SymbolFlags.Class) {
       if (!isClass) {
         // Closure interfaces cannot extend or implements classes.
         mtt.debugWarn(decl, `omitting interface deriving from class: ${expr.getText()}`);
-        return null;
+        return;
```
mprobst: Do you intentionally not add to `relatedSuperTypes` in this case? Document why so?

src/jsdoc_transformer.ts
```diff
@@ -138,59 +123,55 @@ export function maybeAddHeritageClauses(
...
       }
     } else if (sym.flags & ts.SymbolFlags.Value) {
       // If it's something other than a class in the value namespace, then it will
       // not be a type in the Closure output (because Closure collapses
       // the type and value namespaces).
       mtt.debugWarn(
           decl, `omitting heritage reference to a type/value conflict: ${expr.getText()}`);
-      return null;
+      return;
```
mprobst: Same here, should we still record the relationship?

src/jsdoc_transformer.ts
```diff
@@ -138,59 +123,55 @@ export function maybeAddHeritageClauses(
...
     } else if (sym.flags & ts.SymbolFlags.TypeLiteral) {
       // A type literal is a type like `{foo: string}`.
       // These can come up as the output of a mapped type.
       mtt.debugWarn(decl, `omitting heritage reference to a type literal: ${expr.getText()}`);
-      return null;
+      return;
```
mprobst: I think we could even emit function code for this case.

src/jsdoc_transformer.ts
```diff
@@ -307,6 +288,55 @@ function createMemberTypeDeclaration(
...
+        /* trailing newline */ false);
+
+    const subParam = ts.createParameter(
+        /* decorators */ undefined, /* modifiers */ undefined,
+        /* dotDotDot */ undefined, 'subtype');
+    ts.addSyntheticLeadingComment(
+        subParam, ts.SyntaxKind.MultiLineCommentTrivia, `* !${subType} `,
+        /* trailing newline */ false);
+
+    relationStatements.push(ts.createFunctionDeclaration(
```
mprobst: It looks a bit like `CreateProcessEx`, half a million optional args...

src/jsdoc_transformer.ts
```diff
@@ -307,6 +288,55 @@ function createMemberTypeDeclaration(
...
+  // This is an illegal assignment because the types aren't known to be related,
+  // but because checkTypes is off(!) it instead signals to the compiler that the
+  // types should be treated as if they were related.
+  const relationStatements: ts.Statement[] = [];
+  for (const superType of superTypes) {
+    const superParam = ts.createParameter(
+        /* decorators */ undefined, /* modifiers */ undefined,
+        /* dotDotDot */ undefined, 'supertype');
+    ts.addSyntheticLeadingComment(
+        superParam, ts.SyntaxKind.MultiLineCommentTrivia, `* !${superType} `,
```
mprobst: I've raised a thread on `js-compiler-users@` on whether this syntax is reliable. I'm always a bit spooked out by this (does `/* {foo: string} */` work? Does `/* X<A>|B|* */` work? it's so hard to tell...). I think it's fine to go with this for the time being, but you might also consider re-using our support for JSDoc and `@param` and serializing those, just for consistency.

src/jsdoc_transformer.ts
```diff
@@ -307,6 +288,55 @@ function createMemberTypeDeclaration(
...
+ * Normally you'd do this with \@implements or \@extends, but this is used in
+ * cases where we cannot express the types using those tags.
+ */
+function createSuperTypeRelationship(subType: string, superTypes: string[]): ts.IfStatement {
+  // For each supertype-subtype relationship, we create an unused function that does:
+  //   function tsickleSuperType(/** !Fruit */ supertype, /** !Apple */ subtype) {
+  //     supertype = subtype;
+  //   }
+  // This is an illegal assignment because the types aren't known to be related,
+  // but because checkTypes is off(!) it instead signals to the compiler that the
```
mprobst: I think even if it was on, it'd generate those signals - it being off just prevents the error from being emitted, but the internal optimizer model of JSCompiler still records the fact.

test_files/partial/partial.js
```diff
@@ -1,4 +1,3 @@
-// test_files/partial/partial.ts(7,1): warning TS0: omitting heritage reference to a type literal: Partial<Base>
```
mprobst: Shouldn't this still do something with `Partial<Base>`? I think this is one of the `debugWarn` cases I commented on, isn't it?

test_files/jsdoc_types/jsdoc_types.js
```diff
@@ -53,7 +53,6 @@ let useNeverTyped2;
 let useNeverTypedTemplated;
 /**
  * Note: JSDoc should not reference NeverTyped because the type is blacklisted.
- * @implements {InexpressibleType}
```
mprobst: I wonder if we should emit something like:

```
if (false) {
  function supertype(/** boolean */ supertype, /** ImplementsNeverTyped */ subtype) { supertype = subtype; }
}
```

That is, assign `subtype` into something nonsensical to prevents its properties to be disambiguated? This might not work at all, but we should raise the situation as a question with the Closure team on how to handle this situation. That is, how do we stop a subtype from being disambiguated if we cannot spell out its supertype?

evmar: The part that is scary here is that if we can't spell out its supertype, then not only do we need to prevent it from getting ambiguated, but we *also* need the supertype to not get mangled so that it matches at runtime.  I think in the presence of mapped types in particular this gets pretty gross.

mprobst:
Is this blocked on somethign?

src/jsdoc_transformer.ts
evmar: To answer this question and the rest, I was attempting to not change the rules about which things were related to which in this change, but just the way we mark the relationships we already were capturing.  But it still fails in google3, which I have finally tracked down, so I'll take another shot at this once I see why it still did the wrong thing.

PR #953, opened 2018-12-05
Emit enum types with a `!`.

### LINKED ISSUE #951: Loses non-nullable assertion when importing an Enum
Input:
```ts
// export.ts
export enum E {
  A = 1,
}
```
```ts
// import.ts
import {E} from './export';
class C {
  p: E;
  constructor(p = E.A) {
    this.p = p;
  }
}
```

Actual output:
```js
/**
  * @suppress {checkTypes,extraRequire,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
  */
goog.module('test_files.import_enum.import');
var module = module || { id: 'test_files/import_enum/import.ts' };
module = module;
exports = {};
const tsickle_forward_declare_1 = goog.forwardDeclare("test_files.import_enum.export");
const export_1 = goog.require('test_files.import_enum.export');
class C {
    /**
     * @param {tsickle_forward_declare_1.E.A=} p
     */
    constructor(p = export_1.E.A) {
        this.p = p;
    }
}
if (false) {
    /** @type {tsickle_forward_declare_1.E.A} */
    C.prototype.p;
}
```
`C.p` is optional and its type is restricted to `E.A`.

Expected output:
```js
/**
  * @suppress {checkTypes,extraRequire,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
  */
goog.module('test_files.import_enum.import');
var module = module || { id: 'test_files/import_enum/import.ts' };
module = module;
exports = {};
const tsickle_forward_declare_1 = goog.forwardDeclare("test_files.import_enum.export");
const export_1 = goog.require('test_files.import_enum.export');
class C {
    /**
     * @param {tsickle_forward_declare_1.E.A=} p
     */
    constructor(p = export_1.E.A) {
        this.p = p;
    }
}
if (false) {
    /** @type {!tsickle_forward_declare_1.E} */
    C.prototype.p;
}
```
I tried to replace the Enum with a class and with a type alias. Both worked as expected so the problem might be something in the Enum.

mprobst:
In Closure, enum types are non-null by default - so emitting `/** @type
{MyEnum} */` without a `!` creates a non-nullable type. That's
surprising and outlawed in style guides. For consistency, adjust
tsickle's emit to that style.

Fixes #951.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -499,7 +499,13 @@ export class TypeTranslator {
       this.warn(`EnumLiteralType without a symbol`);
       return '?';
     }
-    return this.symbolToString(enumLiteralBaseType.symbol) || '?';
+    const name = this.symbolToString(enumLiteralBaseType.symbol);
+    if (!name) return '?';
+    // In Closure, enum types are non-null by default, so we wouldn't need to emit the `!` here.
+    // However that's confusing to users, to the point that style guides and linters require to
+    // *always* specify the nullability modifier. To be consistent with that style, include it here
+    // as well.
+    return '!' + name;
   }
 
   // translateObject translates a ts.ObjectType, which is the type of all
--- a/test_files/enum/enum.js
+++ b/test_files/enum/enum.js
@@ -18,26 +18,28 @@ EnumTest1[EnumTest1.XYZ] = 'XYZ';
 EnumTest1[EnumTest1.PI] = 'PI';
 // Verify that the resulting TypeScript still allows you to index into the enum with all the various
 // ways allowed of enums.
-/** @type {EnumTest1} */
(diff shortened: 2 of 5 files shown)


src/type_translator.ts
```diff
@@ -499,7 +499,13 @@ export class TypeTranslator {
       this.warn(`EnumLiteralType without a symbol`);
       return '?';
     }
-    return this.symbolToString(enumLiteralBaseType.symbol) || '?';
+    const name = this.symbolToString(enumLiteralBaseType.symbol)
```
bowenni: missing `;`

mprobst: Done!

PR #955, opened 2018-12-06
Handle rest parameters with template constraints.

### LINKED ISSUE #944: Function declaration throws invalid rest param error
tsickle 0.33.1 and 0.34.0 throws "Error: invalid rest param"

```
declare function setImmediate<T extends unknown[]>(
  callback: (...args: T) => void,
  ...args: T
): number;
```

I am using it in context of Angular

<pre>
Angular CLI: 7.0.5
Node: 10.12.0
OS: darwin x64
Angular: 7.0.3
... animations, common, compiler, compiler-cli, core, forms
... http, language-service, platform-browser
... platform-browser-dynamic, router

Package                            Version
------------------------------------------------------------
@angular-devkit/architect          0.10.5
@angular-devkit/build-angular      0.10.5
@angular-devkit/build-ng-packagr   0.10.5
@angular-devkit/build-optimizer    0.10.5
@angular-devkit/build-webpack      0.10.5
@angular-devkit/core               7.0.5
@angular-devkit/schematics         7.0.5
@angular/cli                       7.0.5
@ngtools/json-schema               1.1.0
@ngtools/webpack                   7.0.5
@schematics/angular                7.0.5
@schematics/update                 0.10.5
ng-packagr                         4.4.0
rxjs                               6.3.3
typescript                         3.1.6
webpack                            4.19.1
</pre>

mprobst:
Rest parameters must be array types in TypeScript. tsickle unwraps them
to emit the underlying type for Closure. However the Array type in
TypeScript may be expressed in a template parameter constraints, instead
of the parameter itself:

    function f<T extends SomeType[]>(... ts: T) {}

Handle this specific case by resolving the template parameter
constraint, then emitting the underlying type nested in it.

Fixes #944.

### DIFF
--- a/test_files/functions/functions.ts
+++ b/test_files/functions/functions.ts
@@ -34,6 +34,20 @@ TestSplat3(1, 2);
 function defaultBeforeRequired(x = 1, y: number, ...z: any[]) {}
 defaultBeforeRequired(undefined, 2, 'h', 3);
 
+// The array reference below happens on the template parameter constraint, not on the parameter
+// itself. Make sure tsickle unwraps the right type.
+function templatedBoundRestArg<T extends string[]>(...str: T) {
+  return str.length;
+}
+// But only do so if the parameter is not an array reference type by itself.
+function templatedBoundRestArg2<T extends string>(...str: T[]) {
+  return str.length;
+}
+// Also handle the case where it's both.
+function templatedBoundRestArg3<T extends number[]>(...str: T[]) {
+  return str.length;
+}
+
 function templated<T>(t: T): number {
   return 1;
 }
--- a/src/module_type_translator.ts
+++ b/src/module_type_translator.ts
@@ -465,9 +465,20 @@ export class ModuleTypeTranslator {
           // In TypeScript you write "...x: number[]", but in Closure
           // you don't write the array: "@param {...number} x".  Unwrap
           // the Array<> wrapper.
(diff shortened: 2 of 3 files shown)


test_files/functions/functions.ts
```diff
@@ -34,6 +34,20 @@ TestSplat3(1, 2);
 function defaultBeforeRequired(x = 1, y: number, ...z: any[]) {}
 defaultBeforeRequired(undefined, 2, 'h', 3);
 
+// The array reference below happens on the template parameter constraint, not on the parameter
+// itself. Make sure tsickle unwraps the right type.
+function templatedBoundRestArg<T extends string[]>(...str: T) {
```
evmar: I tried this on the playground and got:
ERROR(1,52): : A rest parameter must be of an array type.

mprobst: Maybe it's a version issue? This works (and we do also check for diagnostics in these tests): 

http://www.typescriptlang.org/play/#src=function%20templatedBoundRestArg%3CT%20extends%20string%5B%5D%3E(...str%3A%20T)%20%7B%20%7D%0D%0AtemplatedBoundRestArg('a'%2C%20'b')%3B

evmar: Wow it fails on TS29, and works on TS31!  (I was looking at this to try to understand what the behavior is...)

PR #956, opened 2018-12-07
Remove _Instance redirections for clutz symbols.

rkirov:
Since
https://github.com/angular/clutz/commit/b86a51c6ec5d910c7d7b177fcefe8c3c1ae28eb8
clutz no longer splits classes into _Instance and static side.


evmar:
Does this break any user code?

evmar:
Oh it's just for the clutz symbols, I misread this as removing the _Instance hack altogether.

PR #957, opened 2018-12-07
Blacklist based on a symbol's declaration path.

mprobst:
When emitting forward declarations, tsickle would previously inspect the
import path given at the use site. However the list of blacklisted
locations is a set of (absolute) paths of files that contain symbols
that must not have types emitted for them. These paths can but do not
have to be related to import paths - for example for ambient external
modules (`declare module 'ambient' {...}`), the import path has no
relationship to the declaration site's path.

The fix is to check the declaration site forof the import's module
symbol, which gives the correct path.


src/jsdoc_transformer.ts
```diff
@@ -722,8 +722,7 @@ export function jsdocTransformer(
         // forces any global declarations from the library to be visible, which is what users use
         // this for. No symbols from the script need forward declaration, so just return.
         if (!sym) return importDecl;
-        // Write the export declaration here so that forward declares come after it, and
```
mprobst: FYI this comment was just a misleading leftover, unrelated to the main change here.

src/module_type_translator.ts
```diff
@@ -219,7 +219,7 @@ export class ModuleTypeTranslator {
     if (this.host.untyped) return;
     // Already imported? Do not emit a duplicate forward declare.
     if (this.forwardDeclaredModules.has(moduleSymbol)) return;
-    if (this.host.typeBlackListPaths && this.host.typeBlackListPaths.has(importPath)) {
+    if (typeTranslator.isBlacklisted(this.host.typeBlackListPaths, moduleSymbol)) {
       return;  // Do not emit goog.forwardDeclare or goog.require for blacklisted paths.
```
bowenni: Change the comment to `// ... symbols.`?

mprobst: Done.

Issue #961, opened 2018-12-22
Enum output differences between es5 and es2015 (Declaration Merging Support)


dallastjames:
So, I'll prefix this with the fact that this is my first foray into typescript compiler code, so if I'm just missing obvious things, I apologize in advance. :+1: 

I posted a bug in the [ng-packagr library](https://github.com/ng-packagr/ng-packagr/issues/680) back in March related to an issue during compiling with declaration merging. Since it hasn't received any solutions, I figured I'd try my hand at tracking down the cause of the error. The tl;dr of that issue is that enum declaration merging with namespaces throws a `Identifier {name} has already been declared` error. This is the initial code based on the [typescript declaration merging documentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces-with-classes-functions-and-enums)

```Typescript
enum AdapterType {
    MapValue = '[Adapter Type] map value',
    MapClass = '[Adapter Type] map class'
}
namespace AdapterType {
    /**
     * Contains an ordered array of the valid AdapterType members
     */
    export const members: AdapterType[] = [AdapterType.MapClass, AdapterType.MapValue];
}

export { AdapterType };
```

And this is the output when compiling to `es2015`.

```Typescript
// es2015, es6, etc.
/**
 * @fileoverview added by tsickle
 * @suppress {checkTypes,extraRequire,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
 */
/** @enum {string} */
const AdapterType = {
    MapValue: '[Adapter Type] map value',
    MapClass: '[Adapter Type] map class',
};
var AdapterType;
(function (AdapterType) {
    AdapterType.members = [AdapterType.MapClass, AdapterType.MapValue];
})(AdapterType || (AdapterType = {}));
export { AdapterType };
```

You'll notice that the enum is declared initially with the `const` keyword, which means that the code that follows is invalid since you can't redeclare a `const`. The `es5` version of this code that gets output uses the var keyword initially, which allows the build to continue successfully. 

```Typescript
// es5
/**
 * @fileoverview added by tsickle
 * @suppress {checkTypes,extraRequire,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
 */
/** @enum {string} */
var AdapterType = {
    MapValue: '[Adapter Type] map value',
    MapClass: '[Adapter Type] map class',
};
var AdapterType;
(function (AdapterType) {
[...]

PR #962, opened 2018-12-28
fix: adjust enum transpile to generate var if not const enum

dallastjames:
matches tsc output to correctly support declaration merging.

This would close this [issue](https://github.com/angular/tsickle/issues/961) I opened in reference to supporting declaration merging. It adjusts any enum that is not a `const enum` to use the `var` declaration after transpile instead of the `const` declaration so that declaration merging works correctly. Since declaration merging is not possible with a `const enum`, those are left with the `const` declaration. Updated the appropriate test files to reflect this change.


evmar:
Do we know that Closure supports reopening enums in this manner?  I don't really understand the rules for how enums work.

PR #968, opened 2019-01-09
Handle rest parameters explicitly typed as any

KenanSulayman:
This adds special handling for rest parameter in function signatures explicitly typed as `any`.

Example code:

```
type CallbackMaps = {
    [k: string]: ( ( ...args: any ) => void );
};
```

Before this change, the `convertParams` method of the type translator would throw, specifically because the possibility of a null type of the type arguments of the type of a declaration in the method signature is explicitly silenced (https://github.com/angular/tsickle/blob/master/src/type_translator.ts#L824). This changeset identifies this condition and falls back to a `?` type annotation.

I didn't follow the exact logic of the type translator test (specifically `isBuiltinLibDTS`), therefore I have not added a test case at this time.


mprobst:
Nice!

Can you add a test case? Create a new TS file, e.g. `test_files/rest_parameters_any/rest_parameters_any.ts`, then run `UPDATE_GOLDENS=y bazel run //test:golden_test`, then check if the generated `.js` file looks good and add both to your change.

mprobst:
@KenanSulayman is this still current? Would you mind adding a test (maybe you forgot to `git add` the files)?

KenanSulayman:
Sorry for the lack of communication! I was trying to add the respective tests and run them using the golden_test command, but constantly ran into a silent error (the new test file is listed, but doesn’t execute correctly and is marked as red). I thought this might be related to the feature I was testing and therefore copied another test; it still failed so I gave up. Is there a way I can enable a verbose mode on blaze?

KenanSulayman:
This time the tests went through. I included some very generic tests related to this issue.

After rebasing to master I couldn't replicate my issue, so I reverted my additions to tsickle. To my suprise it worked just fine, so I reverted my previous commit.w

mprobst:
Odd, but maybe we fixed something related in the mean time without noticing. Better test coverage is better in any case, so thanks for the contribution!

PR #971, opened 2019-01-17
Emit aliases for exports of a module in externs.

mprobst:
If a namespace re-exports symbols with different names, tsickle must
generate a matching externs entry so that the symbol does not get
renamed by Closure Compiler. This comes up in the Chrome externs:

    module chrome {
      module _debugger { ... }
      export {_debugger as debugger};
    }

This works so far. However when converting user code to Closure output,
TypeScript cannot quite emit the correct name for
`chrome.debugger.Something`, because it doesn't take the re-export in
the chrome module (namespace) into account. See `user.js` for the
incorrect emit.

While this fix is partial then, it's still better than nothing: it'll
prevent renaming of the re-exported symbol (`debugger`), and Closure
Compiler will ignore the incorrect type annotation and back off due to
`@suppress {checkTypes}`.

See also See https://github.com/Microsoft/TypeScript/issues/29459.


evmar:
Did you drop a file?  I don't see user.js in this

mprobst:
Whoops, forgot to add the test dir as usual.

src/externs.ts
```diff
@@ -451,6 +451,25 @@ export function generateExterns(
...
+    }
+    for (const exportSpecifier of exportDeclaration.exportClause.elements) {
+      if (!exportSpecifier.propertyName) {
+        emit(`\n// TODO(tsickle): unnamed export of ${exportSpecifier.name.text} in ${
+            locationStr(exportDeclaration, namespace)}`);
+        continue;
+      }
+      emit('/** @const */\n');
+      writeVariableStatement(
+          exportSpecifier.name.text, namespace, exportSpecifier.propertyName.text);
```
evmar: Will this do the wrong thing when name == propertyName?  Or is propertyName null in that case and we've already bailed out above?
Maybe add an "export {foo}" to the test just to show it?

src/externs.ts
```diff
@@ -580,6 +599,10 @@ export function generateExterns(
     return [];
   }
 
+  function locationStr(node: ts.Node, namespace: ReadonlyArray<string>) {
```
evmar: Can you one-liner doc this?  It looks wrong to me but I see it's only for debugging below.

src/type_translator.ts
```diff
@@ -240,7 +240,7 @@ export class TypeTranslator {
     }
 
     const name = this.typeChecker.symbolToEntityName(
-        sym, ts.SymbolFlags.None, this.node, ts.NodeBuilderFlags.UseFullyQualifiedType);
+        sym, ts.SymbolFlags.Type, this.node, ts.NodeBuilderFlags.UseFullyQualifiedType);
```
evmar: What is this for?

src/externs.ts
mprobst: Done.

src/externs.ts
mprobst: Added docs. In my defense, I'm just moving this code around, and you originally added it ;-)

src/type_translator.ts
mprobst: I was trying to fix the emit problem mentioned in the description. While doing that I learned that we should set `Type` here, because that's how we want to use the name (as opposed to as a `Value`). It doesn't have any effect though, just an API use cleanup.

PR #972, opened 2019-01-18
build: switch to circleci

devversion:


.circleci/config.yml
```diff
@@ -1,17 +1,31 @@
...
+    paths:
+      - "node_modules"
+      - "~/bazel_repository_cache"
+
 jobs:
   build:
-    working_directory: ~/tsickle
-    docker:
-      - image: angular/ngcontainer:0.0.1
+    <<: *job_defaults
```
mprobst: given we have only one job, do we need this indirection here?

.circleci/config.yml
```diff
@@ -1,17 +1,31 @@
...
+var_1: &cache_key angular-tsickle-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
+var_2: &default_docker_image l.gcr.io/google/bazel:0.18.0
+
+# Settings common to each job
+var_3: &job_defaults
+  working_directory: ~/tsickle
+  docker:
+  - image: *default_docker_image
+
+# Saves the cache for the current Yarn lock file. We store the node modules and the Bazel
```
mprobst: not just for the yarn lock file, right?

.circleci/config.yml
```diff
@@ -1,17 +1,31 @@
-version: 2
+# Cache key for CircleCI. We want to invalidate the cache whenever the Bazel workspace or the
+# NPM dependencies changed.
+var_1: &cache_key angular-tsickle-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
+var_2: &default_docker_image l.gcr.io/google/bazel:0.18.0
```
mprobst: What's the version number here? is this another place that we have to bump the version when we move to a newer bazel?

Can you add a comment?

devversion: Done.

mprobst:
Thanks Paul!

PR #973, opened 2019-01-21
Fix a missing line break.

mprobst:
The missing line break can cause parse errors if the next lines starts
with a block comment that is then ignored due to the line comment.

This is a follow up to #971.


mprobst:
Sorry, missed your comment. Let's try with yet another PR ;-)

PR #978, opened 2019-01-30
carefully handle @template in jsdoc handling

evmar:
Some @tags are disallowed in the user's *input* jsdoc, like
@private, because we don't want users writing them.

Other @tags are escaped in the tsickle *output* jsdoc, because
we don't want Closure compiler interpreting them.

It turns out @template is special in that we want to allow it
in the input, and we want tsickle to be able to generate it
in the output, but we don't want the user's input to make it
to the output.

See the test case for some more rationale.

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -46,12 +46,18 @@ export interface Tag {
 
 /**
  * A list of all JSDoc tags allowed by the Closure compiler.
+ * All tags other than these are escaped before emitting.
+ *
+ * Note that some of these tags are also rejected by tsickle when seen in
+ * the user-provided source, but also that tsickle itself may generate some of these.
+ * This whitelist is just used for controlling the output.
+ *
  * The public Closure docs don't list all the tags it allows; this list comes
  * from the compiler source itself.
  * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/Annotation.java
  * https://github.com/google/closure-compiler/blob/master/src/com/google/javascript/jscomp/parsing/ParserConfig.properties
  */
-const JSDOC_TAGS_WHITELIST = new Set([
+const JSDOC_TAGS_OUTPUT_WHITELIST = new Set([
   'abstract',
   'argument',
   'author',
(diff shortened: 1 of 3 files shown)


src/jsdoc.ts
```diff
@@ -46,12 +46,18 @@ export interface Tag {
 
 /**
  * A list of all JSDoc tags allowed by the Closure compiler.
+ * All tags other than these are escaped before emitting.
+ *
+ * Note that some of these tags are also rejected by tsickle when seen in
+ * the user-provided source, but also that tsickle itself may generate some of these.
+ * This whitelist just is used for controlling the output.
```
rkirov: s/just is/is just/

src/jsdoc.ts
```diff
@@ -213,9 +220,20 @@ export function parseContents(commentText: string): ParsedJSDocComment|null {
...
       }
       let type: string|undefined;
-      if (JSDOC_TAGS_BLACKLIST.has(tagName)) {
-        warnings.push(`@${tagName} annotations are redundant with TypeScript equivalents`);
-        continue;  // Drop the tag so Closure won't process it.
+      if (JSDOC_TAGS_INPUT_BLACKLIST.has(tagName)) {
+        if (tagName !== 'template') {
+          // Tell the user to not write blacklisted tags, because there is TS
+          // syntax available for them.
+          warnings.push(`@${tagName} annotations are redundant with TypeScript equivalents`);
```
rkirov: does this warning have some other context when printed to the user?

src/jsdoc.ts
```diff
@@ -213,9 +220,20 @@ export function parseContents(commentText: string): ParsedJSDocComment|null {
...
+          continue;  // Drop the tag so Closure won't process it.
+        } else {
+          // But @template in particular is special: it's ok for the user to
+          // write it for documentation purposes, but we don't want the
+          // user-written one making it into the output because Closure interprets
+          // it as well.
```
rkirov: Document that for simplicity we don't check that the template type in the "doc-only" comment, name really matches the TS declared template.

test_files/jsdoc/jsdoc.ts
```diff
@@ -115,4 +115,15 @@ const somePolymerBehavior = {};
...
+/**
+ * This class has a 'template' tag, which we want to allow (because this is
+ * how to doc this) but not let Closure interpret (because we emit our own).
+ * The desired behavior is that the user-written @template comment (which
+ * talks about T) is dropped, but the tsickle-generated @template comment
+ * (which talks about T2) is preserved.
+ *
+ * @template T User-written comments on the template.
+ */
+class Foo<T2> {}
```
rkirov: Should we also have a test for the proper usage of this "feature", i.e. matching TS and doc template name.

src/jsdoc.ts
evmar: Yeah, it shows up via the TS api so it points at the relevant file/line etc.

src/jsdoc.ts
evmar: Done.

test_files/jsdoc/jsdoc.ts
evmar: Done.

PR #985, opened 2019-02-03
drop flag for requireType

evmar:
We now successfully emit requireType(), so we no longer need
the forwardDeclare() fallback code.


src/module_type_translator.ts
```diff
@@ -252,53 +245,14 @@ export class ModuleTypeTranslator {
                 ts.createCall(
                     ts.createPropertyAccess(
                         ts.createIdentifier('goog'),
-                        USE_REQUIRE_TYPE ? 'requireType' : 'forwardDeclare'),
```
mprobst: optional: you could consider changing the variable name with this from `tsickle_forward_declare_XXX` to `requireType` or better. I recently thought it might help us debug a bit if we didn't just number the forward declares but also used the last bit of the module name, but that might be overkill.

evmar: I think I'll do this, but in another change, because it causes a huge amount of test skew (every reference to module names change).

PR #988, opened 2019-02-04
Handle renamed exports in Clutz declarations.

mprobst:
When a symbols is renamed in its export specifier, tsickle must use the
module local name to generate a Clutz declaration, not the exported
name. Paraphrasing:

    export {X as Y};
    // --> should generate -->
    declare global {
      namespace ಠ_ಠ.clutz {
        export {X as module$contents$file_Y}
      }
    }

### DIFF
--- a/test_files/reexport.declaration/export.ts
+++ b/test_files/reexport.declaration/export.ts
@@ -1,1 +1,2 @@
 export const NUM_CONSTANT = 3;
+export const OTHER = 1;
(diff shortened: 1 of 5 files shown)


test_files/reexport.declaration/export.ts
```diff
@@ -1 +1,2 @@
 export const NUM_CONSTANT = 3;
+export const OTHER = 1;
```
evmar: Shouldn't this be some {OTHER as BAR} thing?

mprobst:
Ah, I forgot to git add the new test file (as usual...). Yes, there's an
export as in that one.

PR #989, opened 2019-02-06
 Handle bigint.

mprobst:
Currently this refers to a type in closure_externs.js, to be replaced with a proper Closure externs definition.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -420,7 +420,7 @@ export class TypeTranslator {
       case ts.TypeFlags.Undefined:
         return 'undefined';
       case ts.TypeFlags.BigInt:
-        return 'bigint';
+        return 'bigintPlaceholder';
       case ts.TypeFlags.Null:
         return 'null';
       case ts.TypeFlags.Never:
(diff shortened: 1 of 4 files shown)


src/type_translator.ts
```diff
@@ -419,6 +419,8 @@ export class TypeTranslator {
         return 'void';
       case ts.TypeFlags.Undefined:
         return 'undefined';
+      case ts.TypeFlags.BigInt:
+        return 'bigint';
```
evmar: Is it worth naming this something like "bigintPlaceHolder"?

I'm thinking that when Closure does implement support for bigint, landing it might require simultaneously removing our externs definition above.  Where if we instead emitted some slightly-off type, then we can cycle tsickle and closure separately.

Maybe this is easy enough to work around either way though, your call.

test_files/type/type.js
```diff
@@ -21,6 +21,8 @@ let typeArr2;
 let typeNestedArr;
 /** @type {*} */
 let typeUnknown;
+/** @type {bigint} */
```
evmar: This should be using placeholder

mprobst:
Done!

Am Mi., 6. Feb. 2019 um 19:54 Uhr schrieb Evan Martin <
notifications@github.com>:

PR #990, opened 2019-02-08
Add a conditional check for splat on er.emittedFiles.

rkirov:
Under some compilations this was type error without the conditional
check.


rkirov:
Replaced by https://github.com/angular/tsickle/pull/992

PR #991, opened 2019-02-09
Clutz patch exported

rkirov:


src/tsickle.ts
```diff
@@ -227,8 +227,33 @@ function addClutzAliases(
...
+    for (const d of e.declarations) {
+      // This check helps us filter out export * rewritten exports. It appears,
+      // those point to the source file without any need to get through the
+      // getExportSpecifierLocalTargetSymbol API below.
+      if (d.getSourceFile() !== origSourceFile) {
+        return false;
+      }
+
+      // export specifiers are ambigious - import {X}; export {X} vs
+      // export {X}, but everything else is a local export.
```
evmar: I couldn't quite parse this,   export {X} vs export {X}?  Add some newlines?

src/tsickle.ts
```diff
@@ -227,8 +227,33 @@ function addClutzAliases(
...
+      if (!ts.isExportSpecifier(d)) {
+        return true;
+      }
+
+      // Using ts API to resolve to a local symbol.
```
evmar: I don't get what this comment is saying.  Maybe comment this whole block with what it's trying to achieve?  Something about looking for a matching local symbol in the correct file?

test_files/reexport.declaration/reexport_named.ts
```diff
@@ -0,0 +1,3 @@
+// reexport_named.ts exports symbols, but doesn't declare any of them, so its .d.ts
```
evmar: "declare" is a bit overloaded here, is there a better word for it?  "declare them in local scope"?

src/tsickle.ts
```diff
@@ -225,8 +227,33 @@ function addClutzAliases(
     // default is a keyword in typescript, so the name of the export being default means that it's a
     // default export.
     if (e.name === 'default') return false;
+
     // Otherwise check that some declaration is from the local module.
-    return e.declarations.some(d => d.getSourceFile() === origSourceFile);
+    for (const d of e.declarations) {
+      // This check helps us filter out export * rewritten exports. It appears,
```
mprobst: I think a symbol that is just exported from another file does not have a local declaration. Note that you also cannot reference it locally:

```
export {Foo} from './foo';
let f: Foo; // error
```

So it follows their declaration site is the original file.

src/tsickle.ts
```diff
@@ -225,8 +227,33 @@ function addClutzAliases(
...
+      // those point to the source file without any need to get through the
+      // getExportSpecifierLocalTargetSymbol API below.
+      if (d.getSourceFile() !== origSourceFile) {
```
mprobst: so, rephrasing, and maybe you can adjust the comments: if (1) a symbol is just a straight export, its declaration is in the original file, so the original check (this one) works. Then if (2) the export is renaming (`export {X as Y}`) then we want to consider this a local name. However if (3) it's an `import` followed by an `export` we don't want to do so.

Right?

I don't understand the reason behind (3) though. It looks to me like it's the same as (2) - sure, it's the same name, but in a different module, so it's the same as giving it a new name from that angle? I think TS allows you to reference an imported `X`:

```
import {X} from './foo';
export {X};
let y2: X;
```

test_files/reexport.declaration/reexport_named.ts
mprobst: but why? What does this fix?

rkirov: Without my change this test emits clutz aliases which are invalid. Admittedly, I still don't have a great understanding of how TS APIs model all these cases, hence the confusing comments and code (I have no idea what I am doing). But I can tell you that this test is legitimately broken before my change.

src/tsickle.ts
rkirov: That is not the case, and hence the bug that I am trying to fix. Here is example debug output, printing for each test case the location of the export.declaration[0] (they all have a single declaration).

```
.===IN FILE export.ts
SYMBOL "NUM_CONSTANT" flags:0x2 BlockScopedVariable
decl loc export.ts
SYMBOL "OTHER" flags:0x2 BlockScopedVariable
decl loc export.ts
SYMBOL "NON_DIRECT" flags:0x200000 AliasExcludes
decl loc export.ts
SYMBOL "NEW_NAME" flags:0x200000 AliasExcludes
decl loc export.ts
===IN FILE import_reexport.ts
SYMBOL "NUM_CONSTANT" flags:0x200000 AliasExcludes
decl loc import_reexport.ts
===IN FILE reexport.ts
SYMBOL "NUM_CONSTANT" flags:0x2 BlockScopedVariable
decl loc export.ts
SYMBOL "OTHER" flags:0x2 BlockScopedVariable
decl loc export.ts
SYMBOL "NON_DIRECT" flags:0x200000 AliasExcludes
decl loc export.ts
SYMBOL "NEW_NAME" flags:0x200000 AliasExcludes
decl loc export.ts
===IN FILE reexport_named.ts
SYMBOL "OTHER" flags:0x200000 AliasExcludes
decl loc reexport_named.ts
SYMBOL "NUM_CONSTANT" flags:0x200000 AliasExcludes
decl loc reexport_named.ts
===IN FILE renamed_export.ts
SYMBOL "MOTHER" flags:0x200000 AliasExcludes
decl loc renamed_export.ts
.
```

Only in the case of 'export *' and actual local export the declaration points to the original file. In all other cases the declaration is in the reexporting file, which made me add this extra code using the getExportSpecifierLocalTargetSymbol API.

src/tsickle.ts
rkirov: I think the renaming bits can be handled separately. The issue is purely about (3) vs pure reexport (without import).

src/tsickle.ts
rkirov: reworded the comments.

src/tsickle.ts
rkirov: reworded the comments.

test_files/reexport.declaration/reexport_named.ts
rkirov: reworded

src/tsickle.ts
```diff
@@ -217,18 +222,64 @@ function addClutzAliases(
...
+  // For now we 'fix' 2) by simply not emitting a clutz alias, since clutz
+  // interop is used in minority of scenarios.
+  //
+  // TODO(radokirov): attempt to add appropriate imports for 2) so that
+  // currently finding out local appears even harder than fixing exports.
   const localExports = moduleExports.filter(e => {
-    // If there are no declarations, be conservative and emit the aliases.
-    if (!e.declarations) return true;
+    // If there are no declarations, be conservative and don't emit the aliases.
+    // I don't know how can this happen, we have no tests that excercise it.
```
mprobst: I think this can happen for builtin types, like "boolean"?

src/tsickle.ts
```diff
@@ -217,18 +222,64 @@ function addClutzAliases(
...
+      if (d.getSourceFile() !== origSourceFile) {
+        return false;
+      }
+
+      if (!ts.isExportSpecifier(d)) {
+        // we are in case 1) thus safe to emit clutz alias.
+        return true;
+      }
+
+      // Original declarations are useless to separate 2) and 3) because they
```
mprobst: not sure what original means here?

src/tsickle.ts
```diff
@@ -217,18 +222,64 @@ function addClutzAliases(
 
   // .d.ts files can be transformed, too, so we need to compare the original node below.
   const origSourceFile = ts.getOriginalNode(sourceFile);
-  // The module exports might be re-exports, and in the case of "export *" might not even be
-  // available in the module scope, which makes them difficult to export. Avoid the problem by
-  // filtering out symbols who do not have a declaration in the local module.
+  // In order to write aliases, the exported symbols need to be available in the
+  // the module scope. That is not always the case:
+  //
+  // 1) export const X;           // works
```
mprobst: nit: (1) (2) (3) below is a bit hard to understand, maybe name these "export", "reexport" and "imported reexport"?

src/tsickle.ts
rkirov: e.declarations in the current context. Reworded.

src/tsickle.ts
rkirov: My hunch was that the confusion comes from us being sloppy with the words 'export' and 'reexport'. Does 'export' include a 'reexport', etc? Anyways, gave it a show, using only words, not sure if it is clearer. Time will tell.

PR #992, opened 2019-02-11
Update to latest Bazel rules

alexeagle:
- Updates required version of Bazel to latest
- Now use Bazel-managed deps (Bazel runs yarn install, user does not have to)
- Use fine-grained deps, rather than every action depending on all node_modules
- Global (ambient) typings for jasmine are now declared per-file so this typing does not pollute non-test code
- rules_typescript now comes purely as an npm package

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,9 +2,9 @@
 # NPM dependencies changed.
 var_1: &cache_key angular-tsickle-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
 # Default docker image for CircleCI jobs with Bazel installed. The version of Bazel is controlled
-# by the docker image version. e.g "google/bazel:0.18.0" installs Bazel v0.18.0. When updating this
+# by the docker image version. e.g "google/bazel:0.22.0" installs Bazel v0.22.0. When updating this
 # version, consider updating the minimum required Bazel version in the "WORKSPACE" file.
-var_2: &default_docker_image l.gcr.io/google/bazel:0.18.0
+var_2: &default_docker_image l.gcr.io/google/bazel:0.22.0
 
 # Settings common to each job
 var_3: &job_defaults
--- a/BUILD
+++ b/BUILD
@@ -13,12 +13,12 @@ filegroup(
     srcs = glob([
         "test_files/**/*",
         "third_party/**/*",
+    ]) + [
         # The test suites run version-specific TypeScript compilers,
-        # but they all need the 'tslib' package to be available in
-        # a node_modules directory *above* the test/ directory.  So
-        # we must provide this extra directory as data to those tests.
-        "node_modules/tslib/*",
-    ]),
+        # but they all need the 'tslib' package to be available. So
+        # we must provide this extra filegroup as data to those tests.
+        "@npm//tslib:tslib__files",
+    ],
 )
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "npm_package")
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -61,7 +61,9 @@ export function mergeEmitResults(emitResults: EmitResult[]): EmitResult {
   for (const er of emitResults) {
     diagnostics.push(...er.diagnostics);
     emitSkipped = emitSkipped || er.emitSkipped;
-    emittedFiles.push(...er.emittedFiles);
+    if (er.emittedFiles) {
+      emittedFiles.push(...er.emittedFiles);
+    }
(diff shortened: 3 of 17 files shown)


src/tsickle.ts
```diff
@@ -61,7 +62,7 @@ export function mergeEmitResults(emitResults: EmitResult[]): EmitResult {
   for (const er of emitResults) {
     diagnostics.push(...er.diagnostics);
     emitSkipped = emitSkipped || er.emitSkipped;
-    emittedFiles.push(...er.emittedFiles);
+    emittedFiles.push(...(er.emittedFiles as string[]));
```
evmar: I vaguely recall Rado has a patch for this, do you know why it fails?  I think we concluded to put a `!` on it instead of the cast.

test/e2e_clutz_dts_test.ts
```diff
@@ -6,6 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
+///<reference types="jasmine"/>
```
evmar: In every instance but this one you had a blank line after this import

BUILD
```diff
@@ -13,12 +13,12 @@ filegroup(
...
+    ]) + [
         # The test suites run version-specific TypeScript compilers,
-        # but they all need the 'tslib' package to be available in
-        # a node_modules directory *above* the test/ directory.  So
-        # we must provide this extra directory as data to those tests.
-        "node_modules/tslib/*",
-    ]),
+        # but they all need the 'tslib' package to be available. So
+        # we must provide this extra filegroup as data to those tests.
+        "@npm//tslib:tslib__files",
```
evmar: Sorry about the crazy hacks here

src/tsickle.ts
alexeagle: it failed because the type of `er.emittedFiles` is `unknown`. Happy to take some other fix

BUILD
alexeagle: s'okay, I took a stab at removing the special handling for tslib but it was pretty easy to just update the location instead

test/e2e_clutz_dts_test.ts
alexeagle: if the formatter doesn't format it then how can a human possibly do it? :man_shrugging:

src/tsickle.ts
alexeagle: changed to non-null assertion operator, it works too

rkirov: see https://github.com/angular/tsickle/pull/990

test/test_support.ts
```diff
@@ -37,7 +38,7 @@ function rootDir(): string {
  * tslib if we let it do its normal module resolution.  So instead we fix
  * the path here to the known path to the desired file.
  */
-const tslibPath = path.join(rootDir(), 'node_modules/tslib/tslib.d.ts');
+const tslibPath = path.join(rootDir(), '../npm/node_modules/tslib/tslib.d.ts');
```
rkirov: This won't sync nicely into g3. Why is this change needed?

.circleci/config.yml
```diff
@@ -32,5 +32,5 @@ jobs:
           key: *cache_key
       - run: bazel run @nodejs//:yarn
       - run: bazel build ...
-      - run: bazel test ...
+      - run: bazel test ... --nolegacy_external_runfiles
```
rkirov: any comment about this flag, so that we know when is it safe to remove?

src/tsickle.ts
rkirov: Let's get this in first, and I will do the "proper fix" with #990 afterwards.

test/test_support.ts
evmar: We can fix the g3 sync after, no big deal

src/tsickle.ts
alexeagle: I went ahead and applied your same change here too rather than check in an intermediate state

test/test_support.ts
alexeagle: This change is needed because the old path assumed that node_modules was a directory in our repository (meaning you manually installed it before running bazel)
Now that bazel manages the node modules, it lays them out in an external repository called npm

.circleci/config.yml
alexeagle: Under current bazel, the external files are layed out in two places:
`runfiles_root/my_proj/external/npm/node_modules`
`runfiles_root/npm/node_modules`

They will eventually stop doing the first one. I added this flag to ensure that we are forwards compatible, opting into that breaking change. We could leave it here indefinitely.
Added a comment

PR #999, opened 2019-03-11
Handle rest arguments resolves to zero args.

mprobst:
If a rest argument type is instantiated with no arguments, its type
paramters array is undefined (not empty). This change avoids a crash in
that situation, and emits a zero argument function.

### DIFF
--- a/test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.js
+++ b/test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.js
@@ -1,0 +1,24 @@
+// test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.ts(3,1): warning TS0: var args type is not an object type
+// test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.ts(3,1): warning TS0: var args type is not an object type
+/**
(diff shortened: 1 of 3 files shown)


test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.js
```diff
@@ -0,0 +1,24 @@
...
+ * @fileoverview added by tsickle
+ * @suppress {checkTypes,extraRequire,missingOverride,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
+ */
+goog.module('test_files.rest_parameters_generic_empty.rest_parameters_generic_empty');
+var module = module || { id: 'test_files/rest_parameters_generic_empty/rest_parameters_generic_empty.ts' };
+module = module;
+exports = {};
+/**
+ * @template A
+ * @param {function(!Array<?>): void} fn
```
rkirov: We inline the lower bound? Then why emit `@template A` at all?

mprobst: We unconditionally emit `@template A`. We don't really inline the bound, we just fail to handle rest arguments that use template type variables from the outer scope.

PR #1000, opened 2019-03-11
Avoid emitting double `@this` types.

### LINKED ISSUE #997: Double @this generated with this returns + this type
This (ha!) generates a duplicated `@this` tag:

```ts
class Y {}
class X { foo(this: Y): this { ... } }
```

Along the lines of:

```ts
class X {
  /**
   * @template THIS
   * @this {THIS}
   * @this {Y}
   * @template THIS
  */ 
  foo(this: Y): this { ... } }
```

The first `@this` is generated by the return-value `foo(): this` handling, with the second one being generated by the parameter `this: Y` line. We must only emit one for valid code - presumably the explicit `this: Y` is the correct one.

Given that we cast every access to `this` within the method body, I actually wonder why we emit `@this {THIS}` in the first place, maybe we don't even need that?

mprobst:
TypeScript methods can specify a `this` parameter type together with a
`this` return type, e.g. `foo(this: Bar): this`, which caused tsickle to
emit double `@this` tags, which is an error in Closure.

This change fixes #997 by prefering the `this` parameter type over the
templatized return type.

### DIFF
--- a/src/module_type_translator.ts
+++ b/src/module_type_translator.ts
@@ -411,11 +411,13 @@ export class ModuleTypeTranslator {
       if (sig.declaration.kind === ts.SyntaxKind.JSDocSignature) {
         throw new Error(`JSDoc signature ${fnDecl.name}`);
       }
+      let hasThisParam = false;
       for (let i = 0; i < sig.declaration.parameters.length; i++) {
         const paramNode = sig.declaration.parameters[i];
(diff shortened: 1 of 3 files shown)


mprobst:
@rkirov it's much more powerful and more decomposed. But indeed, it's also much harder to use.

PR #1001, opened 2019-03-11
Emit inline symbols in `.d.ts` without `declare`.

### LINKED ISSUE #996: .d.ts externs files don’t include anonymous types from params/returns
If I have a `.d.ts` file that looks like this:

```typescript
export class Thing {
  call(param: {key: string}): {value: string};
}
```

the externs that are generated don’t include the `key` as part of the object:

```javascript
/**
 * @externs
 * @suppress {duplicate,checkTypes}
 */
// externs from experimental/users/pfish/tsickle/example.d.ts:
/** @const */
var google3$experimental$users$pfish$tsickle$example = {};
/**
 * @constructor
 * @struct
 */
google3$experimental$users$pfish$tsickle$example.Thing = function() {};

/**
 * @param {?} param
 * @return {?}
 */
google3$experimental$users$pfish$tsickle$example.Thing.prototype.call = function(param) {};
```

```typescript
const someThing = new Thing();
// while .call will work, both `key` and `.value` may be renamed by Closure compiler.
someThing.call({key: '5'}).value;
```

I would expect that `key` and `value` would be included in the externs somehow to prevent their renaming.

The equivalent code of:

```typescript
export class Thing {
  call(param: CallParam): CallReturn;
}

export interface CallParam {
  key: string;
}

export interface CallReturn {
  value: string;
}
```

does produce the appropriate extern declarations.

mprobst:
tsickle tries to carefully not emit non-ambient types in externs. To do
so, it looks for the `declare` keyword on type declarations and their
ancestors.

However in `.d.ts` files the `declare` keyword is optional, so tsickle
should also take into account whether the file itself is a declaration
file.

Fixes #996.


src/externs.ts
```diff
@@ -437,6 +437,7 @@ export function generateExterns(
     for (const methodVariants of Array.from(methods.values())) {
       const firstMethodVariant = methodVariants[0];
       let parameterNames: string[];
+      debugger;
```
thetorpedodog: stray `debugger` statement!

test_files/declare_anonymous_type/declare_anonymous_type.d.ts
```diff
@@ -0,0 +1,4 @@
+// `ambientFn` is ambient, but has no `declare` keyword (which is legal TS). This test makes sure
```
rkirov: It is ambient because it lives in a '.d.ts' file. It wasn't clear to me at first, add that to the comment.

rkirov: I tried the same thing with a non-module, but it is a TS error:
```TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.```

src/externs.ts
mprobst: whoops, thanks!

test_files/declare_anonymous_type/declare_anonymous_type.d.ts
mprobst: Yeah, odd isn't it? I think that's a usability bug, should be consistent.

PR #1002, opened 2019-03-12
Fix a crash in addClutzAliases. 

mprobst:
`localSymbol.declarations` can be undefined for builtin symbols that
have no declarations, such as `unknown`, despite TypeScript's type
definition not betraying the fact.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -270,6 +270,8 @@ function addClutzAliases(
       const localSymbol = typeChecker.getExportSpecifierLocalTargetSymbol(d);
       // I don't know how can this happen, but err on the side of less emit.
       if (!localSymbol) return false;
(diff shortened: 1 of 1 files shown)


src/tsickle.ts
```diff
@@ -269,7 +269,9 @@ function addClutzAliases(
       // one.  However, there is another ts API that can do a deeper resolution.
       const localSymbol = typeChecker.getExportSpecifierLocalTargetSymbol(d);
       // I don't know how can this happen, but err on the side of less emit.
-      if (!localSymbol) return false;
+      if (!localSymbol) return;
```
rkirov: `return false` to be more explicit.

mprobst: Fixed.

PR #1003, opened 2019-03-14
Do not emit @type if @define is specified

shicks:
Instead, just trust whatever type is specified in @define, which should be either string, number, or boolean.

This is a blocker for rolling out the new module-compatible style of `goog.define` to TypeScript code.

### DIFF
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -150,15 +150,23 @@ const JSDOC_TAGS_INPUT_BLACKLIST = new Set([
   'record',   'static',     'template',   'this',        'type',      'typedef',
 ]);
 
+/** What to do with types found in JSDoc @tags. */
+enum TagWithTypePolicy {
+  FORBIDDEN,
+  REQUIRED,
+}
+
 /**
  * A list of JSDoc @tags that might include a {type} after them. Only banned when a type is passed.
  * Note that this does not include tags that carry a non-type system type, e.g. \@suppress.
  */
-const JSDOC_TAGS_WITH_TYPES = new Set([
-  'const',
-  'export',
-  'param',
-  'return',
+const JSDOC_TAGS_WITH_TYPES = new Map([
+  ['const', TagWithTypePolicy.FORBIDDEN],
+  ['define', TagWithTypePolicy.REQUIRED],
+  ['export', TagWithTypePolicy.FORBIDDEN],
+  ['param', TagWithTypePolicy.FORBIDDEN],
+  ['return', TagWithTypePolicy.FORBIDDEN],
+  ['suppress', TagWithTypePolicy.REQUIRED],
 ]);
 
 /**
@@ -234,17 +242,21 @@ export function parseContents(commentText: string): ParsedJSDocComment|null {
           // Drop it without any warning.  (We also don't ensure its correctness.)
           continue;
         }
-      } else if (JSDOC_TAGS_WITH_TYPES.has(tagName) && text[0] === '{') {
-        warnings.push(
-            `the type annotation on @${tagName} is redundant with its TypeScript type, ` +
-            `remove the {...} part`);
-        continue;
-      } else if (tagName === 'suppress') {
-        const suppressMatch = text.match(/^\{(.*)\}(.*)$/);
-        if (!suppressMatch) {
-          warnings.push(`malformed @suppress tag: "${text}"`);
+      } else if (JSDOC_TAGS_WITH_TYPES.has(tagName)) {
+        const policy = JSDOC_TAGS_WITH_TYPES.get(tagName);
+        if (policy === TagWithTypePolicy.FORBIDDEN && text[0] === '{') {
+          warnings.push(
+              `the type annotation on @${tagName} is redundant with its TypeScript type, ` +
+              `remove the {...} part`);
+          continue;
+        }
+        const typeMatch = text.match(/^\{(.*)\}(.*)$/);
+        if (!typeMatch) {
+          if (policy === TagWithTypePolicy.REQUIRED) {
+            warnings.push(`malformed @${tagName} tag: "${text}"`);
+          }
         } else {
-          [, type, text] = suppressMatch;
+          [, type, text] = typeMatch;
         }
       } else if (tagName === 'dict') {
         warnings.push('use index signatures (`[k: string]: type`) instead of @dict');
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -153,6 +153,7 @@ const JSDOC_TAGS_INPUT_BLACKLIST = new Set([
 /** What to do with types found in JSDoc @tags. */
 enum TagWithTypePolicy {
   FORBIDDEN,
+  OPTIONAL,
   REQUIRED,
 }
 
@@ -162,7 +163,7 @@ enum TagWithTypePolicy {
  */
 const JSDOC_TAGS_WITH_TYPES = new Map([
   ['const', TagWithTypePolicy.FORBIDDEN],
-  ['define', TagWithTypePolicy.REQUIRED],
+  ['define', TagWithTypePolicy.OPTIONAL],  // NOTE: this could be made forbidden later.
   ['export', TagWithTypePolicy.FORBIDDEN],
   ['param', TagWithTypePolicy.FORBIDDEN],
   ['return', TagWithTypePolicy.FORBIDDEN],
--- a/src/jsdoc.ts
+++ b/src/jsdoc.ts
@@ -158,12 +158,13 @@ enum TagWithTypePolicy {
 }
 
 /**
- * A list of JSDoc @tags that might include a {type} after them. Only banned when a type is passed.
- * Note that this does not include tags that carry a non-type system type, e.g. \@suppress.
+ * JSDoc \@tags that might include a {type} after them. Specifying a type is forbidden except in the
+ * case of \@suppress, whose argument is a suppression type, rather than a type system type.
  */
 const JSDOC_TAGS_WITH_TYPES = new Map([
   ['const', TagWithTypePolicy.FORBIDDEN],
-  ['define', TagWithTypePolicy.OPTIONAL],  // NOTE: this could be made forbidden later.
+  // NOTE: passing a type to @define will soon be forbidden as well, after usages are migrated.
+  ['define', TagWithTypePolicy.OPTIONAL],
   ['export', TagWithTypePolicy.FORBIDDEN],
   ['param', TagWithTypePolicy.FORBIDDEN],
   ['return', TagWithTypePolicy.FORBIDDEN],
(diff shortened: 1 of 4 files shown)


src/jsdoc_transformer.ts
```diff
@@ -602,8 +602,9 @@ export function jsdocTransformer(
             tags = null;
           }
           // Add an @type for plain identifiers, but not for bindings patterns (i.e. object or array
-          // destructuring) - those do not have a syntax in Closure.
-          if (ts.isIdentifier(decl.name)) {
+          // destructuring - those do not have a syntax in Closure) or @defines, which already
```
mprobst: What about propagating the TypeScript type into the `@define` tag? I think that leaves less chance for error (e.g. `/** @define {number} */` let mydefine = 'string';` would be a bit of a foot gun). In reality this is probably always string, but generally speaking if we can avoid having types specified in two locations, we should.

So in the block below:

```typescript
if (!blackListedInitialized) {
  const defineTag = localTags.find(...);
  if ([define tag has type specified]) reportDiagnostic(...);
  localTags.push({tagName: defineTag ? 'define' : 'type', type: typeStr});
}
```

You'll also need to add `@define` to the list of tags conflicting with `@type`, so that the original tag gets escaped and we don't end up with two `@define` tags in the emit.

shicks: The escape-list doesn't quite work as you might hope.  I made the JSDOC_TAGS_WITH_TYPES Set into a Map with an enum value type indicating what to *do* with the type (forbidden or required - we could add optional at some point if that became relevant).  And then I just overwrite the `@define` with the inferred type rather than escaping it, with a warning if it didn't match.

src/jsdoc.ts
```diff
@@ -150,15 +150,24 @@ const JSDOC_TAGS_INPUT_BLACKLIST = new Set([
...
-  'export',
-  'param',
-  'return',
+const JSDOC_TAGS_WITH_TYPES = new Map([
+  ['const', TagWithTypePolicy.FORBIDDEN],
+  ['define', TagWithTypePolicy.OPTIONAL],  // NOTE: this could be made forbidden later.
+  ['export', TagWithTypePolicy.FORBIDDEN],
+  ['param', TagWithTypePolicy.FORBIDDEN],
+  ['return', TagWithTypePolicy.FORBIDDEN],
+  ['suppress', TagWithTypePolicy.REQUIRED],
```
rkirov: Add a comment that this is required, because the thing in the type position is not a "type"

src/jsdoc.ts
```diff
@@ -150,15 +150,24 @@ const JSDOC_TAGS_INPUT_BLACKLIST = new Set([
...
 
+/** What to do with types found in JSDoc @tags. */
+enum TagWithTypePolicy {
+  FORBIDDEN,
+  OPTIONAL,
+  REQUIRED,
+}
+
 /**
  * A list of JSDoc @tags that might include a {type} after them. Only banned when a type is passed.
```
rkirov: No longer 'a list'

src/jsdoc_transformer.ts
```diff
@@ -615,7 +616,19 @@ export function jsdocTransformer(
...
+              // If @define is present then do not add @type, but instead verify that any type is
+              // correct. Specifying any type at all here may become an error later.
+              const defineTag = localTags.find(({tagName}) => tagName === 'define');
+              if (!defineTag) {
+                localTags.push({tagName: 'type', type: typeStr});
+              } else if (defineTag.type !== typeStr) {
+                if (defineTag.type !== undefined) {
+                  moduleTypeTranslator.error(
+                      varStmt,
+                      `incorrect type in @define: found '${defineTag.type}' required '${typeStr}'`);
```
rkirov: Since one of those is inferred the comment can be confusing. I would add some more text like:

'required' -> 'TypeScript type was'
'found' -> 'written Closure type was'

Otherwise it looks too much like a TS type checking error, which it is not.

test_files/jsdoc/jsdoc.ts
```diff
@@ -128,3 +128,18 @@ Polymer({behaviors: ['test' as any]});
...
+
+/**
+ * @define {boolean}
+ */
+const BAD_DEFINE = 42;
+
+/**
+ * @define
+ */
+const MISSING_DEFINE = false;
```
rkirov: More like TS_INFERRED_CLOSURE_MISSING_DEFINE

Maybe add also:
```ts
/**
 * @define
 */
const TS_EXPLICIT_CLOSURE_MISSING_DEFINE: boolean = false;
```

src/jsdoc.ts
shicks: Added to the top-level doc.

src/jsdoc.ts
shicks: Done.

src/jsdoc_transformer.ts
shicks: Done.

test_files/jsdoc/jsdoc.ts
shicks: Changed the names a bit - missing was back from before the type was optional, so it was actually an error.

shicks:
Addressed the points.  PTAL.

src/jsdoc_transformer.ts
```diff
@@ -616,7 +616,20 @@ export function jsdocTransformer(
...
+              // correct. Specifying any type at all here may become an error later.
+              const defineTag = localTags.find(({tagName}) => tagName === 'define');
+              if (!defineTag) {
+                localTags.push({tagName: 'type', type: typeStr});
+              } else if (defineTag.type !== typeStr) {
+                if (defineTag.type !== undefined) {
+                  moduleTypeTranslator.error(
+                      varStmt,
+                      `incorrect type in @define: written Closure type was '${
+                          defineTag.type}' but TypeScript inferred '${typeStr}'`);
```
mprobst: nit: inferred is a bit misleading here, it could so be an explicitly declared type.

src/jsdoc_transformer.ts
```diff
@@ -616,7 +616,20 @@ export function jsdocTransformer(
               // getOriginalNode(decl) is required because the type checker cannot type check
               // synthesized nodes.
               const typeStr = moduleTypeTranslator.typeToClosure(ts.getOriginalNode(decl));
-              localTags.push({tagName: 'type', type: typeStr});
+              // If @define is present then do not add @type, but instead verify that any type is
```
mprobst: `s/any/the declared type (if any)/` ? `any` is an overloaded term in TypeScript types :-)

src/jsdoc.ts
mprobst: Code seems fine as is. On corner case issues such as this (not too many uses of `@define` in TS), it's also been possible to just sync and fix users in one CL, which I think would save us the intermediary state (and the `enum` and `Map` here).

shicks: I think the enum and map have value on their own.  I'm maybe a little concerned that if we mark it forbidden, then the `@define` is completely gone: `/** @define {string} */` would end up as `/** @type {string} */`, which means something very different.  But maybe it's fine because it should lead to an error somewhere later, I guess.

src/jsdoc_transformer.ts
shicks: No longer applies

src/jsdoc_transformer.ts
shicks: No longer applies

src/jsdoc.ts
shicks: And it's just a warning here, rather than an error.

I notice that this also causes things like `/** @const {string} */` to lose their constness, etc.  Would we be open to retaining the annotation as if it had had no type (so ignoring the braces but not the annotation itself) or doing the same sort of checking as I had in an earlier revision (where it's an error if the types don't match)?  This would be tough for `@param` but all the other forbidden tags-with-types would work fine for that check.

src/jsdoc_transformer.ts
shicks: Changed to 'found'

shicks: Deleted again.

shicks:
This should be ready to merge, though it will need some cleanup internally to go along with any CL that brings it in, since all the types on `@define`s are now warnings.

src/jsdoc.ts
mprobst: Within Google (and bazel builds), these warnings are actually hard errors.

Not sure what you mean with `@const` or similar tags being dropped, they are currently retained in the output, aren't they?

The general idea here is if at all possible avoid a discrepancy between types seen by Closure and types on the TypeScript level. For comments, that means if at all possible, scrubbing or error'ing on all Closure annotations that contain types.

Ideally I think we'd remove all load-bearing comments on the TS level. It's confusing to have first level syntax for types, but still needing to fall back to comments, plus comments are very fickle. That'll require design time decorators or similar (for things such as `@export`). That's a longer term project though.

In either case, I'm happy to land this code as is, but I'm not clear on the value of the `enum`/`Map` indirection. It makes sense if we anticipate adding more tags with special behavior in the long run (not just `suppress`); do you have a plan there? Otherwise I think the plain `if` statement captures better what's going on. I'm happy to make that change myself though.

shicks:
No, if you look at the jsdoc.js test case, you'll see the badConstThing property declared as `@type {string}` instead of `@const {string}`.  It doesn't matter if warnings are promoted to errors, but if folks are running this without promoting warnings to errors then this could be unfortunate.

With only a single required annotation and no optional, I'm less sold on the win with the map, though in my opinion it is more readable: (a) using an enum rather than an inline condition makes it clear why it's doing what it's doing, and (b) it's potentially less repetition of the condition, depending on how it's structured.  But I tried removing it and the diff looks a little cleaner, so I'm fine with that.

In terms of adding more in the future, we have a few internal annotations around mods that take arguments and might want to go here.  Though I can also see an argument for adding those in a separate Set JSDOC_TAGS_WITH_NON_TYPE_ARGS (or just ..._WITH_ARGS).

mprobst:
> No, if you look at the jsdoc.js test case, you'll see the badConstThing property declared as @type {string} instead of @const {string}. It doesn't matter if warnings are promoted to errors, but if folks are running this without promoting warnings to errors then this could be unfortunate.

Yes, tsickle behaves like TS in here (and like JSCompiler) in that it produces output even for incorrect user code. The question is how much effort we should put into emitting something that'll still likely work. With tsickle in general and `@define` in particular, I'd wager that essentially all users are in bazel or google3, and thus always have these warnings emit errors. So effectively, users should be safe.

There's a question whether we should have ever had this warnings mechanism, I'm not convinced it's useful for anyone. Then again we apparently have some users who use tsickle as a preprocessor to produce JSDoc, they don't care about the effects on JSCompiler at all. Can't really pick your use cases ...

I'll take this over and land it.

shicks:
Thanks!

PR #1007, opened 2019-03-21
Omit this parameters from paramTags

alur:
The jsdoc tag-merging function currently throws when there is a mix of overloads and this parameters, because the this tag is treated as a parameter here.


rkirov:
A test would help me review this easier, also good to have for maintenance. Is there anything tricky around writing a test for this?

alur:
Added a golden test.

It's not too tricky, I'm just unfamiliar with the codebase, and very rusty working with git :)

PR #1008, opened 2019-03-27
Clean up requireType generation.

mprobst:
Instead of `tsickle_forward_declare_X`, emit a unique string generated
based on the import path name. This should make debugging the generated
Closure source code easier, by helping tie back import prefixes to the
imported modules.

The algorithm strips common file extensions, takes the last path or
namespace step, ignoring `/index`, and replaces non-ident characters.

While at it:
- drop some unused imports and variables.
- drop the unused isExplicitImport parameter.


rkirov:
Lgtm

PR #1014, opened 2019-04-09
Add suppression for constantProperty check in tsickle emit.

rkirov:
We are doing this mostly for supporting namespaces. When one declares
namespace foo.bar the emit implicilty reopens foo, which results in a `
|| export.foo = {}` potential reassignment in the emit. See added test
for the actual code.

In Closure export.foo is implicitly const, so this triggers an error.
For now, we just suppress it using @supress, since TS has checked the
const-ness of the original file.

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -58,6 +58,10 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
   suppressions.add('unusedPrivateMembers');
   // 5) Suppress checking for @override, because TS doesn't model it.
   suppressions.add('missingOverride');
+  // 6) Suppress constantProperty checking, which errors when a namespace is
+  // reopened. Namespace reopening happens when one writes namespace foo {}
+  // or namespace foo.* {} more than once.
+  suppressions.add('constantProperty');
   suppressTag.type = Array.from(suppressions.values()).sort().join(',');
 
   return tags;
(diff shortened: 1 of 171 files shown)


src/fileoverview_comment_transformer.ts
```diff
@@ -58,6 +58,9 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
   suppressions.add('unusedPrivateMembers');
   // 5) Suppress checking for @override, because TS doesn't model it.
   suppressions.add('missingOverride');
+  // 6) Suppress constantProperty checking, which errors when a namespace is
```
LucasSloan: Explain why/when namespaces get reopened?

test_files/namespaced/reopen_ns.ts
```diff
@@ -0,0 +1,12 @@
+export namespace ns {
+  export let x = 0;
+}
+
+export namespace ns {
+  export let y = 0;
+}
+
+// this implicitly re-emits exports.ns, so it is in a way reopening ns.
```
LucasSloan: Sorry, I don't understand how this implicitly re-emits the other namespace.

src/fileoverview_comment_transformer.ts
rkirov: done

test_files/namespaced/reopen_ns.ts
rkirov: see now

PR #1029, opened 2019-06-14
remove direct dependencies on node builtin modules

evmar:
For 'assert', just replace it with a throw.

For 'path', use the TypeScript's path manipulation functions.  These
are likely better than using node's path functions anyway because we
interact heavily with TypeScript and using their functions ensures we
match.  On the downside, we must access internal functions to do this.

This is a step towards making it possible to run tsickle in a browser
playground.

### DIFF
--- a/src/externs.ts
+++ b/src/externs.ts
@@ -65,7 +65,6 @@
  *   whatever third party library the .d.ts describes).
  */
 
-import * as path from 'path';
 import * as ts from 'typescript';
 
 import {AnnotatorHost, moduleNameAsIdentifier} from './annotator_host';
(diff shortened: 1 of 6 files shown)


src/externs.ts
```diff
@@ -616,7 +616,7 @@ export function generateExterns(
    * cases.
    */
   function debugLocationStr(node: ts.Node, namespace: ReadonlyArray<string>) {
-    return namespace.join('.') || path.basename(node.getSourceFile().fileName);
+    return namespace.join('.') || node.getSourceFile().fileName.replace(/.*[/\\]/, '');
```
rkirov: add a comment because regexps are hard to read.

src/path.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+/**
+ * @fileoverview Path manipulation functions.
+ * These are the functions exposed by nodejs in the 'path' module.
```
rkirov: AlexE was talking about tsickle supporting multiple versions of the TypeScript compiler simultaneously because that's how Angular wants to use it. This might make this harder to achieve, but then again I don't think it is a good idea to begin with :/

src/externs.ts
evmar: Done.

PR #1030, opened 2019-06-15
always emit type aliases to closure

evmar:
Before, we tried to be clever and for
  export type X = ...
we'd emit something, but then for just
  type X = ...
without an export, we'd skip emit for Closure.

But that trivially fails when code is like
  type X = ...
  export {X};
where the Closure emit for the export needs to refer to the "local-only" X.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -657,9 +657,6 @@ export function jsdocTransformer(
         // namespaces, the two emits would conflict if tsickle emitted both.
         const sym = moduleTypeTranslator.mustGetSymbolAtLocation(typeAlias.name);
         if (sym.flags & ts.SymbolFlags.Value) return [];
-        // Type aliases are always emitted as the resolved underlying type, so there is no need to
-        // emit anything, except for exported types.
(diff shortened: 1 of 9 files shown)


evmar:
This passes in g3, ready for review.

src/jsdoc_transformer.ts
```diff
@@ -670,23 +667,38 @@ export function jsdocTransformer(
...
+          // Given: export type T = ...;
+          // Produce: exports.T;
+
+          // We cannot emit `export var foo;` and let TS generate from there because TypeScript
+          // drops exports that are never assigned values, and Closure requires us to not assign
+          // values to typedef exports. Introducing a new local variable and exporting it can cause
+          // bugs due to name shadowing and confusing TypeScript's logic on what symbols and types
+          // vs values are exported. Mangling the name to avoid the conflicts would be reasonably
+          // clean, but would require a two pass emit to first find all type alias names, mangle
+          // them, and emit the use sites only later.
```
rkirov: Feels like there is a missing final sentences - "so we do ..."

PR #1042, opened 2019-07-12
update to latest bazel

alexeagle:


tsconfig.json
```diff
@@ -21,9 +21,5 @@
     "downlevelIteration": true,
     "target": "es2015",
     "strict": true,
-  },
-  "include": [
-    "src/*.ts",
-    "test/*.ts"
```
evmar: Do we need to do something here so the editor doesn't get confused by the ts files in testdata?

package-lock.json
```diff
@@ -1,910 +0,0 @@
-{
```
evmar: Do we need this file for 'npm release', or can yarn do it?

alexeagle: Having two lock files almost certainly doesn't make sense.

I don't think `npm release` should have anything to do with installing packages so I imagine this file is totally unused.

I'm happy to  just update it instead to make the PR more conservative, though.

tsconfig.json
alexeagle: I found that non-sandboxed Bazel TS worker was trying to read these files from the cache for every build, following https://github.com/bazelbuild/rules_nodejs/pull/756/files
error looks like
```
ERROR: /usr/local/google/home/alexeagle/Projects/tsickle/src/BUILD:5:1: Compiling TypeScript (devmode) //src:src failed (Exit 1)
Compilation failed Error: missing input digest for /usr/local/google/home/alexeagle/.cache/bazel/_bazel_alexeagle/c5eeaf028cc4e9cd3f3c738037c2514c/execroot/tsickle/test/ast_printing_transform.ts.(only have [MEGABYTES])
```

couple options:
- turn on `--worker_sandboxing` to prevent discovering files outside the compilation unit
- use exclude rather than include in the tsconfig

alexeagle: Another option would be to continue disabling worker mode in the `.bazelrc` to match old behavior, but it makes the Dev RTT slow of course

alexeagle: I went with the first option and force-pushed a new commit PTAL

alexeagle: (note that Bazel 0.27 introduced a new action strategy where workers get enabled by default with no settings needed in .bazelrc)

PR #1044, opened 2019-07-15
fix(symbol): create closure type declaration for computed function names

### LINKED ISSUE #913: "anonymous abstract function" with symbol-identified methods
The following code:

```typescript
export interface AccessorBase<T> {
    readonly $accessPath: AccessPath<any, T>;
    readonly $parent: ObjectAccessor<{}> | ListAccessor<object> | undefined;

    [accessPathSymbol](): AccessPath<any, T>;
}
```

generates the error *anonymous abstract function*. This is most likely the case because [ `propertyName`](https://github.com/angular/tsickle/blob/9f47abe532473895db3c494ff5a3266696472b4b/src/jsdoc_transformer.ts#L313) only considers identifiers and string literals, and not symbols. I have not checked, but I would gess `['methodName'](): string` causes the same error.

The problem has been introduced in version 0.33.0 - 0.32.1 works.

ayazhafiz:
Functions with computed symbol names are not processed because their
names are not strings. Resolve this by handling for the specific case of
computed name in a function declaration and creating and element access
rather than a property access for the symbol naming the function.

Issue close #913.


ayazhafiz:
I'm not familiar with why the `if (false)` construct is created. I understand that it's intentional, but please let me know if I'm missing something.

src/jsdoc_transformer.ts
```diff
@@ -270,16 +270,22 @@ function createMemberTypeDeclaration(
...
+    // name can be resolved to a string.
+    const name = fnDecl.name && ts.isComputedPropertyName(fnDecl.name) ? fnDecl.name.expression :
+                                                                         propertyName(fnDecl);
     if (!name) {
       mtt.error(fnDecl, 'anonymous abstract function');
       continue;
     }
     const {tags, parameterNames} = mtt.getFunctionTypeJSDoc([fnDecl], []);
     if (hasExportingDecorator(fnDecl, mtt.typeChecker)) tags.push({tagName: 'export'});
+    // Element access for computed names.
```
evmar: This was hard to follow because of no verb, maybe something like "Use element access instead of property access for computed names" or so?

ayazhafiz: Fixed.

PR #1049, opened 2019-08-16
update to typescript 3.5

evmar:
The one change in type output looks like an improvement.


test_files/type/type.js
```diff
@@ -96,7 +94,7 @@ val => val + 1));
  */
 function typeGenericCallback(callback) { }
 typeGenericCallback((/**
- * @param {?} val
- * @return {?}
+ * @param {*} val
```
rkirov: It would be nice to be intentional about this. If think that `?` was better add a TODO. I think `*` is fine to force an explicit cast.

PR #1051, opened 2019-08-20
rewrite toplevel comma expressions as statements

evmar:
In various contexts TypeScript emits code like:
  exports.a = 1, exports.b = 2;
Closure wants all export statements to be separate toplevel statements:
  exports.a = 1; exports.b = 2;
(note the semi).

This new transformer pass rewrites all toplevel comma expressions as
statements.  See the test for the sorts of code we saw that is affected
by this.

### DIFF
--- a/src/googmodule.ts
+++ b/src/googmodule.ts
@@ -198,6 +198,33 @@ function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
 }
 
 /**
+ * Convert a series of comma-separated expressions
+ *   x = foo, y(), z.bar();
+ * with statements
+ *   x = foo; y(); z.bar();
+ * This is for handling in particular the case where
+ *   exports.x = ..., exports.y = ...;
+ * which Closure rejects.
+ *
+ * @return An array of statements if it converted, or null otherwise.
+ */
+function rewriteCommaExpressions(expr: ts.ExpressionStatement): ts.Statement[]|null {
+  // Early exit if the outer statement isn't a comma statement.
+  if (!ts.isBinaryExpression(expr.expression)) return null;
+  if (expr.expression.operatorToken.kind !== ts.SyntaxKind.CommaToken) return null;
+
+  // Recursively visit comma-separated subexpressions, and collect them all as
+  // separate expression statements.
+  return visit(expr.expression);
+  function visit(expr: ts.Expression): ts.Statement[] {
+    if (ts.isBinaryExpression(expr) && expr.operatorToken.kind === ts.SyntaxKind.CommaToken) {
+      return visit(expr.left).concat(visit(expr.right));
+    }
+    return [ts.setOriginalNode(ts.createExpressionStatement(expr), expr)];
+  }
+}
+
+/**
  * commonJsToGoogmoduleTransformer returns a transformer factory that converts TypeScript's CommonJS
  * module emit to Closure Compiler compatible goog.module and goog.require statements.
  */
@@ -441,6 +468,15 @@ export function commonJsToGoogmoduleTransformer(
             const modExports = rewriteModuleExportsAssignment(exprStmt);
             if (modExports) {
               stmts.push(modExports);
+              return;
+            }
+            // Check for use of the comma operator.
+            // This occurs in code like
+            //   exports.a = ..., exports.b = ...;
+            // which we want to change into multiple statements.
+            const commaExpanded = rewriteCommaExpressions(exprStmt);
+            if (commaExpanded) {
+              stmts.push(...commaExpanded);
               return;
             }
             // Check for:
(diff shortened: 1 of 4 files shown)


evmar:
Note that at first I attempted to look for `exports.a = ...` specifically, but it turns out that at the time this transformer sees the code it is just `a = ...` and the `exports.` bit is added in by the TS substitution stuff.  I think it's harmless to always rewrite.

src/googmodule.ts
```diff
@@ -197,6 +197,37 @@ function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
...
+ * @return An array of statements if it converted, or null otherwise.
+ */
+function rewriteCommaExpressions(expr: ts.ExpressionStatement): ts.Statement[]|null {
+  // Early exit if the outer statement isn't a comma statement.
+  if (!ts.isBinaryExpression(expr.expression)) return null;
+  if (expr.expression.operatorToken.kind !== ts.SyntaxKind.CommaToken) return null;
+
+  const stmts: ts.Statement[] = [];
+  visit(expr.expression);
+  function visit(expr: ts.Expression) {
```
mprobst: the usual pattern here is to have `visit()` return a `ts.Node|ts.Node[]`, and calls to `visit` use its return value recursively. It took me a bit to understand that you avoid this to flatten the comma expression by explicit pushes on stmts. Maybe you could add a comment explaining that and why you're deviating from the normal pattern?

src/googmodule.ts
```diff
@@ -197,6 +197,37 @@ function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
...
+ * This is for handling in particular the case where
+ *   exports.x = ..., exports.y = ...;
+ * which Closure rejects.
+ *
+ * @return An array of statements if it converted, or null otherwise.
+ */
+function rewriteCommaExpressions(expr: ts.ExpressionStatement): ts.Statement[]|null {
+  // Early exit if the outer statement isn't a comma statement.
+  if (!ts.isBinaryExpression(expr.expression)) return null;
+  if (expr.expression.operatorToken.kind !== ts.SyntaxKind.CommaToken) return null;
```
mprobst: if you wanted to be more precise, could you check for the `export` modifier (I think it's a modifier - or maybe a flag?) and/or check `ts.getOriginalNode`? I agree it should theoretically be harmless to convert like this, but we generally try not to muck with the user's original code if possible...

test_files/export_multi/export_multi.js
```diff
@@ -0,0 +1,27 @@
+goog.module('test_files.export_multi.export_multi');
+var module = module || { id: 'test_files/export_multi/export_multi.ts' };
+module = module;
+exports = {};
+var _a;
```
mprobst: it'd odd how for `_a` gets hoisted up here, but I presume that's just what TS does?

src/googmodule.ts
evmar: I'm not sure I followed you here, can you point me at one of the examples of the normal pattern?

src/googmodule.ts
evmar: Ah right, I had been thinking that this was working over the "es5 output" JS so the export flag is lost, but it's part of a transformation pipeline so it must still be there.  (In retrospect this is obvious because it manages to later jam the `exports.` in front.)  I think this is rare enough to not matter though.

test_files/export_multi/export_multi.js
evmar: I also found that pretty strange.  I will dig into this more.  It's possible this transformation nukes some important info.

(On the other hand, it passes global TAP...)

evmar: This is unrelated to this change.  TS somehow hoists the 'var' above the fileoverview in general, so I am not too concerned about it.

src/googmodule.ts
evmar: I started to write this, but it turns out that one of these forms ends up like:

```
_a = { a: 0, b: 0 }, exports.x = _a.a, exports.y = _a.b;
```

Where we still need to split to statements even though it has subexpressions that aren't export-related.

I could maybe try to special case it so we only split statements where one subexpression is an export, but I think I convinced myself again that this is a rare case.

src/googmodule.ts
mprobst: Here's an example:
https://github.com/angular/tsickle/blob/master/src/jsdoc_transformer.ts#L932

Note how `visitor` returns `ts.Node|ts.Node[]`, and (with more precise types) so do the functions it dispatches to. The functions then always return a node or list of nodes, either the input unchanged, or the modified nodes.

evmar: PTAL

PR #1053, opened 2019-08-21
Transform down-leveled await expressions.

mprobst:
TypeScript down-levels await expressions to a library call, passing a
generator function with the tail of the async function:
    tslib_1.__awaiter(this, void 0, void 0, function* () { ... });

The generator function containing the remainder of the function's code
needs an `@this` annotation for Closure to understand `this.` accesses
within it.

Additionally, because all async functions unconditionally pass
`tslib_1.__awaiter(this, ...)`, even functions without a `this` context
need to be annotated with `@this {*}` (i.e. an unknown this type) to
avoid a warning by Closure Compiler.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -417,8 +417,8 @@ export function removeTypeAssertions(): ts.TransformerFactory<ts.SourceFile> {
  */
 export function jsdocTransformer(
     host: AnnotatorHost, tsOptions: ts.CompilerOptions, typeChecker: ts.TypeChecker,
-    diagnostics: ts.Diagnostic[]): (context: ts.TransformationContext) =>
-    ts.Transformer<ts.SourceFile> {
+    diagnostics: ts.Diagnostic[], thisTypeByAsyncFunction: Map<ts.FunctionLikeDeclaration, string>):
+    (context: ts.TransformationContext) => ts.Transformer<ts.SourceFile> {
   return (context: ts.TransformationContext): ts.Transformer<ts.SourceFile> => {
     return (sourceFile: ts.SourceFile) => {
       const moduleTypeTranslator = new ModuleTypeTranslator(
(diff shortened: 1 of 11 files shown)


src/jsdoc_transformer.ts
```diff
@@ -542,6 +542,24 @@ export function jsdocTransformer(
...
+      function getContextThisType(node: ts.Node): ts.Type|undefined {
+        do {
+          if (ts.isClassDeclaration(node) && node.name) {
+            return typeChecker.getTypeAtLocation(node.name);
+          }
+          if (ts.isFunctionDeclaration(node) && node.parameters.length > 0) {
+            const firstParam = node.parameters[0];
+            if (ts.isIdentifier(firstParam.name) &&
+                transformerUtil.getIdentifierText(firstParam.name) === 'this' && firstParam.type) {
+              return typeChecker.getTypeAtLocation(firstParam.type);
```
evmar: This feels a bit sketchy to me, I wonder if it's better to ask for the function's type and get it from there?

src/jsdoc_transformer.ts
```diff
@@ -542,6 +542,24 @@ export function jsdocTransformer(
...
+            }
+          }
+          node = node.parent;
+        } while (node);
```
evmar: I'm surprised this compiles, because Node isn't declared as nullable.

...I bet node.parent isn't declared as nullable either. :(

src/jsdoc_transformer.ts
mprobst: The TS API does not expose this type declarations on signature, so we cannot do much better here. Left a comment.

src/jsdoc_transformer.ts
mprobst: Yes. Changed the declaration to `|undefined` and also converted to a while loop.

PR #1055, opened 2019-08-21
skip emit for properties named 'prototype'

evmar:
This would end up emitting Closure code like
  /** @type {string} */
  Foo.prototype.prototype;
which ends up confusing the compiler.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -339,6 +339,14 @@ function createClosurePropertyDeclaration(
         prop, `Skipping unnamed member:\n${escapeForComment(prop.getText())}`);
   }
 
+  if (name === 'prototype') {
+    // Code that declares a property named 'prototype' typically is doing something
+    // funny with the TS type system, and isn't actually interested in naming a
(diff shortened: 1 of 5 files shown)


mprobst:
@evmar have we inquired whether this is a fundamental limitation, or just a bug, or what?

evmar:
I started a thread about it.  It's not clear exactly what's going wrong.  I think in terms of runtime behavior, no code actually wants to access this 'prototype' member, it's just doing it as a way of expressing a supertype constraint to TS, so I think it is harmless to swallow in this way.

src/jsdoc_transformer.ts
```diff
@@ -339,6 +339,12 @@ function createClosurePropertyDeclaration(
         prop, `Skipping unnamed member:\n${escapeForComment(prop.getText())}`);
   }
 
+  if (name === 'prototype') {
+    // It's illegal to name a property 'prototype'.
```
mprobst: Maybe expand this comment with what you wrote in the comment thread here?

evmar: Done.

PR #1056, opened 2019-08-22
Run npm test on CircleCI.

mprobst:
npm test includes clang-format checks and linting, in addition to bazel
test.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,5 +32,5 @@ jobs:
           key: *cache_key
       - run: bazel run @nodejs//:yarn
       - run: bazel build ...
-      - run: npm test
+      - run: bazel-tsickle/external/nodejs_linux_amd64/bin/npm test
       - *save_cache
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,5 +32,5 @@ jobs:
           key: *cache_key
       - run: bazel run @nodejs//:yarn
       - run: bazel build ...
-      - run: bazel test ...
+      - run: npm test
       - *save_cache
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -673,7 +673,7 @@ export class TypeTranslator {
       const paramsStr = params.length ? (', ' + params.join(', ')) : '';
       const constructedType = this.translate(ctors[0].getReturnType());
       const constructedTypeStr =
-          constructedType[0] == '!' ? constructedType.substring(1) : constructedType;
+          constructedType[0] === '!' ? constructedType.substring(1) : constructedType;
       // In the specific case of the "new" in a function, the correct Closure
       // type is:
       //


mprobst:
@alexeagle @gregmagolan any idea how to reach a `node` or `npm` binary for my docker command?

The problem is the docker image does not contain "npm" and "node" binaries, as far as I can tell. Now we do have node installed via bazel, and ideally we'd probably want to use the same node and npm binaries as are installed in the bazel environment.

I don't know how though. The binaries live in bazel-tsickle/external/nodejs_linux_amd64/bin/npm, which I'd much rather not hard code. "bazel run @nodejs//:bin/npm -- test $PWD" fails with all kinds of problems due to the working directory not being the project root.

I also cannot make these tests bazel tests, because they (by design) need all sources of the project; it's the linter and formatter run.

alexeagle:
One answer is to switch the docker image used to boot up the test machine from a Bazel one to https://hub.docker.com/r/circleci/node/
Then you'd just run the CI as if it's a JavaScript project, letting the Bazel dependency be hidden in the `@bazel/bazel` package, and the package.json script with `bazel test blah` uses Bazel behind the scenes
I think this is probably better since tsickle is an OSS npm project, and to the extent it has external contributors, it's easier for them to understand if it's set up like other npm-based projects.

I imagine we should also be able to fix the method you tried with `@nodejs//:bin/npm` to grab the npm used by bazel, if you want to stay bazel-first

mprobst:
Re using the node docker image, wouldn't that make my build much less efficient, when I always have to first download bazel (~400 MB or so?).

I'm not sure it'd make the project much more approachable to use bazel from npm. We're still running a build system and overall process that's far off the beaten path, but I'm not too worried about that anyway.

Do you have an idea how we could go about locating that binary, or running it? I really just need to set the working directory to the current project root.

alexeagle:
bazel will be cached the same as your other node_modules dependencies. Also it's only 40MB now (though there are still a pair of bugs in npm & yarn where they download tarballs despite already knowing they don't match the local OS and will never be installed)

I tried switching to the circleci/node docker image but you also have a test that depends on finding `java` on the machine:
https://circleci.com/gh/angular/tsickle/922

I'll try another approach before giving more suggestions...

alexeagle:
Here's an approach that is green on CI: use a nodejs docker container for running a `lint` job, which runs in parallel with the `bazel` job running in the current container.
It has the benefit that they run in parallel so your total CI latency is lower.
However it does use the circleci workflow feature which makes the UX a bit more complex
https://github.com/angular/tsickle/pull/1058/files

alexeagle:
Still debugging why your approach doesn't work.
I added `set -x` inside the script we write to `@nodejs//:bin/npm`
Even running the final command in the tsickle directory
`/usr/local/google/home/alexeagle/.cache/bazel/_bazel_alexeagle/c5eeaf028cc4e9cd3f3c738037c2514c/external/nodejs_linux_amd64/bin/node /usr/local/google/home/alexeagle/.cache/bazel/_bazel_alexeagle/c5eeaf028cc4e9cd3f3c738037c2514c/external/nodejs_linux_amd64/bin/nodejs/lib/node_modules/npm/bin/npm-cli.js test`
fails the same way.
(Note that `bazel run` does now run subprocesses in your pwd)

alexeagle:
`bazel run @nodejs//:bin/yarn test` does work however.

alexeagle:
https://github.com/angular/tsickle/pull/1059 is closer to what you want: keep a single circleci job and a single docker image.
I simplified the circleci config since most of the variables were referenced a single time in the single job and could therefore be inlined.

mprobst:
This has been superseded by Alex' fix (thanks!)

PR #1059, opened 2019-08-22
circleci: run the package.json test script

alexeagle:


.circleci/config.yml
```diff
@@ -1,36 +1,26 @@
...
+      - image: l.gcr.io/google/bazel:0.28.0
     steps:
       - checkout
       - restore_cache:
           key: *cache_key
       - run: bazel run @nodejs//:yarn
-      - run: bazel build ...
-      - run: bazel test ...
-      - *save_cache
+      - run: bazel run @nodejs//:bin/yarn lint
```
mprobst: Just so that I understand - running yarn somehow sets the correct working directory?

.circleci/config.yml
```diff
@@ -1,36 +1,26 @@
 # Cache key for CircleCI. We want to invalidate the cache whenever the Bazel workspace or the
 # NPM dependencies changed.
 var_1: &cache_key angular-tsickle-{{ checksum "WORKSPACE" }}-{{ checksum "yarn.lock" }}
-# Default docker image for CircleCI jobs with Bazel installed. The version of Bazel is controlled
```
mprobst: Are these changes just cleanup, or in some way load bearing?

alexeagle: cleanup, see the three commits here, the last one is just the refactoring

.circleci/config.yml
alexeagle: the working directory is correct either way, but something in the npm-cli.js script caused the problem you saw, I still didn't figure out what - see the comments on your PR

mprobst: Good to know.

PR #1063, opened 2019-08-30
Improve externs handling

mprobst:
Separately reviewable commits. See commit message of each for details.


test_files/import_equals/exporter.d.ts
```diff
@@ -0,0 +1,3 @@
+export declare namespace exportedSymbol {
```
rkirov: I found the name `exportedSymbol` confusing when reading it in a module that imports it. I think modules need to have some sort of names to differentiate them - `Top,Middle,Bottom` or `Consumer, Producer` or my favorite - `A,B,C`. Exporter is just too confusing since every module exports.

src/externs.ts
```diff
@@ -181,6 +181,35 @@ export function generateExterns(
...
+    const entityName = getEntityNameText(name);
+    let symbol = typeChecker.getSymbolAtLocation(name);
+    if (symbol) {
+      // If this is an aliased name (e.g. from an import), use the alias to refer to it.
+      if (symbol.flags & ts.SymbolFlags.Alias) {
+        symbol = typeChecker.getAliasedSymbol(symbol);
+      }
+      const alias = mtt.symbolsToAliasedNames.get(symbol);
+      if (alias) return alias;
+      const isGlobalSymbol = symbol && symbol.declarations && symbol.declarations.some(d => {
```
rkirov: isGlobalSymbol looks like an useful general utility fn to extract. I am surprised it doesn't already exist in tsickle.

src/externs.ts
```diff
@@ -543,21 +564,32 @@ export function generateExterns(
...
     }
+    if (symbol.flags & ts.SymbolFlags.Alias) {
+      symbol = typeChecker.getAliasedSymbol(symbol);
+    }
+    // While type_translator does add the mangled prefix for ambient declarations, it only does so
+    // for non-aliased (i.e. not imported) symbols. That's correct for its use in regular modules,
+    // which will have a local symbol for the imported ambient symbol. However within an externs
+    // file, there are no imports, so we need to make sure the alias already contains the correct
+    // module name, which means the mangled module name in case of imports symbols.
+    // This only applies to non-Closure ('goog:') imports.
```
rkirov: does that fix my clutz import/export issue? I don't see any test with 'goog:' changing.

rkirov: Also, I don't undestand what id non-Closure (goog:) import, isn't 'goog:' only closure imports?

src/externs.ts
```diff
@@ -681,17 +713,13 @@ export function generateExterns(
...
-          break;
-        }
         if (importEquals.moduleReference.kind === ts.SyntaxKind.ExternalModuleReference) {
           addImportAliases(importEquals);
           break;
         }
-        const qn = getEntityNameText(importEquals.moduleReference);
+        const qn = qualifiedNameToMangledIdentifier(importEquals.moduleReference);
         // @const so that Closure Compiler understands this is an alias.
-        if (namespace.length === 0) emit('/** @const */\n');
```
rkirov: why drop the if check?

src/externs.ts
```diff
@@ -174,6 +174,18 @@ export function generateExterns(
     rootNamespace = moduleNameAsIdentifier(host, sourceFile.fileName);
   }
 
+  // moduleNamespace holds the namespace for this ES module, whereas rootNamespace is where we
+  // generate symbols (see below).
+  const moduleNamespace = rootNamespace;
+  // There can only be one export =, and if there is one, there cannot be any other exports.
```
rkirov: I understand your comments yet still don't see why do we need the '_', i.e. I am missing the big picture.

There must be some sort of constraint that passes in TS and is problematic in closure. Maybe we need a quick VC on Monday morning?

test_files/import_equals/exporter.d.ts
mprobst: Interesting. I find that `A,B,C` is really hard for me to understand, same for top middle bottom. Those depend on what direction you think your imports point (or in the case of A, B, C, might just be 100% random).

Most tsickle tests use `exporter` for the file that exports a symbol that's needed by the test, with the other file being the actual test. That is, `exporter` is always test setup, and is always the module whose only relevance is exporting a symbol.

We could do a PR to move to consumer / producer or similar, but I'm not 100% sold that that'd make things easier to understand.

src/externs.ts
mprobst: It does feel like that, and I've checked the code base again just now (and did before). We have some code that deals with similar situations in `type_translator.ts`, but its semantics are subtly different. The major difference is that we're emitting into `externs` here, which are necessarily a global namespace, as opposed to `goog.module`.

src/externs.ts
mprobst: A Closure import is an import with `'goog:'` prefix. A non-Closure import is any other import. This code only applies to the non-Closure bit.

I think this is independent of your import/export issue (though I might be misunderstanding that). There's a test for this, which is (paraphrapsing) `import X from 'goog:y.X'; export declare const x: X;` in a `.d.ts` file. The important bit is that we emit the `y.X` type as `y.X` even in the externs file.

src/externs.ts
mprobst: I don't actually know why we had the check there, I think it was inherited over many generations of this file. It's simply not right, we should mark all generated namespaces as `@const` so Closure can infer types better, and they can be used in `typeof` etc.

src/externs.ts
mprobst: SGTM. Monday is labor day though, let's aim for Tuesday :-)

mprobst: I've still reworded the comment here, hopefully it's a bit more clear.

src/externs.ts
```diff
@@ -160,18 +160,42 @@ export function generateExterns(
...
+  }
+
+  // Symbols are generated starting in rootNamespace. For script .d.ts with global symbols, this is
+  // the empty string. For most module `.d.ts` files, this is the mangled namespace object. The
+  // remaining special case are `.d.ts` files containing an `export = something;` statement. In
+  // these, the effective exports object, i.e. the object containing the symbols that importing code
+  // receives, is different from the main module scope.
+  // tsickle handles the `export =` case by generating symbols in a different namespace (escaped
+  // with a `_`) below, and then assigning whatever is actually exported into the `moduleNamespace`
+  // below.
```
rkirov: Testing my understanding of this code, I think that  `exports {foo: bar}` will also need some special handling. Add a TODO if I am right (not worth fixing now).

src/externs.ts
rkirov: I understand this now after our VC, thanks! The missing piece is that we need to separate the module-locals and module-exports. With `exports = ` those it's not clear how those two are even related.

src/externs.ts
rkirov: ack.

test_files/import_equals/exporter.d.ts
rkirov: I see, as long as we have a good rule-of-thumb as you describe.

PR #1064, opened 2019-09-04
Add explcit @this {*} anotation to static async methods.

rkirov:
Because the async downleveler emits 'this' for all async methods we have
to explicitly annotate the 'this' usage so that closure does not
complain.

Technically for static methods the right anotation would be `@this
{typeof C}` but for now the already existing `*` should suffice.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -595,12 +595,24 @@ export function jsdocTransformer(
           }
         }
 
-        // top-level async functions when down-leveled access `this` to pass it to
-        // tslib.__awaiter. Closure requires a @this tag for that.
-        if ((tsOptions.target !== undefined && tsOptions.target <= ts.ScriptTarget.ES2015) &&
-            transformerUtil.hasModifierFlag(fnDecl, ts.ModifierFlags.Async) &&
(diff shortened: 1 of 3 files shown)


src/jsdoc_transformer.ts
```diff
@@ -595,12 +595,20 @@ export function jsdocTransformer(
           }
         }
 
+        const needsExplicitThis =
+            // Methods/getters/setters/ctors already have an implicit this.
+            fnDecl.kind === ts.SyntaxKind.FunctionDeclaration ||
+            // static methods also need explicit this
+            // TODO(rado): The correct emit for static methods would be
+            // {typeof C} where C is the enclosing class.
+            (fnDecl.kind === ts.SyntaxKind.MethodDeclaration &&
```
mprobst: You can actually have static getters:
https://typescript.googleplex.com/?codez=S85JLC5WcFaoViguSSzJTFZITy1RSMvP19AEChWllpQW5SkYWivUKtRycTnrASWsAQ%3D%3D

But I believe you cannot have async getters. Maybe this code would be more readable (and that's my fault!) if we started with the "is async" condition?

rkirov: Yep, I moved the comments around so it is more clear.

PR #1069, opened 2019-09-10
always obey decorator downlevelling flag

evmar:
If the caller asked us to transform Closure types, we accidentally
would *always* add the decorator downleveller, regardless of the
setting of the decorator downleveller flag.

Once nvIvy is done, we will be in a config where we want Closure
types transformed and decorator downlevelling off, which was
inexpressible due to the above bug.  The fix here is trivial.


evmar:
CC @alxhub

Issue #1072, opened 2019-09-16
Skipping an @implements can cause misoptimization with typed optimizations


evmar:
In some cases we back off from emitting an "@implements" tag.  But with typed optimizations, every "@implements" is potentially load-bearing.

At the TS level, every "implements" is always optional.  At the Closure level, it is dangerous to ever leave out an "@implements".  We have no good user-facing way of reconciling these two.

I think we should instead say something like: if you *ever* write "implements" in TS, we *always* emit an "@implements", and if the latter fails for some reason you should delete the "implements" from the TS side.

evmar:
After some work in this area, I now better understand all the cases here.

1. if the symbol comes from Clutz?  then we should emit (<- this change)
2. if the symbol is a TS builtin? then it probably exists in Clousure and we can emit it
3. if the symbol comes from a user-written d.ts, then we can probably emit it with name mangling
4. if the symbol comes from a tsickle-transpiled file, then we can probably emit it with name mangling, possibly depending on whether namespaces are involved or not

Note that fixing #2 ends up triggering conformance because there's an Angular type that extends Window and by using that indirection people were calling unsafe window functions.

PR #1073, opened 2019-09-16
emit @implements on 'implements' of type+value

evmar:
If tsickle gets input that contains a type-value conflict, e.g.
  interface X { ... }
  const X = 3;
then it drops the 'interface' declaration in the output because
at the Closure level, the definition of "X" clobbers the @record.

Because of this, if someone later attempts to *use* the interface, e.g.
  class C implements X {}
tsickle would silently drop the '@implements' of that interface in
the Closure output.

But dropping @implements is dangerous in the presence of typed
optimzations; and this actually caused a real misoptimization in a
Google app.

The reason was that you actually can have legitimate type+value
interfaces in Closure:
  /** @interface */ class I { ... }
  I.someValue = 3;
and Clutz then dutifully converts this to the matching TypeScript,
and then if someone implements that interface, dropping the @implements
can cause a misoptimization.

In #1072 I convinced myself that we should instead never drop any
'implements' clause.  If the user writes the above pattern with X
and C, we should just compile fail instead, because there is no
way we can safely emit it.

In the test case here I added some usage of 'export import', but it
turns out that was irrelevant to this bug; it's just that when Clutz
started using 'export import' it got better at correctly describing
the above 'class I' pattern, which cascaded into this tsickle failure.
The clutz 'export import' change is
  https://github.com/angular/clutz/commit/cc9c9e6c8ca704aa05b5a3ec6e15ddf73d0c9a23


test_files/export_import/test.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
mprobst: nit: I usually name these `export_import/export_import.ts` so that we don't end up with a million files called `test.ts`. The repetition sucks too, not sure if it's substantially better...

test_files/export_import/test.ts
```diff
@@ -0,0 +1,17 @@
...
+// Simple implementation of a type+value.
+// We expect an @implements to show in the output.
+class C1 implements export_import.IFace {
+  a = 'a';
+}
+
+// Simple implementation of a type+value that is indirected through an
+// 'export import' clause.
+// We expect an @implements to show in the output.
+class C2 implements export_import.Reexport.Reexported {
```
mprobst: are you going to remove the export import stuff here before submission? It's distracting from what the test really tests, isn't it?

src/jsdoc_transformer.ts
```diff
@@ -90,6 +90,10 @@ export function maybeAddHeritageClauses(
           type: heritage.parentName,
         });
       }
+      // In the heritage==null case, we silently have dropped an @implements.
+      // In issue 1072 I advocate that we should never drop an @implements here,
```
mprobst: I don't understand this comment. It sounds as if you advocate something, but then didn't act on it? Is that a follow up?

The way I read the code it suggests that we below fall through and do emit?

test_files/export_import/iface.d.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * @fileoverview This file vaguely resembles the shape of the clutz output of
```
mprobst: Can you narrow this down to a test that just creates the type/value conflict?

test_files/class/class.ts
```diff
@@ -117,22 +117,6 @@ let abstractClassVar: AbstractClass;
 abstractClassVar = new ClassImplementsAbstractClass();
 abstractClassVar = new ClassExtendsAbstractClass();
 
-// Reproduce issue #333: type/value namespace collision.
-// Because Zone is both a type and a value, the interface will be dropped
-// when converting to Closure, so the "implements" should be ignored for
-// both the direct use and the use via a typedef.
-interface Zone {
```
mprobst: You're removing coverage here. Can we keep a test that shows what happens when we have a type/value conflict in a regular `.ts` file?

mprobst:
I think we should keep a test and show what happens for a type/value conflict in `.ts` code.

evmar:
Abandoning this in favor of https://github.com/angular/tsickle/commit/ef80516869e915fa2fd7c74d0ab7f08ad16d9501

PR #1077, opened 2019-09-17
Mangle type names when they conflict with values

evmar:
This series of commits changes our emit for code like

```ts
interface X { ... }
const X = 3;
```

Previously, we'd just quietly drop the `X` because of the conflict.
With this change, we instead emit the type with a mangled name:

```js
/** @record */
function X$$TSType() { ... }
const X = 3;
```

Then when some code wants to *use* that type, it needs to know whether that type has a mangled JS name.  To do so I add a header to any tsickle-emitted d.ts.  This actually let me remove a hack related to standard library types too.


evmar:
CC @mprobst FYI, not ready for review yet.

Basic idea is: we know only tsickle will do the type-value-conflict rename, so we make tsickle-emitted d.ts carry a tag and use that to judge whether a type got rewritten.

evmar:
I'm gonna sit on this.  I think we will need to do something like it eventually, but I fixed the larger clutz issue separately.

PR #1079, opened 2019-09-20
improve user visibility of heritage backoffs

evmar:
When tsickle decides to omit a user-provided 'extends' or 'implements',
put a comment in the source saying that it did so and why it did.

Also improve the text of these messages to be more readable for users.

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -81,32 +81,26 @@ export function maybeAddHeritageClauses(
 
     // Otherwise, if we get here, we need to emit some jsdoc.
     for (const expr of heritage.types) {
-      const heritage = heritageName(isExtends, hasExtends, expr);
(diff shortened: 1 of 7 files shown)


src/jsdoc_transformer.ts
```diff
@@ -162,28 +157,39 @@ export function maybeAddHeritageClauses(
           tagName = 'extends';
         } else {
           // Closure can only @implements an interface, not a class.
-          mtt.debugWarn(decl, `omitting @implements of a class: ${expr.getText()}`);
-          return null;
+          warn(decl, `dropped extends of class: ${expr.getText()}`);
```
mprobst: this is an `implements` here from the perspective of the TS user, right?

evmar: Good catch, fixed.

evmar: Good catch, fixed.

PR #1080, opened 2019-09-20
add more tests around type/value conflicts

evmar:
Add:
- augmented browser global
- externs
- an enum (which is also a type/value).


test_files/type_and_value/type_and_value.js
```diff
@@ -31,3 +42,12 @@ let useAsType;
 // Use a templatized user-defined interface/value pair as a type.
 /** @type {?} */
 let useAsTypeTemplatized;
+// Use the extern-defined types, found in typing.d.ts.
+/** @type {?} */
+let extUseAsType;
+/** @type {number} */
+let extUseAsValue = ExtTypeAndValue;
+/** @type {?} */
+let extUseEnumAsValue = ExtEnum;
```
mprobst: I would have thought that we correctly declare the type here, but I guess I'm wrong?

evmar: This is using the enum objects themselves as values, so I think the type is like `typeof Enum`, which we don't generate.  (Also this may trigger collapsing warnings now that I think of it.)  I included it here for completeness because I realized that enums are also special in the world of type/value conflicts.

PR #1081, opened 2019-09-20
always emit clutz types in heritage clauses

evmar:
When there is a type-value conflict, e.g.
  interface B {}
  const B = 3;
then tsickle doesn't emit a Closure type for the interface.

Then when
  class A implements B
tsickle omits the @implements for B, because B doesn't exist.

But when B comes from a Clutz definition, we know that the type
B does exist on the Closure side, and we must emit the @implements.

This change keys off a Clutz-emitted file header to control this.

Partial fix of #1072.


evmar:
Depends on https://github.com/angular/clutz/pull/926

src/type_translator.ts
```diff
@@ -50,6 +50,23 @@ function isClosureProvidedType(symbol: ts.Symbol): boolean {
       symbol.declarations.some(n => isBuiltinLibDTS(n.getSourceFile().fileName));
 }
 
+/**
+ * Returns true if the source file is generated by Clutz, i.e. has the magic
+ * Clutz header.
+ */
+function isClutzDts(sourceFile: ts.SourceFile): boolean {
+  console.error('isclutz', sourceFile.fileName);
```
mprobst: xxx debug code

evmar:
Post-review note: I had to add the underlying goog.module that we're simulating Clutz for here for the test that verifies Closure compilation to pass.  It's just a straightforward JS file, so I think it's harmless to submit.

PR #1082, opened 2019-09-24
move main driver program to a demo/ subdir

### LINKED ISSUE #1075: Remove command-line program, change to being library-only
We have a bunch of bugs here that are various forms of "the command-line tsickle program doesn't do what I want".  In practice, using tsickle well requires a driver program like the Angular compiler, or maybe https://github.com/theseanl/tscc .

I think we should consider removing our command-line interface entirely and instead say to use one of these other tools.  Right now our CLI tempts people into trying it, but it doesn't really work, and we don't have resources to improve it.

evmar:
This makes it more clear that the main entry point is just a demo.

This also lets us drop a bunch of cruft from the main tsickle
package.json.

Resolves #1075, more or less.


evmar:
The one part I haven't resolved here is what to do with the e2e main "test", which tries to test the demo module.  In this patch I moved it to the demo/ subdir, but it doesn't actually compile there because it wants to use jasmine and the main tsickle test_support module.

My current thought is to just delete it, because we won't run it as part of the CI anyway.

README.md
```diff
@@ -33,21 +33,21 @@ to try it you should expect to spend some time debugging and reporting bugs.
...
+Some known clients are:
 
-If you have complicated tsc command lines and flags in a build file (like a
-gulpfile etc.) Tsickle won't know about it. Another reason it's nice to put
-everything in `tsconfig.json` is so your editor inherits all these settings as
-well.
+1. Within Google we use tsickle inside the [Bazel build
+   system](https://bazel.build/). That code is published as
+   open source as part of [Bazel's nodejs/TypeScript
+   build rules](https://bazelbuild.github.io/rules_nodejs/.
```
mprobst: I think you're missing a `)` here

demo/tsconfig.json
```diff
@@ -0,0 +1,66 @@
...
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es2017",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */
+    "module": "commonjs",                     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', or 'ESNext'. */
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true,                       /* Allow javascript files to be compiled. */
+    // "checkJs": true,                       /* Report errors in .js files. */
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
+    // "declarationMap": true,                /* Generates a sourcemap for each corresponding '.d.ts' file. */
```
mprobst: consider trimming this a bit?

demo/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "Apache-2.0",
+  "private": true,
+  "scripts": {
+    "build": "node_modules/typescript/bin/tsc",
+    "demo": "node build/demo.js -- --outDir `pwd`/tsickle-out"
+  },
+  "dependencies": {
+    "minimist": "^1.2.0",
+    "mkdirp": "^0.5.1",
+    "tsickle": "../bazel-bin/npm_package",
```
mprobst: xxx should this be actual `tsickle`? you can still npm link it for local dev, I think.

mprobst:
> My current thought is to just delete it, because we won't run it as part of the CI anyway.

We could also leave it around as something we can run manually, or have a small shell script that drives the demo. I'm fine either way.

README.md
evmar: Done.

demo/tsconfig.json
evmar: Done.

demo/package.json
evmar: Done.

PR #1083, opened 2019-09-25
fix(perf): cache symbol strings within modules.

mprobst:
tsickle has bad performance for certain specific patterns, union types
with many members that are declared in namespaces with many members.

TypeScript compiler expands unions eagerly, including literal type
members. E.g. `undefined|SomeEnum` gets represented as
`undefined|SomeEnum.A|SomeEnum.B|...`. We then generate the type for
each member and deduplicate. For large enums (in the wild we see >2000
members), this means we generate many symbolToString calls.

That wouldn't be so bad, but it turns out `symbolToString`, or rather
the TS API `symbolToEntityName` that it calls, can be very slow.

Consider this example:

    namespace foo {
      enum OurEnum {A, B, C, ...}
    }
    namespace foo {
      // more declarations
    }
    namespace foo {
      // even more declarations
    }
    ...

`symbolToEntityName` is called for each enum member, then gets the
parent symbol chain for the enum member, i.e. `[foo, OurEnum]` here, and
then for each declaration of `foo` inspects whether that particular
declaration is better to name the symbol (e.g. would have a more
accessible file name/path for the user, be shorter in the current scope,
etc.).

This looks esoteric, but it turns out this is exactly the kind of code
that Clutz generates for top level namespaces with many declarations in
separate files contained in them. We counted >100'000 declarations for a
specific root namespace, with enums with many members, causing multi
minute tsickle runtimes on very small sources.

The cache is a quick fix to avoid the duplicated work. It might be
possible to (a) avoid generating these in Clutz and (b) improve
performance of `symbolToEntityName`.

### DIFF
--- a/src/module_type_translator.ts
+++ b/src/module_type_translator.ts
@@ -98,6 +98,12 @@ export class ModuleTypeTranslator {
   symbolsToAliasedNames = new Map<ts.Symbol, string>();
 
   /**
+   * A cache for expensive symbol lookups, see TypeTranslator.symbolToString. Maps symbols to their
+   * Closure name in the current file scope.
+   */
+  private symbolToNameCache = new Map<ts.Symbol, string>();
+
+  /**
    * The set of module symbols requireTyped in the local namespace.  This tracks which imported
    * modules we've already added to additionalImports below.
    */
@@ -151,8 +157,9 @@ export class ModuleTypeTranslator {
     const translationContext = this.isForExterns ? this.sourceFile : context;
 
     const translator = new typeTranslator.TypeTranslator(
-        this.host, this.typeChecker, translationContext, this.host.typeBlackListPaths,
-        this.symbolsToAliasedNames, (sym: ts.Symbol) => this.ensureSymbolDeclared(sym));
+        this.host, this.typeChecker, translationContext, this.host.typeBlackListPaths || new Set(),
+        this.symbolsToAliasedNames, this.symbolToNameCache,
+        (sym: ts.Symbol) => this.ensureSymbolDeclared(sym));
     translator.isForExterns = this.isForExterns;
     translator.warn = msg => this.debugWarn(context, msg);
     return translator;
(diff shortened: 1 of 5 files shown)


src/module_type_translator.ts
```diff
@@ -97,6 +97,9 @@ export class ModuleTypeTranslator {
    */
   symbolsToAliasedNames = new Map<ts.Symbol, string>();
 
+  /** A cache for expensive symbol lookups, see TypeTranslator.symbolToString. */
```
evmar: Can you make this comment say something about what the variable is?  E.g. the values in the map are the closure translated names for the symbols.

src/type_translator.ts
```diff
@@ -497,12 +507,13 @@ export class TypeTranslator {
...
-    let parts = type.types.map(t => this.translate(t));
     // Union types that include literals (e.g. boolean, enum) can end up repeating the same Closure
     // type. For example: true | boolean will be translated to boolean | boolean.
     // Remove duplicates to produce types that read better.
-    parts = parts.filter((el, idx) => parts.indexOf(el) === idx);
-    return parts.length === 1 ? parts[0] : `(${parts.join('|')})`;
+    const parts = new Set(type.types.map(t => this.translate(t)));
+    // If it's a single element set, return the single member.
+    if (parts.size === 1) return parts.values().next().value;
+    return `(${Array.from(parts.values()).join('|')})`;
```
evmar: Does this preserve order?  I guess Sets must behave like other JS structures, or all the tests would become unstable.

mprobst: Yes, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#Description

src/module_type_translator.ts
mprobst: Done.

mprobst:
Mh. There's an interaction with blacklisting function type template arguments (`x: <T>(): T;`) that I don't understand yet. The symbol for `T` here should consistently always be `?` in the entire file, but whether we know it's `?` depends on when we visit it, and apparently my change tickles things such that ordering has an impact on the cache.

I'll debug more tomorrow.

PR #1086, opened 2019-09-27
try harder to provoke compiler warnings

evmar:
The JSC_UNKNOWN_THIS failure mode only happens under very specific
circumstances, as described in the comments here.  Without these
settings, some of the code in test_files/async/ passes the test
suite even if they don't include the fixes they're trying to verify.


mprobst:
/CC @concavelenz might be a good reviewer for this, too.

evmar:
Context for @concavelenz, this is the tsickle test suite, which here is invoking JSCompiler on the tsickle output for all its test files to verify there are no warnings/errors.  I discovered our test for handling of "unknown this" wouldn't fail even when I added more test cases, and here is my fix.

PR #1088, opened 2019-09-30
add more async test cases, simplify code

evmar:
These additional cases came from some problematic google3 code,
as well as from Rado's reverted patch.

Also, we don't need any Promises to exercise the this-checking logic,
so remove them to make the tests easier to read.

### DIFF
--- a/test_files/async_functions/async_functions.js
+++ b/test_files/async_functions/async_functions.js
@@ -8,26 +8,23 @@ module = module;
 exports = {};
 const tslib_1 = goog.require('tslib');
 /**
+ * Exercises various forms of async functions.  When TypeScript downlevels these functions, it
+ * inserts a reference to 'this' which then tickles a Closure check around whether 'this' has a
+ * known type.
+ */
+/**
  * @param {string} param
  * @return {!Promise<string>}
  * @this {*}
  */
 function asyncTopLevelFunction(param) {
     return tslib_1.__awaiter(this, void 0, void 0, function* () {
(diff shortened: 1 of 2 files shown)


test_files/async_functions/async_functions.js
```diff
@@ -1,71 +1,50 @@
-/**
- * @fileoverview added by tsickle
- * @suppress {checkTypes,constantProperty,extraRequire,missingOverride,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
- */
 goog.module('test_files.async_functions.async_functions');
 var module = module || { id: 'test_files/async_functions/async_functions.ts' };
 module = module;
 exports = {};
 const tslib_1 = goog.require('tslib');
+/**
```
mprobst: hu, why does this end up below the `goog.module` etc? That seems like a bug.

Can you figure out why this happens? What's different here?

evmar: It appears to be something going wrong when it's a @fileoverview comment.

Since it's not relevant to this PR, I undid that part of it.

mprobst: I don't see the change yet?

evmar: The @fileoverview is still at the top of hte file, unmodified.  This is just a regular comment.

evmar: Ping, don't forget me!

evmar: Wait, you approved and I just missed it, ha!

PR #1089, opened 2019-09-30
Async functions revisit

evmar:
See commits.

Not ready for submit yet, just gonna try this against google3.


evmar:
Current status: this might be getting fixed upstream in Closure compiler, waiting to see if that works out.

mprobst:
Is this ready for review? Can you rebase?

evmar:
Now that I figured https://github.com/angular/tsickle/commit/d4bfe83e378ba45e474b6760f87ef70719b4237f , we can now repro the g3 problems in the test suite.  But this change still doesn't pass all those cases, I'm still tinkering with it.  I rebased, but this currently fails on L115 of async_functions.js (top-level async function expression, ew)

evmar:
This now fixes one case in async_functions.ts, and passes in g3.

test_files/async_functions/async_functions.ts
```diff
@@ -52,11 +52,10 @@ class Container {
 }
 
 // TODO(#1099): uncomment this test after the async emit is fixed.
-// const asyncFnExpression = async function f() {};
+// export const asyncFnExpression = async function f() {};
 
-const asyncArrowFn = async () => {};
+export const asyncArrowFn = async () => {};
```
mprobst: I don't understand the test change here. Why does this need to be exported, and why do we drop the `this` type from the function below? IIRC that was intentional to test the behaviour when a `this` type is set (`hasExistingTag`).

mprobst: ... but maybe should not use `this: unknown` but something like `this: SomeType`.

src/jsdoc_transformer.ts
```diff
@@ -431,14 +431,24 @@ export function removeTypeAssertions(): ts.TransformerFactory<ts.SourceFile> {
   };
 }
 
+/**
+ * @return true if node lexically contains an 'async' function.
```
mprobst: Google style is to prefer text description over the return tag. Also, call out that it's the entire subtree?

`containsAsync returns true if node or any of its descendants lexically contain ...`

evmar: I think repeating the function name is Go style, right?  In any case, fixed.

test_files/async_functions/async_functions.ts
evmar: I added 'export' everywhere because that triggers yet more JSCompiler subtlety, but I see it is confusing, so I fixed it by duplicating these.

It was not intentional that the lower function had a 'this' type -- that's why there was the TODO.  We had been adding "this: unknown" in g3 to make functions like this one compile, and this test was verifying that fix.  But this patch now fixes top-level async functions, so I can now remove the 'this' bit.

I added an additional function with a 'this' tag for that test coverage.

src/jsdoc_transformer.ts
mprobst: The function name indeed is Go style, I guess in JS you'd do "Returns ..." (but I think the Go style is kosher as well, and nicer).

PR #1091, opened 2019-10-01
Upgrade @bazel dependencies.

mprobst:
Just for hygiene.


mprobst:
I filed https://github.com/bazelbuild/rules_nodejs/issues/1220 for the non-hermeticity I encountered. I think it doesn't explain the failure I'm seeing here though, as I'd guess on CircleCI things should be hermetic, at least after emptying the cache?

evmar:
This also broke for me after sync.  The workaround appears to be `bazel clean --expunge`

Issue #1009, opened 2019-03-27
tsickle gets in an infinite loop for function that use their own type in their type declartion


rkirov:
For example:

```ts
export function f(): typeof f { return f;}
```

Surprisingly TS is ok with that, so tsickle should not get into an infinite loop and emit something sensible (maybe '?' if closure doesn't support it).

This did really happen in real code - https://github.com/angular/angular/commit/e4c1c88cbc4c28f1a9d9fca4b93432b3beab620e#diff-a66758223216005c664ce98175c4164fR1191

blickly:
I tried to repro this issue in a unit test (see linked PR), but didn't hit the crash.  Is it possible this has been fixed since March?

brad4d:
Google internal issue http://b/142073208 created

blickly:
This was fixed by https://github.com/angular/tsickle/pull/1092

PR #1092, opened 2019-10-03
Fix for infinite loop issue reported in #1009

blickly:
As a first step to fixing https://github.com/angular/tsickle/issues/1009, I've tried to reproduce the issue in a unit test.

Surprisingly, it already passes.

### DIFF
--- a/src/type_translator.ts
+++ b/src/type_translator.ts
@@ -624,26 +624,7 @@ export class TypeTranslator {
       }
       return typeStr;
     } else if (type.objectFlags & ts.ObjectFlags.Anonymous) {
-      if (!type.symbol) {
-        // This comes up when generating code for an arrow function as passed
-        // to a generic function.  The passed-in type is tagged as anonymous
-        // and has no properties so it's hard to figure out what to generate.
-        // Just avoid it for now so we don't crash.
-        this.warn('anonymous type has no symbol');
-        return '?';
-      }
-
-      if (type.symbol.flags & ts.SymbolFlags.Function ||
-          type.symbol.flags & ts.SymbolFlags.Method) {
-        const sigs = this.typeChecker.getSignaturesOfType(type, ts.SignatureKind.Call);
-        if (sigs.length === 1) {
-          return this.signatureToClosure(sigs[0]);
-        }
-        this.warn('unhandled anonymous type with multiple call signatures');
-        return '?';
-      } else {
-        return this.translateAnonymousType(type);
-      }
+      return this.translateAnonymousType(type);
     }
 
     /*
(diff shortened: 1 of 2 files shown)


test/tsickle_test.ts
```diff
@@ -105,6 +105,21 @@ describe('emitWithTsickle', () => {
...
+    };
+
+    const jsSources = emitWithTsickle(
+        tsSources, {
+          preserveConstEnums: true,
+          module: ts.ModuleKind.ES2015,
+        },
+        {es5Mode: false, googmodule: false});
+
+      expect(jsSources['a.js']).toContain(`export function f() { return f; }`);
```
evmar: I would expect this output to contain some Closure stuff in it (e.g. "exports.f = ..."), is it possible this isn't running the right emit?

blickly: The full emit that I saw was:
```ts
/**
     * @suppress {checkTypes,constantProperty,extraRequire,missingOverride,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
     */
    /**
     * @return {?}
     */
    export function f() { return f; }
``

So it looks like it is producing the closure type annotations (which I has assumed would be enough to trigger the bug).  Do you think trying goog.module emit would be needed to trigger the bug?

Comment (evmar) [reply test/tsickle_test.ts]: I suspect this:
https://github.com/angular/tsickle/blob/3276f179c3d52ab5a7d3f2d88f57e92797f2bd1b/test/tsickle_test.ts#L37

This is the flag that we flipped in google but forgot about in this test suite.  :(

I think we should drop the flag entirely, but try flipping it here for now to see if that unsticks you.

Comment (blickly) [reply test/tsickle_test.ts]: Aha, that was it. Thanks Evan!

**Comment by blickly** [comment]
OK, this looks good to go now

File: src/type_translator.ts [hunk 2]
```diff
@@ -624,6 +624,7 @@ export class TypeTranslator {
       }
       return typeStr;
     } else if (type.objectFlags & ts.ObjectFlags.Anonymous) {
+      this.seenAnonymousTypes.add(type);
```
evmar: It feels like this belongs in translateAnonymousType.  Maybe the rest of the code in this block (before the call to this.translateAnonymousType) belongs in translateAnonymousType as well?

blickly: Fair enough.  After doing so, it made it more obvious that type.symbol was being checked redundantly, so I simplified slightly.

evmar: Looks great.

evmar:
I think the CI is failing because you need to run the formatter (it's printed in the CI log)

blickly:
Yes, that was exactly it.  Thanks!

evmar:
What are the chances we need to TGP this first?  I always think "what could possibly go wrong" and then something inevitably goes horribly wrong...

PR #1093, opened 2019-10-03
drop the 'untyped' flag

evmar:
We no longer use this in Google, and instead always use type info.
We shouldn't keep this code path when we aren't using it.


mprobst:
Didn't we still use this for the `ES5` mode output (and maybe got a bit of performance out of it)?

I think we should still drop it, but it might be worth confirming this first, and also testing if making ES5 emit typed works out in google3.

evmar:
(This was prompted by blickly discovering that one of our tests was running in untyped mode.)
I thought for ES5 we don't use jsdoc_annotator at all, so it's irrelevant there.

What I really discovered with this change is that we still use untyped mode, for the bugs linked from b/142075912.  We'll need to resolve that before we can land this.  Since it wasn't hard to make this change I might just delete it because we likely won't be able to use it for a while.

PR #1094, opened 2019-10-04
update tsickle to closure compiler 20190929

evmar:
The version of closure compiler referenced here is only used in the
tsickle test suite, so it isn't so important, but it's worth using
a recent-ish version just to ensure compatibility.


yarn.lock
```diff
@@ -220,7 +220,16 @@ camelcase@^4.1.0:
   resolved "https://registry.yarnpkg.com/camelcase/-/camelcase-4.1.0.tgz#d545635be1e33c542649c69173e5de6acfae34dd"
   integrity sha1-1UVjW+HjPFQmScaRc+Xeas+uNN0=
 
-chalk@^1.0.0, chalk@^1.1.3:
+chalk@2.x, chalk@^2.3.0:
```
evmar: Note: I was confused by this bit.  It appears tslint uses chalk 1.x, while the new google-closure-compiler uses chalk 2.x.

PR #1097, opened 2019-10-14
Fix enum emit for constant string values.

mprobst:
**MULTIPLE COMMITS** please review independently :-)

TypeScript resolves and inlines compile time constant values for enums.
In doing so, it also elides any imports that are only used in
initializer expressions that resolve to compile time constants:

    import {Constant} from 'somewhere';
    enum MyEnum { Val: Constant }

Here, if `Constant` really is a compile time constant string or number,
the import for `'somewhere'` will not be emitted.

tsickle transforms enums into variable/object literal pairs. Because
TypeScript still elides the import, but no longer does the constant
substitution, tsickle would emit expressions that referenced module
imports that weren't actually imported, causing reference errors.

This change fixes the problem by unconditionally replacing compile time
constant enum members with their known value.

Another possible fix is enabling the `--isolatedModules` compiler flag
that prevents this sort of inlining. We should possibly advise that, but
it also has other consequences (e.g. disallowing const enums in `.d.ts`
files), so fixing this bug in isolation is still worthwhile.

### DIFF
--- a/test_files/enum/enum.js
+++ b/test_files/enum/enum.js
@@ -93,8 +93,8 @@ EnumWithNonConstValues[EnumWithNonConstValues.Scheme] = 'Scheme';
 EnumWithNonConstValues[EnumWithNonConstValues.UserInfoRenamed] = 'UserInfoRenamed';
 /** @enum {string} */
 const StringEnum = {
-    STR: 'abc',
-    OTHER_STR: 'xyz',
+    STR: "abc",
+    OTHER_STR: "xyz",
 };
 /** @enum {number} */
 const StringKeyEnum = {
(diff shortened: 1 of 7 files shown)


test_files/enum/enum.js
```diff
@@ -93,8 +93,8 @@ EnumWithNonConstValues[EnumWithNonConstValues.Scheme] = 'Scheme';
 EnumWithNonConstValues[EnumWithNonConstValues.UserInfoRenamed] = 'UserInfoRenamed';
 /** @enum {string} */
 const StringEnum = {
-    STR: 'abc',
-    OTHER_STR: 'xyz',
+    STR: "abc",
```
evmar: FWIW I typically merge change+test rebaseline in a single commit, so that tools like "git bisect" see a green test suite on every commit.  Pretty minor though.

test_files/enum_ref_import/enum_ref_import.js
```diff
@@ -0,0 +1,24 @@
+/**
+ *
```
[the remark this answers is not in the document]
mprobst: I don't know either. Man, comments.

test_files/enum/enum.js
mprobst: Good point, done.

PR #1098, opened 2019-10-15
Update docs on developing tsickle.

mprobst:
There's an annoying catch-22 between using `bazel run @nodejs//:yarn` to
install dependencies and using the bazel version installed by yarn,
which would require using `yarn install` to get bazel in the first
place.

In practice, I've had fewer problems with the bazel version, so this
changes the docs to recommend installing a local bazel version and using
it to install dependencies. This also removes the workaround around
`bazel run` and the associated script; I believe this has been fixed a
long time ago in bazel itself.

This change also deletes the .vscode/launch configuration and changes
documentation to warn on debugging using VS Code. I think this has been
broken ever since we moved to bazel and it seems impossible to get VS
Code to understand the mapping between paths in bazel-bin and the
original source tree.


mprobst:
/CC @alexeagle in case there's a known way to get VS Code debugging of NodeJS tests to work.

mprobst:
I think this is somewhat related to https://github.com/bazelbuild/rules_nodejs/issues/177, but the deeper problem is that the NodeJS launcher resolves everything to absolute paths like `/usr/local/google/home/martinprobst/.cache/bazel/_bazel_martinprobst/1676c3df3a56b0dc61e541b515b67b3c/execroot/tsickle/bazel-out/k8-fastbuild/bin/test/golden_test_loader.js`. I believe there's no way to re-establish a mapping back. The issue is probably confounded by our source maps containing relative compiler paths (i.e. `../../../../test/golden_test_loader.ts`, accounting for `bazel-out/k8-fastbuild/bin`).

README.md
```diff
@@ -113,26 +113,31 @@ Example:
...
+### Bazel install
 
-> This avoids occupying the `bazel` server, so that `yarn` can call `bazel`
-> again.
-> Ideally we should just use `bazel-run.sh @nodejs//:yarn`, see
-> https://stackoverflow.com/questions/47082298/how-can-users-get-bazel-run-sh
+We use [bazel](https://bazel.build/) to build, and are pinned to a specific
+version of it in `package.json` for reproducible builds. The build rules check
+for a compatible version of bazel, so it is generally safe to use your local
+installed version. If in doubt, you can run `yarn bazel` instead of `bazel` in
```
evmar: I discovered that 'yarn bazel' does the wrong thing in some situations, because yarn runs its tools with a new temp dir on each invocation which confuses the bazel cache.  I think instead you should recommend something like

"you can run `node_modules/.bin/bazel` instead"

README.md
```diff
@@ -141,21 +146,25 @@ to install the dependencies.
...
+
+The break in specific code locations you can add `debugger;` statements in the
+source code.
 
-VS Code can also connect using the inspect protocol. It doesn't support automatically reconnecting
-or any way to re-run the test suite though, so it is a less convenient. You can start the node
-process passing an extra `--node_options=--debug-brk` (in addition to the parameters above) to have
-Node wait before program execution, so you have time to attach VS Code.
+Note: IDEs such as VS Code have support for the inspect protocol, but the
+integration does not work due to bazel's complex directory layout.
```
evmar: This is fine, but if you wanted to tinker with it more I think you might need to run `nodejs --preserve-symlinks` .  That at least fixed another thing I found when tinkering with bazel.

README.md
alexeagle: that was fixed by adding `--incompatible_strict_action_env` - `yarn bazel` is fine (Angular team does it that way)

README.md
alexeagle: Yes we always run `node --preserve-symlinks --preserve-symlinks-main` - the latter avoids the entry point script getting `realpath`ed

alexeagle:
Even in repos with a dependency on `@bazel/bazel`, I myself still use a locally installed bazel. But it has downfalls. In particular on glinux there seems to be no way to get an earlier version than latest, and not all repos (and their transitive deps) work with the latest due to breaking changes.

If you want to rely on the Bazel on your machine rather than the one in npm, you ought to have a `.bazelversion` file and recommend that users use Bazelisk. At this point Bazel is 1.0 and so I'd expect any 1.x to work, but they haven't promised any more than 3 months before Bazel 2.0 and you'll probably have bad error messages when we account for breaking changes and you still have 1.x installed.

alexeagle:
As for sourcemap paths, the fact that the output directory isn't in the same tree with the inputs makes relative sourcemap urls inherently broken. It doesn't matter if we resolve to absolute paths, there's just no path that makes sense to the tool writing the sourcemap other than one with lots of `../../` prefix. Making matters worse, such a prefix has three leading segments corresponding to `bazel-out/arch/bin` which is the short_path of the bin_dir in starlark - but then you have the sugar symlink `bazel-bin` so if you reference the `.js` file that way, you don't even map correctly back to the `.ts` file since you walk two segments too high in the filesystem.

You could strip them and pretend that everything is in a unionfs like the runfiles forest (I think this was always John Barton's position), but that doesn't help an editor that sees a `.js` in bazel-bin when the corresponding `.ts` is not there.

I think the only principled answer is that the tool writes a relative path based on what it saw (sources in execroot/srcdir and outputs in outputdir) and then you have to fix up the sourcemaps if you move the outputs next to the inputs later (like in runfiles tree). In practice I don't see how to do that reliably.

Another possible answer which is truer to the nodejs ecosystem is to copy all the sources to the outputdir before running the tool, then hand the tool inputs from the outputdir, then the sourcemaps are truly pointing to files in the same directory.

README.md
mprobst: The issue is that when we launch node, we've already resolved all the paths to absolute locations that are in some `$TMP/bazel-cache` directory. At that point source map resolution appears impossible, as Alex says.

mprobst:
Re bazelversion etc, I think I'd like to keep this as simple as possible, without asking users to install another tool. Let's see how well this works, we can still revisit if it turns out to be a problem.

Re sourcemaps, indeed as commented above, this seems unfixable for the time being. I think using execpaths (runfiles/unionfs style) has the advantage that we typically have only a few compilers that are configurable in their path resolution, but many more tools that execute JS and less control over those. But in either case, it won't solve this issue here.

PR #1101, opened 2019-10-16
Add support to tsickle for unwrapping CommaListExpressions

blickly:
This allows it to generate statement-level exports from comma expressions with
10 or more children.


src/googmodule.ts
```diff
@@ -220,16 +220,16 @@ function rewriteCommaExpressions(expr: ts.ExpressionStatement): ts.Statement[]|n
   // 2) a tree of  "binary expressions" whose contents are comma operators
   if (ts.isBinaryExpression(expr.expression) &&
       expr.expression.operatorToken.kind === ts.SyntaxKind.CommaToken) {
-    return visit(expr.expression)
+    return visitBinaryCommas(expr.expression)
```
evmar: Trailing semi here.

test_files/export_multi/export_multi.ts
```diff
@@ -15,7 +15,15 @@ export const {
   ORANGE,
 } = Fruit;
 
-export const {a, b} = {
+export const {a, b, c, d, e, f, g, h, i, j} = {
```
mprobst: can you add a comment why we're testing 10 elements here?

mprobst: this test was here (I think) to ensure the intializer on the right hand side doesn't get repeated. Maybe add a separate export statement for the "10 members" case?

src/googmodule.ts
```diff
@@ -209,19 +209,30 @@ function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
...
+  if (ts.isBinaryExpression(expr.expression) &&
+      expr.expression.operatorToken.kind === ts.SyntaxKind.CommaToken) {
+    return visitBinaryCommas(expr.expression)
+  }
 
   // Recursively visit comma-separated subexpressions, and collect them all as
   // separate expression statements.
-  return visit(expr.expression);
-  function visit(expr: ts.Expression): ts.Statement[] {
+  function visitBinaryCommas(expr: ts.Expression): ts.Statement[] {
```
mprobst: I guess technically we could end up with a binary expression containing a comma list expression in turn, but that's probably practically not going to happen, is it? You could consider adding a comment.

test_files/export_multi/export_multi.ts
blickly: Added

src/googmodule.ts
blickly: Added.

test_files/export_multi/export_multi.ts
blickly: OK, sure.  It doesn't hurt to have more cases tested.  Added separate 2 element and 10 element cases.

src/googmodule.ts
blickly: I don't think TS produces cases with mixed comma structure, so I can't e.g. write a unit test, but I reworked the algorithm so it at least looks like it should work with mixed comma expression trees.

test_files/export_multi/export_multi.ts
blickly: Added

blickly: Added both 2-element and 10-element cases.

src/googmodule.ts
blickly: I don't think TS ever produces "mixed comma trees" with both styles of expressions, so I can't e.g. write a good unit test, but I've shifted things around so it at least looks like it would work with a mixed comma tree.

src/googmodule.ts
```diff
@@ -209,17 +209,29 @@ function rewriteModuleExportsAssignment(expr: ts.ExpressionStatement) {
...
+    // 1) a tree of  "binary expressions" whose contents are comma operators
     if (ts.isBinaryExpression(expr) && expr.operatorToken.kind === ts.SyntaxKind.CommaToken) {
       return visit(expr.left).concat(visit(expr.right));
     }
+    // or,
+    // 2) a "comma list" expression, where the subexpressions are in one array
+    if (expr.kind === ts.SyntaxKind.CommaListExpression) {
+      const commaList = expr as ts.CommaListExpression;
+      // TODO(blickly): Simplify using flatMap once node 11 available
+      return commaList.elements.reduce((acc: ts.Statement[], x) => acc.concat(visit(x)), []);
```
evmar: I guess this does N separate concats.   You might instead pass all the subarrays to concat, via something like:

`[].concat(elements.map(stuff)...)`

I don't think it matters too much either way though.

src/googmodule.ts
mprobst: Sorry, I wasn't precise. I meant to say that we don't need to handle that problem, and a comment would have been enough. Doesn't hurt to cover it though.

mprobst:
@evmar will you land this once good?

evmar:
Yeah, or Ben can land it himself :)

blickly:
I think the plan is to move this repo out of the angular org, but until we do, I don't think I have write access.

I see the "Only those with write access to this repository can merge pull requests." message, for example

src/googmodule.ts
blickly: Good point. That way is certainly simpler.

src/googmodule.ts
blickly: Now that I've refactored it, the code that does handles the pattern seems simpler anyway.

blickly:
Thanks evan!

PR #1102, opened 2019-10-18
 Add a IterableIterator as an alias of IteratorIterable.

concavelenz:


evmar:
Is there a relevant test we can write?  I think you can do:

```
mkdir test_files/iterator
touch test_files/iterator/iterator.ts
touch test_files/iterator/iterator.js
```

and then edit iterator.ts until it has the code you care about, then run

```
UPDATE_GOLDENS=1 bazel run test:golden_test
```

and then finally,

```
bazel test test:closure_e2e_test
```

verifies the final code succeeds in compilations.

mprobst:
Closing in favor of #1127.

PR #1103, opened 2019-10-21
TS3.7 clutz alias

mprobst:
*** REVIEW COMMITS SEPARATELY ***

* Adjust Clutz alias emit for TypeScript 3.7's change to module visibility. https://github.com/microsoft/TypeScript/issues/34612
* Fix a bug with renamed module exports.

### DIFF
--- a/src/tsickle.ts
+++ b/src/tsickle.ts
@@ -335,10 +335,14 @@ function addClutzAliases(
       // Use the localName for the export then, but publish under the external name.
       localName = declaration.propertyName.text;
     }
-    globalSymbols +=
-        `\t\texport {${localName} as module$contents$${clutzModuleName}_${symbol.name}}\n`;
-    nestedSymbols +=
-        `\t\texport {module$contents$${clutzModuleName}_${symbol.name} as ${symbol.name}}\n`;
+    const mangledName = `module$contents$${clutzModuleName}_${symbol.name}`;
+    globalSymbols += `\t\texport {${localName} as ${mangledName}}\n`;
+    // TODO(mprobst): Once tsickle is on TS3.7, the two lines below can be replaced with
+    // "export {localName};". However in TS3.5, localName resolves within the module, so
+    // exporting {localName} causes a circular definition error. The workaround is to import the
+    // mangled name.
+    nestedSymbols += `\t\timport ${localName}$clutz = ಠ_ಠ.clutz.${mangledName};\n`;
+    nestedSymbols += `\t\texport {${localName}$clutz as ${symbol.name}};\n`;
   }
 
   dtsFileContent += 'declare global {\n';
--- a/test_files/reexport.declaration/clutz_user.d.ts
+++ b/test_files/reexport.declaration/clutz_user.d.ts
@@ -1,0 +1,9 @@
+/** @fileoverview Tests that renamed exports get emitted with the correct clutz alias names. */
+
+export {};
+
+declare const moduleInternalName:
+    typeof ಠ_ಠ.clutz.module$contents$test_files$reexport$declaration$renamed_export_MOTHER;
+
+declare const exportedName:
+    typeof ಠ_ಠ.clutz.module$exports$test_files$reexport$declaration$renamed_export.MOTHER;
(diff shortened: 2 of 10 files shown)


mprobst:
@evmar ready for review. Make sure to review the two commits separately, I'm just bundling up fixes for the Clutz aliases here.

test_files/reexport.declaration/export.d.ts
```diff
@@ -6,15 +6,15 @@ declare const NON_DIRECT_TO_BE_RENAMED = 0;
...
-		export {NON_DIRECT as module$contents$test_files$reexport$declaration$export_NON_DIRECT}
-		export {NUM_CONSTANT as module$contents$test_files$reexport$declaration$export_NUM_CONSTANT}
-		export {OTHER as module$contents$test_files$reexport$declaration$export_OTHER}
+		export {NON_DIRECT_TO_BE_RENAMED as module$contents$test_files$reexport$declaration$export_NEW_NAME};
+		export {NON_DIRECT as module$contents$test_files$reexport$declaration$export_NON_DIRECT};
+		export {NUM_CONSTANT as module$contents$test_files$reexport$declaration$export_NUM_CONSTANT};
+		export {OTHER as module$contents$test_files$reexport$declaration$export_OTHER};
 	}
 	namespace ಠ_ಠ.clutz.module$exports$test_files$reexport$declaration$export {
-		export {module$contents$test_files$reexport$declaration$export_NEW_NAME as NEW_NAME}
```
evmar: Oh I see I missed this part, this is the part you're changing?

mprobst:
@evmar ah no, I actually went full circle in what I was trying to achieve here. Argh.

I think I can only get this to work in TS3.6 & TS3.7 with yet another ugly indirection through an explicit, separate import of the symbol. At least we can get rid of that once we move to TS3.7.

mprobst:
PTAL, that is.

test_files/reexport.declaration/clutz_user.d.ts
```diff
@@ -0,0 +1,9 @@
+/** @fileoverview Tests that renamed exports get emitted with the correct clutz alias names. */
```
evmar: This change adds a d.ts but has no corresponding code or baseline change ... ?

Is it just testing that this code compiles?

src/tsickle.ts
```diff
@@ -335,10 +335,13 @@ function addClutzAliases(
...
     }
-    globalSymbols +=
-        `\t\texport {${localName} as module$contents$${clutzModuleName}_${symbol.name}}\n`;
-    nestedSymbols +=
-        `\t\texport {module$contents$${clutzModuleName}_${symbol.name} as ${symbol.name}}\n`;
+    const mangledName = `module$contents$${clutzModuleName}_${symbol.name}`;
+    globalSymbols += `\t\texport {${localName} as ${mangledName}}\n`;
+    // TODO(mprobst): in TS3.7, this could be just "export {localName};". However in TS3.5,
+    // localName resolves within the module, so exporting {localName} causes a circular definition
+    // error. The workaround is to import the mangled name.
```
evmar: I reread this a bunch of time and still can't follow this comment.

Normally 'mangled' names means you did some damage to a name to avoid a conflict (like $clutz below), but here I think the mangledName var is just the Closure post-module-processing global name of the symbol, isn't it?  Or I think maybe this comment is saying that the $clutz mangling can be removed in TS37, and the mention of mangling here is unrelated to the variable named 'mangled' just above it... ?  Or is the comment referring to the code above it?

test_files/reexport.declaration/clutz_user.d.ts
mprobst: Yes. We have a test (e2e_dts_test or so) that just passes all of these through tsc.

src/tsickle.ts
mprobst: `mangledName` refers to the name as it will be mangled by Closure. I think it's correct to describe Closure's post processing of the name into a single variable as mangling?

The comment says `in TS3.7, this could be just "export {localName};"` in reference to the lines below it (like comments usually do, right?). The intended meaning is that once tsickle is on TS3.7, you can replace the two lines below with `nestedSymbols += '\t\texport {${localName}}';`. I've updated the comment to say `Once tsickle is on TS3.7, the two lines below can be replaced with "export {localName};".`, hopefully that's explicit enough?

evmar: I see.  I think I was confused by "The workaround is to import the mangled name.", where the workaround is to mangle the name with $clutz and use that mangled name to refer to the Closure-mangled name.  Two mangled names.

PR #1105, opened 2019-10-23
Add "Generated from" comment to @fileoverview of generated JS files.

nbeloglazov:
It will be used by other tooling (e.g. Kythe) that need to connect
JS file back to original TS file.

Tested:
  golden tests

### DIFF
--- a/src/fileoverview_comment_transformer.ts
+++ b/src/fileoverview_comment_transformer.ts
@@ -9,6 +9,7 @@
 import * as ts from 'typescript';
 
 import * as jsdoc from './jsdoc';
+import * as path from './path';
 import {createNotEmittedStatement, reportDiagnostic, synthesizeCommentRanges, updateSourceFileNode} from './transformer_util';
 
 /**
@@ -22,12 +23,21 @@ const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
  * Given a parsed \@fileoverview comment, ensures it has all the attributes we need.
  * This function can be called to modify an existing comment or to make a new one.
  *
+ * @param source Original TS source file. Its path is added in \@fileoverview.
  * @param tags Comment as parsed list of tags; modified in-place.
  */
-function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
+function augmentFileoverviewComments(
+    options: ts.CompilerOptions, source: ts.SourceFile, tags: jsdoc.Tag[]) {
   // Ensure we start with a @fileoverview.
-  if (!tags.find(t => t.tagName === 'fileoverview')) {
-    tags.splice(0, 0, {tagName: 'fileoverview', text: 'added by tsickle'});
+  let fileOverview = tags.find(t => t.tagName === 'fileoverview');
+  if (!fileOverview) {
+    fileOverview = {tagName: 'fileoverview', text: 'added by tsickle'};
+    tags.splice(0, 0, fileOverview);
+  }
+  if (options.rootDir != null) {
+    // This comment is read by other tools so it's important that its format
+    // doesn't change.
+    fileOverview.text += `\nGenerated from: ${path.relative(options.rootDir, source.fileName)}`;
   }
 
   // Find or create a @suppress tag.
@@ -71,7 +81,8 @@ function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
  * A transformer that ensures the emitted JS file has an \@fileoverview comment that contains an
  * \@suppress {checkTypes} annotation by either adding or updating an existing comment.
  */
-export function transformFileoverviewCommentFactory(diagnostics: ts.Diagnostic[]) {
+export function transformFileoverviewCommentFactory(
+    options: ts.CompilerOptions, diagnostics: ts.Diagnostic[]) {
   return (): (sourceFile: ts.SourceFile) => ts.SourceFile => {
     function checkNoFileoverviewComments(
         context: ts.Node, comments: jsdoc.SynthesizedCommentWithOriginal[], message: string) {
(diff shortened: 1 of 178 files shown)


nbeloglazov:
I don't know what is the best way to get relative file TS file path so I used `GoogModuleProcessorHost` which seems to be giving correct value for tests. I assume that's because GoogModuleProcessorHost in tests is designed for that. Not sure what it will give in google3. 

For context I need this info in order to implement JS <=> TS Kythe support in Google. For that I need to know from which TS file JS file was generated. TS file must be google3-related, including `blaze-genfiles` prefix if TS file was genereated.

I tried using `source.fileName` but it used full absolute path so it doesn't work.

src/fileoverview_comment_transformer.ts
```diff
@@ -22,13 +23,20 @@ const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
...
-function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
+function augmentFileoverviewComments(
+    host: GoogModuleProcessorHost, source: ts.SourceFile, tags: jsdoc.Tag[]) {
   // Ensure we start with a @fileoverview.
   if (!tags.find(t => t.tagName === 'fileoverview')) {
     tags.splice(0, 0, {tagName: 'fileoverview', text: 'added by tsickle'});
   }
+  const fileOverview = tags.find(t => t.tagName === 'fileoverview');
+  if (fileOverview) {
+    fileOverview.text += `\nGenerated from: ${host.fileNameToModuleId(source.fileName)}`;
```
mprobst: that's some really bad API naming on our side here (module ID turns out to be the path?). Nothing to fix in this CL.

test_files/abstract/abstract.js
```diff
@@ -1,5 +1,6 @@
 /**
  * @fileoverview added by tsickle
+ * Generated from: test_files/abstract/abstract.ts
```
mprobst: It's a bit odd that we add an implementation detail comment like this into a doc comment (`/** */`), but we need to add the doc comment anyway due to JSCompiler only processing `@suppress` directives in JSDoc. Given that, I think this is the right place, there's no point in adding another `// ` comment.

mprobst:
@nbeloglazov have you tested this in google3? You can run `tsync.sh 1105` to create a CL with this change to test.

src/fileoverview_comment_transformer.ts
evmar: I think module id happens to be the path, but it's not part of its guarantee.

Node docs say "The identifier for the module. Typically this is the fully resolved filename."
https://nodejs.org/api/modules.html#modules_module_id

For example, does this include blaze-bin or not?  Should it?  Which behavior do we depend on in both of its use cases?

nbeloglazov: I'll see how it works in google3. I kinda suspect it won't work as expected.

nbeloglazov: Following Evan's suggestion, changed code to use `CompilerOptions.rootDir` to produce root-relative paths. Checked and it works with google3 generated .ts files as well.

nbeloglazov:
Yes, tested in google3 and changed implementation to not use GoogModule host.

src/fileoverview_comment_transformer.ts
```diff
@@ -22,13 +23,20 @@ const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
  * Given a parsed \@fileoverview comment, ensures it has all the attributes we need.
  * This function can be called to modify an existing comment or to make a new one.
  *
+ * @param options Compiler options.
```
evmar: Leave out this line, because it just restates the type of the param

src/fileoverview_comment_transformer.ts
```diff
@@ -22,13 +23,20 @@ const FILEOVERVIEW_COMMENT_MARKERS: ReadonlySet<string> =
...
  */
-function augmentFileoverviewComments(tags: jsdoc.Tag[]) {
+function augmentFileoverviewComments(
+    options: ts.CompilerOptions, source: ts.SourceFile, tags: jsdoc.Tag[]) {
   // Ensure we start with a @fileoverview.
   if (!tags.find(t => t.tagName === 'fileoverview')) {
     tags.splice(0, 0, {tagName: 'fileoverview', text: 'added by tsickle'});
   }
+  const fileOverview = tags.find(t => t.tagName === 'fileoverview');
+  if (fileOverview && options.rootDir != null) {
```
evmar: Put some comment around here about how this 'generated from' string is read by other code and shouldn't be changed

evmar: What about when fileoverview==null?  I know tsickle always inserts a fileoverview tag, so maybe this is at the wrong point (?)

nbeloglazov: Yes, expectation that fileoverview will always be there similar to rootDir. Can you think of a better place to put it?

nbeloglazov: Done.

src/fileoverview_comment_transformer.ts
nbeloglazov: Done.

src/fileoverview_comment_transformer.ts
evmar: Oh I see, I misunderstood this code.

How about

```
let fileOverview = tags.find(t => t.tagName === 'fileoverview');
if (!fileOverview) {
  fileOverview = {tagName: 'fileoverview', text: 'added by tsickle'}
  tags.splice(0, 0, fileOverview);
}
your code here
```

nbeloglazov: Done.

PR #1107, opened 2019-11-01
handle one more async case

evmar:
If you tell JSCompiler to emit ES2017, one of our async test cases
(a static async on a class) starts failing.  It appears that even in
ES5 this code patterns ends up with a this type of {?} but for some
reason the failure doesn't trigger, something about the JSCompiler
passes for ES5 emit.

This change fixes the test by handling static async, by busting the
type of 'this'.  This reads a bit wrong to me, but I am told that
  class C {
    static f() {}
    static g() { this.f(); }
  }
is illegal in JSCompiler anyway, so it's ok to break any code that
uses 'this' in a static.


evmar:
@mprobst take a look and let me know what you think, but I haven't run this one against g3 yet.

mprobst:
Looks a bit odd indeed. Maybe we should formulate the comment as a to-do,
predicated on a fix in closure?

PR #1108, opened 2019-11-04
remove exports = {}

evmar:
Two commits.


evmar:
@blickly This is the final patch we should merge.

test_files/namespaced/export_namespace.ts
```diff
@@ -1,3 +1,8 @@
+// TODO(#132): namespaces currently don't emit properly.
+// This workaround at least makes them compile.
+// @ts-ignore
+exports = {};
+
 // tslint:disable:no-namespace
 
 export namespace valueNamespace {
```
blickly: I think that a nicer way to handle this case is:

```typescript
namespace valueNamespace {
  export class ValueClass {}
}
namespace typeNamespace {
  export interface Interface {}
}

export {valueNamespace, typeNamespace)
```

which I think just happens to have an emit where the `goog.module` exports are in the expected form.

That said, since namespaces are against our style guide, it might not be as important.

test_files/namespaced/export_enum_in_namespace.ts
```diff
@@ -3,6 +3,11 @@
...
  */
 
+// TODO(#132): namespaces currently don't emit properly.
+// This workaround at least makes them compile.
+// @ts-ignore
+exports = {};
+
 // tslint:disable:no-namespace
 
 export namespace foo {
```
blickly: I'd prefer to rewrite the code here as

```typescript
namespace foo { ... }
export {foo}
```

as it seems slightly less ugly than the @ts-ignore hack.  The reopened namespace case truly requires that hack, so we'll still have it in our test suite.

test_files/namespaced/export_namespace.ts
evmar: I left this one alone, and changed the other one.

test_files/namespaced/export_enum_in_namespace.ts
evmar: Done.

evmar:
Now three commits; the first avoids `export namespace`.

blickly:
I'm seeing the change to `export_enum_in_namespace.ts` but not to `export_namespace.ts`.  Can you check that you've pushed that file as well?

evmar:
Oh I intentionally left that one alone, so we'd have coverage of the 'exports = {}' hack as well.  Do you think we should do both?

blickly:
Yeah, I don't think that the `exports={}` hack is necessary for namespaces, so I'm in favor of removing it for those cases so our tests describe the "best practices".

If there is a case with a truly mutable export (i.e. export let ...) that might be a good candidate for actually using the hack.

evmar:
Done.  Also I expanded the comment on the one remaining ts-ignore that we should keep at least one around.

evmar:
One last fixup: deleting this code revealed some functions that are no longer used.

evmar:
I'm going to merge this because I think we are extremely close to landing.  If we need any last-second tsickle changes we can always pull a branch into g3.

PR #1109, opened 2019-11-06
rel: 0.37.1

mprobst:


mprobst:
@IgorMinar tsickle is owned by the Angular user, but I seem to have lost
access to the account (in valentine). Could you hand it over to me?
https://www.npmjs.com/package/tsickle

PR #1111, opened 2019-11-07
Set up google-wombot for npm publish.

mprobst:
google-wombot is an NPM registry that proxies package publishing. It
uses github publish permissions to authenticate users, meaning we have
one fewer ACL/permission set to worry about.


mprobst:
@soldair could you take a look at this? In particular whether users will need to login to the NPM registry first?

soldair:
@evmar, its staffed by the virtual node team :fearful: but is a critical dependency of the cloud client libraries team, and others which should ensure it's support going forward.

soldair:
lgtm. 

a difference between wombat and npm is that tokens that can publish anything are limited by 24 hours. 

If you want to create a long lived token can login with a package specific registry url. If you get tired of logging in or use tokens for automated publishes via CI you might want something like this.
https://sites.google.com/corp/google.com/node-team/publishing-npm-packages#h.p_xUnJvEi73t0O

PR #1113, opened 2019-11-08
Reduce dependencies of demo.ts

rubenlg:
The `mkdirp` dependency is not necessary. fs.mkdirSync has a recursive mode.


evmar:
Thanks!  Can you update package.json (and yarn.lock) as well?

rubenlg:
Done. PTAL.

PR #1114, opened 2019-11-08
Visit enum nodes in jsdocTransformer to validate the enum's JSDoc

### LINKED ISSUE #998: @enum and JSDoc handling
@enum is in our list of banned jsdocs - https://github.com/angular/tsickle/blob/master/src/jsdoc.ts#L148
but they still appear in our codebase. This is because, we completely remove the original comment along with the node as part of the enum rewriting part of tsickle - https://github.com/angular/tsickle/blob/master/src/enum_transformer.ts#L158

and then there is nothing to catch and error in the second phase. Also the original comment text (even if it has no @enum) is not kept in the output.

rrdelaney:
Enum node types were not being validates along with other nodes in jsdocTransformer, and were only being seen after enumTransformer, which turned the enum into a valid closure enum, causing the original JSDoc to never be validated. This change makes jsdocTransformer visit (but not actually transform) the enum nodes and validates them by calling getJSDoc.

Fixes #998

### DIFF
--- a/src/jsdoc_transformer.ts
+++ b/src/jsdoc_transformer.ts
@@ -995,6 +995,16 @@ export function jsdocTransformer(
         return result;
       }
 
+      /**
+       * Visits enum declarations to check for validity of JSDoc comments without transforming the
+       * node at all.
+       */
+      function visitEnumDeclaration(node: ts.EnumDeclaration) {
+        // Calling `getJSDoc` will validate and report any errors, but this code
+        // doesn't really care about the return value.
+        moduleTypeTranslator.getJSDoc(node, /* reportWarnings */ true);
+      }
+
       function visitor(node: ts.Node): ts.Node|ts.Node[] {
         if (transformerUtil.isAmbient(node)) {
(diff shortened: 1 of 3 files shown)


test_files/jsdoc/jsdoc.ts
```diff
@@ -143,3 +143,11 @@ const DEFINE_WITH_INFERRED_TYPE = false;
  * @define
  */
 const DEFINE_WITH_DECLARED_TYPE: string = 'y';
+
+/**
```
mprobst: Can you move this to a new test? `jsdoc.ts` is overflowing, making it hard to debug specific issues.

Also, can you include a `/** @fileoverview */` comment in that new test that documents what's being tested by the code sample?

rrdelaney: Done, added the new test in `test_files/jsdoc`, because that still seems like the right place for it.

rrdelaney:
I'll follow up on the internal bug for cleaning up existing usages.

rrdelaney:
google3 is cleaned up now, so this should be ready for review 😄

mprobst:
Could you rebase on master?

mprobst:
LGTM

PR #1117, opened 2019-11-12
Handle `{default as Foo}` imports from Clutz.

mprobst:
In ES modules (and thus TypeScript), the default export is simply a
property named `default` on the exports object. tsickle however maps
default imports from `goog:` modules directly to the exports object, to
better match common `goog.module` usage.

Previously, tsickle failed to recognize `{default as Foo}` style
renaming, causing it to alias an incorrect type symbol to the export, in
turn causing failed type resolution in Closure (see the test for an
example). This change recognizes the renaming style, fixing the problem.

### DIFF
--- a/test/e2e_closure_test.ts
+++ b/test/e2e_closure_test.ts
@@ -26,6 +26,7 @@ describe('golden file tests', () => {
     goldenJs.push('third_party/tslib/externs.js');
     goldenJs.push('third_party/tslib/tslib.js');
     goldenJs.push('test_files/augment/shim.js');
+    goldenJs.push('test_files/clutz.no_externs/default_export.js');
     goldenJs.push('test_files/clutz.no_externs/some_name_space.js');
     goldenJs.push('test_files/clutz.no_externs/some_other.js');
     goldenJs.push('test_files/clutz_type_value.no_externs/type_value.js');
--- a/test/e2e_closure_test.ts
+++ b/test/e2e_closure_test.ts
@@ -21,21 +21,27 @@ describe('golden file tests', () => {
   it('compile with Closure', (done) => {
     // Declaration tests do not produce .js files.
     const tests = goldenTests().filter(t => !t.isDeclarationTest);
+    // Collect all JavaScript outputs generated from .ts files.
     const goldenJs = ([] as string[]).concat(...tests.map(t => t.jsPaths));
-    goldenJs.push('src/closure_externs.js');
-    goldenJs.push('third_party/tslib/externs.js');
-    goldenJs.push('third_party/tslib/tslib.js');
-    goldenJs.push('test_files/augment/shim.js');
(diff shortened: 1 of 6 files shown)


test_files/clutz.no_externs/import_default.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * @fileoverview Reproduces a problem where a renamed Clutz default export ({default as X}) would
+ * produce type annotations including an indirection to the aliased symbol.
+ *
+ * Generated from: test_files/clutz.no_externs/import_default.ts
+ * @suppress {checkTypes,constantProperty,extraRequire,missingOverride,missingReturn,unusedPrivateMembers,uselessCode} checked by tsc
+ */
+goog.module('test_files.clutz.no_externs.import_default');
+var module = module || { id: 'test_files/clutz.no_externs/import_default.ts' };
+module = module;
```
brad4d: Time for me to learn something, I think.

Why is this `module = module;` assignment here?

test/e2e_closure_test.ts
```diff
@@ -26,6 +26,7 @@ describe('golden file tests', () => {
     goldenJs.push('third_party/tslib/externs.js');
     goldenJs.push('third_party/tslib/tslib.js');
     goldenJs.push('test_files/augment/shim.js');
+    goldenJs.push('test_files/clutz.no_externs/default_export.js');
```
brad4d: Why don't we need to push `import_default.js` here?

test_files/clutz.no_externs/import_default.js
mprobst: Good question, and it's a bit involved.

Some code in Angular requires a module bundler that exposes a `module.id` to the local file, containing the current module's path (not ID!). I.e. it depends on this API: https://nodejs.org/api/modules.html#modules_module_id. That's why there's a `var module = {id: ...}`.

Regarding the `var module = module || ...` bit, I no longer have full context on why that was needed. The code (in `googmodule.ts`) only comments `The following pattern ensures closure doesn't throw an error in advanced optimizations mode.`.

As for the last bit, `module = module`, this is to prevent an unused variable warning if `module` turns out not to be used. This is to accommodate projects that turn lint warnings into errors (which of course is a bad idea, but what can you do...). I think alternatively we could use `unusedLocalVariables` for the entire file, which would probably also be OK. We might not have been aware of that when we put this in, or might not have had the tech to reliably insert of fileoverview comment at the time.

test/e2e_closure_test.ts
mprobst: The line above first collects all `.js` files that are generated from a `.ts` file. The files manually added here are the ones not generated from a `.ts` file, i.e. files that contain the definition of a symbol that's only declared (i.e. in a `.d.ts` file) in TS land.

I've added comments explaining that.

PR #1118, opened 2019-11-13
Add BigIntLiteral to TypeTranslator#translate switch statement

kjin:
Currently, using tsickle with code that features a BigInt literal causes the following error: `Error: unknown type flags 2048 on {type flags:0x800}` -- this should fix that.


mprobst:
Merged!

PR #1119, opened 2019-11-14
Fixes running tsickle in browser

rrdelaney:
`path.normalize` relied on `ts.sys` being defined for access to the `resolvePath` function. `ts.sys` is [not defined when running in the browser](https://github.com/microsoft/TypeScript/blob/master/src/compiler/path.ts#L508-L520). However TypeScript exports another environment-independent [`resolveModule`](https://github.com/microsoft/TypeScript/blob/master/src/compiler/sys.ts#L1492-L1505) function along with the other utilities `path.ts` was using. Switching to this implementation should let tsickle work in the browser again, and all tests pass with this implementation so it seems mostly equivalent.

### DIFF
--- a/src/path.ts
+++ b/src/path.ts
@@ -25,6 +25,7 @@ declare module 'typescript' {
   function convertToRelativePath(
       absoluteOrRelativePath: string, basePath: string,
       getCanonicalFileName: (path: string) => string): string;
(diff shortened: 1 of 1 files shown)


mprobst:
Nice!

PR #1121, opened 2019-11-26
Refactor & fix rest parameter handling.

### LINKED ISSUE #1120: Tsickle emits wrong type for rest args as union of tuple types
This code

```ts
function foo(...args: [number]|[string, number]) {}
```

transforms to:

```js
/**
 * @param {...!Array<?>} args
 * @return {void}
 */
function foo(...args) { }
```

This suggests that each element should be an array, which is not the case. I would expect `Array<?>` to be `number` or `number|string`, but not an array of anything.

mprobst:
This change mostly refactors rest parameter handling to extract some of
the complex logic from the main parameter handling loop.

In addition, it fixes a subtle bug where tsickle would, if it failed to
resolve the array member type of a rest parameter, just continue
emitting the type as is. This "fail open" behaviour can cause issues,
such as emitting `Array<Array<?>>` for a non-representable rest tuple
type, instead of `Array<?>`.

Not a huge improvement and clearly a corner case, but I believe this
overall improves correctness and legibility of the code.

Fixes #1120.

### DIFF
--- a/src/module_type_translator.ts
+++ b/src/module_type_translator.ts
@@ -351,6 +351,35 @@ export class ModuleTypeTranslator {
   }
 
   /**
+   * resolveRestParameterType resolves the array member type for a rest parameter ("...").
+   * In TypeScript you write "...x: number[]", but in Closure you don't write the array:
+   * `@param {...number} x`. The code below unwraps the Array<> wrapper.
+   */
+  private resolveRestParameterType(
+      newTag: jsdoc.Tag, fnDecl: ts.SignatureDeclaration, paramNode: ts.ParameterDeclaration) {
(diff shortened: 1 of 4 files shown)


mprobst:
@kjin PTAL (and also check your inbox - I've invited you as a collaborator to this repo)

mprobst:
@rrdelaney in either case, we cannot express a tuple type rest parameter in Closure Compiler, so I'd rather fail consistently to `Array<?>` then try to be cheeky.

I think it's a really odd feature in TS as well - when would you ever want a rest parameter that's limited to N known parameters? Why not just take named args at that point?

rrdelaney:
Yea I entirely agree — it’s possible but the pattern when limited to known types isn’t useful at all.

PR #1126, opened 2019-12-02
Rationalize type/value conflict handling.

### LINKED ISSUE #1116: Rationalize handling of type/value conflicts
If some code declares a type/value conflict, e.g.

```ts
interface X { ... }
const X = ...;
```

Then tsickle backs off by omitting the type output.  This means nobody else can use "X" as a type. 
 (https://github.com/angular/tsickle/pull/1077 attempts to fix this, by mangling the the type name in these circumstances.  It's a pretty rare pattern though.)

However, in some circumstances we do believe that for a type/value conflict the type still exists:
- browser builtins (which are defined in lib.d.ts, but we assume the Closure externs match)
- clutz-generated types (because they are derived from Closure JS)

To summarize all the cases of conflicts and what we should do:
1. browser builtins: assume type exists
2. clutz-generated types: assume type exists
3. tsickle-generated output: either back off, or use mangled name

These three cases correspond to the functions (1) isClosureProvidedType, (2) isClutzDype, and (3) the above PR tries to catch these by adding a special header to our d.ts

I think it is never the case that we want to treat category 1 and 2 differently from each other, so this bug is about how we shouldn't have both of these functions, but rather one that covers both.  (Also, the impl of these functions is a big hack, so maybe we should instead say you're in case 1/2 any time you're *not* in case 3?)

To see this problem, add these lines to `test_files/clutz_type_value.no_externs`:
```ts
// We expect IFace to show up in the type signature.
let x: IFace;
```
It currently generates {?}.  The buggy code here only checks for category 1, not 2.  You can find said buggy code by grepping for isClosureProvidedType/isClutzDype and convincing yourself that any caller of these should really be calling both.

mprobst:
While separate in TypeScript, in Closure Compiler the type and value
namespaces are merged. This means tsickle must take care not to emit
duplicate symbols for type/value conflicts, and chooses the value over
the type.

That in turn means tsickle cannot generally emit references to types
that have a type value conflict, i.e. symbols in the TS API that have
both the Type and the Value SymbolFlag set.

However we special case two situations: (1) builtin symbols such as
`Array`, where we rely on the Closure definition and thus assume the
type/value conflict is handled on that level; and (2) symbols generated
by Clutz, which are also defined in Closure land so the conflict is not
a problem.

However previously we had two different code paths both handling parts
of this, one for type references in `@implements`, one for all other
type references. Both paths only handled either (1) or (2). This change
rationalizes the situation, handling (1) and (2) in both code paths.

Fixes #1116.


src/type_translator.ts
```diff
@@ -59,11 +48,23 @@ function isClutzDts(sourceFile: ts.SourceFile): boolean {
...
+ *
+ * - symbols provided by Clutz. Given that Closure has a merged type/value namespace, apparent
+ *   type/value conflicts on the TypeScript level are actually fine.
+ * - builtin lib*.d.ts symbols, such as "Array", which are considered Closure-compatible. Note that
+ *   we don't actually enforce that the types are actually compatible, but mostly just hope that
+ *   they are due to being derived from the same HTML specs.
  */
-export function isClutzType(symbol: ts.Symbol): boolean {
+export function typeValueConflictHandled(symbol: ts.Symbol) {
+  // TODO(#1072): if the symbol comes from a tsickle-transpiled file, either .ts or .d.ts with
```
blickly: I don't follow this comment.  Aren't tsickle output files always .js instead of .ts or .d.ts?

src/closure_externs.js
```diff
@@ -92,10 +92,15 @@ var GlobalFetch;
...
+ * @extends {WorkerGlobalScope}
+ * @extends {Window}
  */
 var WindowOrWorkerGlobalScope;
 
 /** @record */
 function CanvasDrawImage() {}
+
+/** @constructor */
+function CryptoKey() {};
```
blickly: Not sure I follow how this change is related to the rest of the PR?

mprobst: Sorry, I should have called that out. These are several commits in one pull request that are logically separate, but related. This change is required because the first change (better type emit) reveals a breakage where a user is missing this symbol, which is a builtin in TS but does not exist as an extern in JSCompiler.

src/type_translator.ts
mprobst: I copied this from a comment authored by Evan in the other file. We might need to ask him, but I think he's talking about a symbol that looks like a type/value conflict because how tsickle generated `.closure.js` or `.externs.js`, which in turn was then consumed by Clutz.

src/closure_externs.js
blickly: OK, makes sense. Thanks for clarifying.

src/type_translator.ts
blickly: If he were here, I'd ask him to suggest a clarification, but I don't think we should block this PR on that.

blickly:
LGTM

src/type_translator.ts
evmar: I tried to go through the cases of type/value conflicts here:
https://github.com/angular/tsickle/issues/1072#issuecomment-533894437
we now handle 1 and 2.  This comment was about 3 and 4.

The idea is that in TS-authored code with type/value conflicts, we could emit the type in JS with a mangled name like MyClass$type.  Due to separate compilation, for an import we only see the original .ts/.d.ts types, and not the emitted .js, so it's hard to know for a given import whether it is one where the name is mangled or not.

I have more info here
https://github.com/angular/tsickle/pull/1077
where I think the goal is mostly right (mangle the names) but the approach isn't right (how to tell when to mangle a name or not).

PR #1129, opened 2020-01-14
Add @suppress before @license tags in file overview comments.

EatingW:
Special case the `@license` tag in the file overview comments so that newly added `@suppress` tags are placed above `@license` tags.

`@license` tag treats all text that comes after it as part of the license, so if `@suppress` is added after `@license` tags, it is essentially ignored by the Closure Compiler.


evmar:
Sometimes we TGP these before merging here.  Do you expect this to be risky?

EatingW:
It shouldn't be risky - if anything changes it should include more suppressions than it had before as the new suppressions after license blocks were silently ignored. 
However, as a fan of TGP I'll do one tonight and report back in the internal change.

evmar:
I'll merge now to prevent another review round trip, we can always roll back if needed.

PR #1131, opened 2020-01-21
Updated Peer Dependencies since getTypeArguments is only available in…

xontab:
… typescript >3.7.x


xontab:
@googlebot I signed it!

mprobst:
Thanks for the fix!

PR #1132, opened 2020-01-24
Fix Tsickle when compiled by Closure Compiler

rrdelaney:
When Tsickle is used to compile Tsickle with Closure Compiler it breaks when encountering an `export` statement like: `export const t = 10`. This is due to some property renaming by the compiler, where a dot-access is used to call an internal method after casting to any. This was fixed by changing the method access to use brackets + string, so Closure Compiler will leave it alone.

### DIFF
--- a/src/googmodule.ts
+++ b/src/googmodule.ts
@@ -281,7 +281,7 @@ export function commonJsToGoogmoduleTransformer(
       // that a bundle by definition cannot be a goog.module()). The cast through any is necessary
       // to remain compatible with earlier TS versions.
       // tslint:disable-next-line:no-any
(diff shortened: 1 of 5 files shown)


rrdelaney:
@evmar This seems like the only one. It's a hack to get to an internal API, so it makes sense it's not common.

evmar:
I guess in this file https://github.com/angular/tsickle/blob/master/src/path.ts the use of 'declare' maybe prevents prop renaming?

rrdelaney:
Oof actually just did a search for a general `as any` and there's a few instances. Let me fix those up too.

PR #1133, opened 2020-02-21
Add a require to tslib for all generated JS files from tsickle.

EatingW:
As the comment in the code says, we need to always emit a goog.require('tslib') so that module manifests of Development mode contains the necessary 'tslib' module.

### DIFF
--- a/src/googmodule.ts
+++ b/src/googmodule.ts
@@ -573,6 +573,35 @@ export function commonJsToGoogmoduleTransformer(
             ts.createAssignment(ts.createIdentifier('module'), ts.createIdentifier('module'))));
       }
 
+      // Add `goog.require('tslib');` if not JS transpilation, and it hasn't already been required.
+      // Rationale:
+      // TS gets compiled to Development mode (ES5) and Closure mode (~ES6)
+      // sources. Tooling generates module manifests from the Closure version.
+      // These manifests are used both with the Closure version and the
+      // Development mode version. 'tslib' is sometimes required by the
+      // development version but not the Closure version. Inserting the import
+      // below unconditionally makes sure that the module manifests are
+      // identical between Closure and Development mode, avoiding breakages
+      // caused by missing module dependencies.
+      if (!host.isJsTranspilation) {
+        // Get a copy of the already resolved module names before calling
+        // resolveModuleName on 'tslib'. Otherwise, resolveModuleName will
+        // add 'tslib' to namespaceToModuleVarName and prevent checking whether
+        // 'tslib' has already been required.
+        const resolvedModuleNames = [...namespaceToModuleVarName.keys()];
+
+        const tslibModuleName =
+            host.pathToModuleName(sf.fileName, resolveModuleName(host, sf.fileName, 'tslib'));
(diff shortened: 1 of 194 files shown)


mprobst:
Can you make the commit message say *why* you're doing this? The PR description will get lost in history, the commit message is here to stay.

EatingW:
Done. Also squashed the check_format fix into one commit.

EatingW:
Actually, I'm going to close this PR now because there are some additional complications. Will reopen again when ready.

EatingW:
Additional complications resolved. Reopening

src/googmodule.ts
```diff
@@ -573,8 +573,38 @@ export function commonJsToGoogmoduleTransformer(
             ts.createAssignment(ts.createIdentifier('module'), ts.createIdentifier('module'))));
       }
 
-      // Insert goog.module() etc after any leading comments in the source file. The comments have
-      // been converted to NotEmittedStatements by transformer_util, which this depends on.
+      // Add goog.require('tslib');
```
mprobst: nit: line wrapping is odd

src/googmodule.ts
```diff
@@ -573,8 +573,38 @@ export function commonJsToGoogmoduleTransformer(
...
+      // sources. Tooling generates module manifests from the Closure version.
+      // These manifests are used both with the Closure version and the
+      // Development mode version. 'tslib' is sometimes required by the
+      // development version but not the Closure version. Inserting the import
+      // below unconditionally makes sure that the module manifests are
+      // identical between Closure and Development mode, avoiding breakages
+      // caused by missing module dependencies.
+      if (!host.isJsTranspilation) {
+        // Need to get a copy of the already resolved module names before
+        // calling resolveModuleName on 'tslib'
```
mprobst: don't document what you're doing, but why you're doing it. E.g. the bit that's missing here is why you need to do this.

src/googmodule.ts
```diff
@@ -573,8 +573,38 @@ export function commonJsToGoogmoduleTransformer(
...
+        if (resolvedModuleNames.indexOf(tslibModuleName) === -1) {
+          const tslibImport = ts.createExpressionStatement(
+              createGoogCall('require', createSingleQuoteStringLiteral(tslibModuleName)));
+
+          // Place the goog.require('tslib') statement right after the
+          // goog.module statements
+          headerStmts.push(tslibImport);
+        }
+      }
+      // Insert goog.module() etc after any leading comments in the source
```
mprobst: don't rewrap the comment. In tsickle open source, column limit is 100 chars (for silly reasons, but worth sticking to for the moment here)

src/googmodule.ts
EatingW: done

src/googmodule.ts
EatingW: Done. How does the new version look?

src/googmodule.ts
EatingW: Done

mprobst:
Looks good, merged!

PR #1134, opened 2020-02-27
Add support for TypeScript 3.8

rrdelaney:
This upgrades the version of TypeScript Tsickle uses as a `devDependency` and a `peerDependency`, which in turn upgrades the `tslib` version.

Support for Private Fields
--------------------------
Tsickle mostly ignores private fields, however it no longer warns when not generating externs for them. Externs are not generated because the fields no not exist on the class when downleveled.

Support for `export * as ns` Syntax
-----------------------------------
Tsickle compiles:

```ts
export * as ns from './namespace';
```

to

```ts
var tsickle_module_1_ = goog.require('project.namespace');
exports.ns = tsickle_module_1_;
```

New `tslib` Functions
---------------------
Private fields require two new functions in `tslib`: `__classPrivateFieldGet` and `__classPrivateFieldSet`. Both of these were added to Tsickle's version of `tslib.js` along with Closure type annotations.

Other Changes
-------------
* The `import_export_typedef_conflict` test was removed because that code is no longer valid in TypeScript 3.8, and produces a compile error.
* `tslib@1.11.0` was added as a `devDependency`. Before it was being pulled in as a transitive dependency, but the packages pulling it in depend on `1.10`. To upgrade to TypeScript 3.8 Tsickle needs the latest version of `tslib`.


src/externs.ts
```diff
@@ -518,6 +518,11 @@ export function generateExterns(
           debugLocationStr(exportDeclaration, namespace)}\n`);
       return;
     }
+    if (ts.isNamespaceExport(exportDeclaration.exportClause)) {
+      emit(`\n// TODO(tsickle): export * as declaration in ${
```
mprobst: can you open an issue to track adding support?

I think it'd be something like the assignment in the loop below.

src/jsdoc_transformer.ts
```diff
@@ -893,8 +902,9 @@ export function jsdocTransformer(
           }
           exportDecl = ts.updateExportDeclaration(
               exportDecl, exportDecl.decorators, exportDecl.modifiers,
-              ts.createNamedExports(exportSpecifiers), exportDecl.moduleSpecifier);
-        } else {
+              ts.createNamedExports(exportSpecifiers), exportDecl.moduleSpecifier, false);
```
mprobst: Can you give the `false` a name?

```suggestion
          const withBananas = false;
          exportDecl = ts.updateExportDeclaration(
              exportDecl, exportDecl.decorators, exportDecl.modifiers,	              exportDecl, exportDecl.decorators, exportDecl.modifiers,
              ts.createNamedExports(exportSpecifiers), exportDecl.moduleSpecifier);	              ts.createNamedExports(exportSpecifiers), exportDecl.moduleSpecifier, withBananas);
```

test/googmodule_test.ts
```diff
@@ -420,6 +420,7 @@ goog.loadedModules_['legacy.bar.baz'] = { exports: exports, type: goog.ModuleTyp
         import * as starImport from './relpath.js';
         import {namedImport, renamedFrom as renamedTo} from '../dotdot/file.js';
         export * from './exportStar.js';
+        export * as ns from './exportStarAsNs.js';
```
mprobst: nit: maybe add a new `it` block for this? This is shoehorning a lot of different functionality into a single test case. Not your fault, but let's maybe stop adding to it :-)

mprobst:
Nice!

src/externs.ts
rrdelaney: Filed #1135 and added a TODO!

src/jsdoc_transformer.ts
rrdelaney: Added as `isTypeOnlyExport`.

test/googmodule_test.ts
rrdelaney: Good suggestion, done! 😄

PR #1139, opened 2020-03-11
Fix bug in output of imported Closure types

lauraharker:
Without the tsickle change, `import_from_goog.js` would have the annotation
```
/** @type {!tsickle_LegacyModule_3.module$contents$closure$LegacyModule_LegacyExport} */
```
instead of the correct annotation
```
/** @type {!tsickle_LegacyModule_3.LegacyExport} */
```


test_files/import_from_goog/externs.js
```diff
@@ -25,3 +25,23 @@ goog_closure$OtherModule.SymA = function() {};
  * @struct
  */
 goog_closure$OtherModule.SymB = function() {};
+/** @const */
+var ಠ_ಠ = {};
```
evmar: This is a little weird; the test suite sees the d.ts file and assumes it needs externs.
I think you can rename the directory (import_from_goog) to add a '.no_externs' suffix:
https://github.com/angular/tsickle/blob/f19e442a3e693b6fadcba2f1c206fefa92862b36/test/golden_tsickle_test.ts#L215
and then delete this file entirely.

I think we don't care too much either way, but might as well?

lauraharker: Sounds good, done.

PR #1146, opened 2020-05-19
Unwrap ES2015+ classes when using TypeScript 3.9

clydin:
Classes that contain TypeScript Decorators and/or static properties will be wrapped in an IIFE when using TypeScript 3.9.  This transformer will remove the wrapping to allow analysis by closure compiler.

Tested against TypeScript 3.9.2 in combination with the changes in #1143

### DIFF
--- a/src/BUILD
+++ b/src/BUILD
@@ -21,6 +21,7 @@ ts_library(
         "transformer_util.ts",
         "tsickle.ts",
         "type_translator.ts",
+        "unwrap_classes.ts",
     ],
     data = ["closure_externs.js"],
     tsconfig = "//:tsconfig.json",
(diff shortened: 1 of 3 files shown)


rrdelaney:
First of all, thanks for the PR! This is something that is definitely blocking adoption of TypeScript 3.9 for codebases using tsickle. We were hoping to [land a compiler option in upstream TypeScript](https://github.com/microsoft/TypeScript/pull/38602) to deal with this rather than relying on their implementation and rewriting it. Having the option there will keep tsickle-flavored emit closer to upstream emit, and less divergence will be good for future maintainability. In the meantime, please follow the [tracking issue in the TypeScript repo](https://github.com/microsoft/TypeScript/issues/38374) for updates!

PR #1147, opened 2020-05-19
forbid '@protected' in TypeScript comments

evmar:
We already forbade @private and @public; I think this was an oversight.

The reason we forbid these is because there are TS builtin equivalents,
and some users aren't aware of them.  I have already fixed the existing
users to make this change safe to land.

PiperOrigin-RevId: 309953800


evmar:
(This is a push of an internal change, it's already been reviewed)

Issue #1155, opened 2020-06-25
Update peerDependency on Typescript to ~3.9.x


bjornharvold:
Cannot update to Angular 10 with forcing an ng update at the moment.

`ng update @angular/core @angular/cli fails`

Have to type
`ng --force update @angular/core @angular/cli`

lauraharker:
I believe @rrdelaney is working on upgrading to TS3.9.

Created Google internal issue http://b/159923702.

aramperes:
Please bump this ticket when a new version of `tsickle` is released. Thanks for the quick turnaround!

rrdelaney:
New version should be published!

aramperes:
Looking at https://registry.npmjs.org/tsickle (`versions -> 0.39.0 -> peerDependencies`), the TypeScript peerDependency is still at `~3.8.2`. Only the devDependency seems to have been updated.

rrdelaney:
Sorry about that, should be fixed! Thanks for the PR!

PR #1161, opened 2020-07-02
Update TypeScript peerDependency (#1155)

aramperes:
TypeScript 3.9 was added as a devDependency, but not a peerDependency.

Ref: #1155


aramperes:
@googlebot I signed it!

PR #1162, opened 2020-07-06
rel: 0.39.1

rrdelaney:
Also overrides changes from internal sync so this commit will pass tests


rrdelaney:
Should be good for review, the PR as marked as a draft because it won't be submitted per https://github.com/angular/tsickle/blob/master/README.md#releasing

PR #1165, opened 2020-07-22
Migrate tsickle to TypeScript 4.0.

mprobst:
This mostly involves adjusting a number of callsites to the new
`ts.factory` API instead of the previous `ts.getMutableClone`.

Additionally it required several test changes:
- an unspecified global this no longer is any
- properties may no longer override getters
- stricter checks of property initialization


package.json
```diff
@@ -11,7 +11,7 @@
     "src/*"
   ],
   "peerDependencies": {
-    "typescript": "~3.9.5"
+    "typescript": "^4.0.0-dev"
```
evmar, line 14: We should use patterns like `~4.0` so that we don't implicitly pull in e.g. 4.1.

src/transformer_util.ts
```diff
@@ -153,10 +153,15 @@ export function updateSourceFileNode(
   if (statements === sf.statements) {
     return sf;
   }
-  // Note: Need to clone the original file (and not use `ts.updateSourceFileNode`)
-  // as otherwise TS fails when resolving types for decorators.
-  sf = ts.getMutableClone(sf);
-  sf.statements = statements;
+  sf = ts.factory.updateSourceFile(
```
evmar, line 156: Man, I just copied the old pattern into my current project too...

test_files/async_functions/async_functions.ts
```diff
@@ -37,9 +37,9 @@ class Container {
       const s = await 3;
       return s + this.field;
     };
-    async function asyncFunctionInMethod(param: string): Promise<string> {
+    async function asyncFunctionInMethod(this: Container, param: string): Promise<string> {
       const s = await 3;
-      return s + this.field;
+      return s + (this as any).field;
```
evmar, line 42: I think this isn't right, and might break dis/ambguation.

I can't tell from this change if it's meant to be a workaround, or if we expect user code to change in the same way?

mprobst: Good catch. I was playing around with this and didn't remove the as any cast again. This ultimately didn't need a change I think.

package.json
mprobst: I went with ~4.0.0-dev so that we get the dev packages and patch levels in 4.0, but not 4.1

test_files/async_functions/async_functions.ts
mprobst: The tricky bit is that TS compiler now disallows having a function without the this type specified here, so I needed to add an explicit this param type. That means a code path in tsickle is now under-exercised (never have a nested fn that needs a this * added), but I think that's OK.

mprobst:
@rrdelaney we can keep it in this branch for the time being. I'll create an internal CL that we can patch to test the new version.

mprobst:
This has been committed by @rrdelaney in 39f36e82c3d29030a66cae8d48993951c8efcc95.

PR #1178, opened 2020-08-13
add ImportMeta to closure_externs.js

theseanl:
ImportMeta is declared in [lib.es5.ts](https://github.com/microsoft/TypeScript/blob/610fa28de55d37b0e46fa1cf6394da117d29be9d/lib/lib.es5.d.ts#L614), and is referenced in
[@types/node](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2833781317b9e009c81195998b1f67c9e67d1c33/types/node/globals.d.ts#L24).
Other than `ImportMeta`, generating externs from `@types/node` at least works (it does not bail out with an error).


mprobst:
Can you add a test case that demonstrates the problem? E.g. a `.ts` file that references `ImportMeta`?

(I'm not sure I follow what the problem is, i.e. what error it bails out with. Steps to repro & output would also help!).

theseanl:
I've added a test for this change. I've added a `use_closure_externs_decl.d.ts` file and `externs.js` file. When these files are fed to closure compiler in `closure_e2e_test`, the compiler will produce an error `ERROR - [JSC_UNDEFINED_EXTERN_VAR_ERROR] name ImportMeta is not defined in the externs.` without the addition of `ImportMeta` to `closure_externs.js`.
This test could be made without introducing a new file by adding some statements in `use_closure_externs.ts` such as `let i: ImportMeta = {};`, but since `closure_e2e_test` is already failing in master and it produces no extra warning, it is not visible that the addition of `ImportMeta` fixes something, so I've make the error occur in an externs file.

mprobst:
I think I understand. Given this TypeScript code:

```
// foo.d.ts
interface ImportMeta { something: string; }
```

tsickle will generate just the declaration of `ImportMeta.something` in the externs, because it thinks `ImportMeta` was already declared (by the builtins).

Correct?

I think @evmar fixed the master build, but indeed we don't emit errors on missing types in Closure. We should fix in the stes suite that some time, even if it means diverging from how we run in production.

So this all makes sense, and thanks for the contribution.

PR #1267, opened 2021-05-27
build & test using npm scripts.

mprobst:
This change migrates the open source build from bazel back to plain npm
scripts. bazel worked well for a while, but the internal and external
build definitions are diverging, and we no longer have the expertise to
maintain either of them.

While simplistic, plain npm scripts work well for our use case, both
from a performance and a maintainability perspective.


lvelden (approved):

mprobst:
Closing this as it landed (though with some merge issues).

PR #1270, opened 2021-06-07
Update and rename node.js.yml to node.js

BigGuy573:


mprobst:
Thanks for your contribution!

I don't quite understand it yet. Can you explain what you intend to achieve here?

FWIW, the change as is looks suspect to me.

mprobst (changes requested):
Please explain what this is useful for.
Please double check the syntax you're introducing, the changes look suspect.

PR #1324, opened 2022-01-25
Improve tsickle release scripts.

mprobst:
* make sure to build before npm publish and npm pack.
* use $(yarn bin) in scripts to make them less dependent on directory
  layout.
* submit yarn.lock files for reproducible dependencies.
* add a .gitignore (which was lost in some internal/external sync).


package.json
```diff
@@ -31,10 +31,11 @@
     "typescript": "h-joo/TypeScript#ts45-no-double-comments"
   },
   "scripts": {
-    "build": "tsc ",
+    "build": "$(yarn bin)/tsc ",
```
frigus02, line 34: I think this is not necessary. At least with `yarn build` it should find and execute any binaries from node_modules automatically.

frigus02 (approved):

package.json
mprobst: Good to know, done.

